PR #1, opened 2017-03-30
RATIS-8: Avoid using Optional for data fields in TransactionContex

hmcl:


bshashikant:
@hmcl , can you please clarify why this patch is required?

LiuJinghao:
Actually, using Optional is a good practice.

LiuJinghao:
这是一封自动回复邮件。已经收到您的来信，我会尽快回复。

PR #2, opened 2018-05-23
RATIS-244. Skip snapshot file if MD5 file does not exist

andy2git:


ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
```diff
@@ -104,7 +104,7 @@ public SingleFileSnapshotInfo findLatestSnapshot() throws IOException {
              Files.newDirectoryStream(smDir.toPath())) {
       for (Path path : stream) {
         Matcher matcher = SNAPSHOT_REGEX.matcher(path.getFileName().toString());
-        if (matcher.matches()) {
+        if (matcher.matches() && MD5FileUtil.existDigestFile(path.toFile())) {
```
andy2git: One concern I have is: do we need to delete the maybe corrupted snapshot file? If we do not purge logs, we can always replay from the commit logs.

PR #4, opened 2018-08-16
RATIS-272 logservice api

joshelser:
WIP. Based on Rajeshbabu's first API suggestions. Also adds:

* LogStream and LogService interfaces
* LogName to uniquely identify a LogStream in a LogService
* Listener (Tail) API for a LogStream


ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogWriter.java
```diff
@@ -0,0 +1,43 @@
...
+   */
+  long write(ByteBuffer data) throws IOException;
+
+  /**
+   * Guarantees that all previous data appended by {@link #write(ByteBuffer)} are persisted
+   * and durable in the LogStream.
+   *
+   * @return TODO Unknown?
+   */
+  long commit() throws IOException;
```
chrajeshbabu: @joshelser This need to be changed to sync.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.List;
+
+/**
+ * Synchronous client interface to read from a LogStream.
+ */
+public interface LogReader extends AutoCloseable {
+
```
ankitsinghal: Need a way to get the recordId of the current record read (getCurrentRecordId()?), it will be needed to keep a track so that we can start from where we left after any failure (using seek)

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogWriter.java
```diff
@@ -0,0 +1,43 @@
...
+   * @return The recordId for the record just written
+   */
+  long write(ByteBuffer data) throws IOException;
+
+  /**
+   * Guarantees that all previous data appended by {@link #write(ByteBuffer)} are persisted
+   * and durable in the LogStream.
+   *
+   * @return TODO Unknown?
+   */
```
ankitsinghal: Return last recordId committed.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
```diff
@@ -0,0 +1,109 @@
...
+   */
+  LogReader createReader();
+
+  /**
+   * Creates a write to write to this LogStream synchronously.
+   *
+   * @return A synchronous writer
+   */
+  LogWriter createWriter();
+
```
ankitsinghal: I believe the plan is to support a single writer(otherwise commit() would get tedious to manage at server), right ? can we make it getWriter()?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
```diff
@@ -0,0 +1,109 @@
...
+   * @param recordId A non-negative recordId for this LogStream
+   */
+  CompletableFuture<Void> truncateBefore(long recordId);
+
+  /**
+   * Returns the recordId which is the start of the LogStream. When there are records which were truncated
+   * from the LogStream, this will return a value larger than {@code 0}.
+   */
+  CompletableFuture<Long> getFirstRecordId();
+
```
ankitsinghal: Similarly getlastRecordId() would be helpfull.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/AsyncLogReader.java
```diff
@@ -0,0 +1,54 @@
...
+public interface AsyncLogReader extends AutoCloseable {
+
+  /**
+   * Seeks to the position before the record at the provided {@code recordId} in the LogStream.
+   *
+   * @param recordId A non-negative, recordId in the LogStream
+   * @return A future for when the operation is completed.
+   */
+  CompletableFuture<Void> seek(long recordId) throws IOException;
+
```
ankitsinghal: Can add reset() also to reset according to the logStream getFirstRecordId().

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogService.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public interface LogService {
+
+  /**
+   * Creates a new {@link LogStream} identified by the given name. Throws
+   * an exception if a {@link LogStream} with the given name already exists.
+   *
+   * @param name Unique name for this LogStream.
+   */
+  CompletableFuture<LogStream> createLog(LogName name);
```
ankitsinghal: While creating log , we may need to add some metadata for the stream, for eg:- Details of SERDE classes, compression etc.
And need the corresponding API in LogStream to get the MetaData.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/ArchivedLog.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * A {@link LogStream} which has been archived in some external
+ * system. This interface is parameterized to allow for implementations
+ * to use their own class to encapsulate how to find the archived log.
+ *
+ * In the majority of cases, this should be transparent to end-users, as
+ * the {@link LogStream} should hide the fact that this even exists.
+ * TODO maybe that means this should be client-facing at all?
+ *
```
ankitsinghal: Yes, it should not be client-facing at all. Just knowing from logStream that if I'm reading from archived or active location should be fine(recordId<getArchivedRecordId()).

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Copies all records from the beginning of the LogStream until the given {@code recordId}
+   * to the configured archival storage.
+   *
+   * @param recordId A non-negative recordId for this LogStream
+   */
+  CompletableFuture<Void> archiveBefore(long recordId);
```
chrajeshbabu: May be we can have a way to say wether the passed id is exclusive or inclusive.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
```diff
@@ -0,0 +1,109 @@
...
+   * @return A synchronous writer
+   */
+  LogWriter createWriter();
+
+  /**
+   * Removes the elements in this LogStream prior to the given recordId.
+   *
+   * @param recordId A non-negative recordId for this LogStream
+   */
+  CompletableFuture<Void> truncateBefore(long recordId);
```
chrajeshbabu: Here also we should have a way whether to include the passed record id or not. If we don't want to include the passed recordid then we client of log service need to increment it.

joshelser:
Thanks @ankitsinghal  and @chrajeshbabu. I have some more changes staged locally, too. Will update what I have (along with some "noop" implementations to show an e2e example) with your feedback!

VladRodionov:
LogName is kind of artificial thing. We should adopt standard Unix convention - everything is a file, Path in our case.  java.io.Path gives us direct way to LogService file system.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogName.java
```diff
@@ -0,0 +1,28 @@
+package org.apache.ratis.logservice.api;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Identifier to uniquely identify a {@link LogStream}.
+ */
+public class LogName {
```
VladRodionov: LogName -> java.io.Path. Let us adopt file system for log service

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
```diff
@@ -0,0 +1,53 @@
...
+   */
+  void seek(long recordId) throws IOException;
+
+  /**
+   * Reads the next record from the LogStream at the current position and advances the current position
+   * to after the record which was just returned.
+   *
+   * @return The data for the next record.
+   */
+  ByteBuffer readNext() throws IOException;
```
VladRodionov: readNext(ByteBuffer) -  to reuse existing buffer

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Reads the next {@code numRecords} records from the LogStream, starting at the current position. This method
+   * may return fewer than requested records if the LogStream does not have sufficient records to return.
+   *
+   * @param numRecords The number of records to return
+   * @return The records, no more than the requested {@code numRecords} amount.
+   */
+  List<ByteBuffer> readBulk(int numRecords) throws IOException;
```
VladRodionov: The same here as well. API call to reuse existing buffers

joshelser:
> We should adopt standard Unix convention - everything is a file, Path in our case. java.io.Path gives us direct way to LogService file system.

Good point. That makes sense.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/ArchivedLog.java
```diff
@@ -0,0 +1,43 @@
...
+ * @param <T> A referent to the log on the external system.
+ */
+public interface ArchivedLog<T> extends AutoCloseable {
```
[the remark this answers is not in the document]
joshelser: Yeah, I was mulling over whether or not to even include it in this initial drop.

I like the idea of a consistent API to read from Log, regardless of it being "active" (still in the state machine) or "archived" (exported to some remote storage).

> Log archiving is the feature HBase requires

I think that this is a generally reusable feature that users would like. It would help keep the state machine "small" and consistent performance.

> HBase client will create archive subdirectory and will be moving old files to this subdirectory

I think automated archival should be a feature that LogService provides. HBase should be unaware that we are moving "segments" of data around.

> I want to enhance LogService with FileSystem-like API:

I have changes that have been sitting locally for too long. I'm prioritizing getting them finished and then update this. I think being able to think of the LogService like a FileSystem is nice. I'm curious if you have put more thought about how the notion of "directories" might be used :)

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
joshelser: Can certainly expand this to be `archive(long recordId, boolean inclusive)`. I like "archive" much more than "archiveBefore" ;)

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/ArchivedLog.java
joshelser: Yeah, Vlad brought up the same point. Let me drop it client API -- this is something we probably want as internal API.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogService.java
joshelser: > we may need to add some metadata for the stream, for eg:- Details of SERDE classes, compression etc.

Good point. I can make a basic configuration class which just wraps a `Map<String,String>` for now? How does that sound while we flesh out what kind of metadata we'll have?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/AsyncLogReader.java
joshelser: If you think that would be helpful, we can. I am wondering about the frequency of use of such an API. Does it make sense to have that "wrapper" when we have something like `seek(getFirstRecordId())` already present? I don't feel strongly either way :)

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
joshelser: My hope was that we could support multiple, independent writers. The writer implementation would be able to hide most details (I hope).

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogService.java
```diff
@@ -0,0 +1,53 @@
...
+   * Fetches the {@link LogStream} identified by the given name.
+   *
+   * @param name The name of the LogStream
+   */
+  CompletableFuture<LogStream> getLog(LogName name);
+
+  /**
+   * Lists all {@link LogStream} instances known by this LogService.
+   */
+  CompletableFuture<Iterator<LogStream>> listLogs();
```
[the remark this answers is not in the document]
joshelser: If we go towards Vlad's suggestion of a FileSystem-like API, we would give a `PathFilter` that would give the same kind of functionality. Having the notion of directories actually gets infinitely easier to find just logs you want too :)

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
joshelser: On second thought "archiveBefore" would still be correct. I still like the inclusivity.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/ArchivedLog.java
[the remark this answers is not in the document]
joshelser: Looking more closely at java.nio.file.Path, java.nio.file.WatchService, java.nio.file.FileSystem, and java.nio.file.spi.FileSystemProvider, I think that there are definitely pieces which make sense for us to use, but I think we also get a lot of baggage that we don't want (that we would just always have implemented with a RuntimeException).

The WatchService is basically what the RecordListener was meant to be. I think using a "filesystem path" (e.g. /a/b/c) does make sense and gives us flexibility for the LogService overall (instead of just "log name").

What about something in between? Take some of the good parts of Path and FileSystem, and pull them into LogService directly?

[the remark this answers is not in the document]
VladRodionov: We are not talking about full fledged FileSystem API here, but some API would help in adoption, taking into account familiar file system abstraction. Besides this, as a HBase developer I have my own requirements (see my comments above) to this Log Service and they can be fulfilled only by adopting FS-like API.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/AsyncLogReader.java
```diff
@@ -0,0 +1,54 @@
...
+  CompletableFuture<ByteBuffer> readNext() throws IOException;
+
+  /**
+   * Reads the next {@code numRecords} records from the LogStream, starting at the current position. This method
+   * may return fewer than requested records if the LogStream does not have sufficient records to return.
+   *
+   * @param numRecords The number of records to return
+   * @return A future providing the records, no more than the requested {@code numRecords} amount.
+   */
+  CompletableFuture<List<ByteBuffer>> readBulk(int numRecords) throws IOException;
```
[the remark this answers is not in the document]
VladRodionov: Sure, let us do not breed buffers - they are expensive :)

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
VladRodionov: Multiple writers - multiple readers is a MUST for generic Log Service (not HBase), but close() must be provided. On commit() - I would replace it with sync() (sync to durable storage). Commit implies rollback :)

joshelser:
Breaking this out to a top-level thread instead of being anchored on a code segment:

> We are not talking about full fledged FileSystem API here, but some API would help in adoption, taking into account familiar file system abstraction.

I had a chat with Ankit and Sergey about this one, trying to weigh the pros and cons. A FileSystem-like API would give users some familiarity in working with the LogService, but the general feeling was that we may be over-burdening ourselves in the first implementation. There would be many things we could potentially implement via a FileSystem like API which we wouldn't necessarily need for HBase as an initial consumer.

I think we could explore using a FileSystem-like API in the future to supplement a more minimal core API (closer to what is here already). Something higher-level could also be done as we get closer to HBase's rearchitecting.

I think it's becoming pretty clear that HBase's new WAL API won't be written using the notion of a File on a FileSystem (and instead use a "stream-like" API at the core). I could see us integrating the Ratis LogService into HBase via a FileSystem facade.

How do you feel about that, @VladRodionov? Are there any immediate concerns which would limit how we're thinking about building out the LogService if we didn't have a FileSystem facade?

VladRodionov:
Hmm, read, write, create, open, seek, close, sync, list, delete etc. What else? Is not this already File I/O? They are not streams as long as you provide positional access. They are files. In any case you will need API, which closely matches File I/O.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/ArchivedLog.java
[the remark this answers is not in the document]
VladRodionov: That is something different. Raft vs Zab/Chubby. Log Service is based on Raft, so we are kosher here :) - we do not build distributed system on top of FS abstraction, just we need something similar to use Log Service at its fullest.  No, we are not providing Posix-compatible API nor even Hadoop - compatible API, although I think, potentially, this would be a nice feature: mount/unmount log service as a file system.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/ArchivedLog.java
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Creates an asynchronous reader over this archived log.
+   */
+  AsyncLogReader createAsyncReader();
```
VladRodionov: It would be nice to have a clarification on all async stuff (readers/writers).

1. Do they improve performance?
2. Do they improve latencies? and can we measure this improvement?
3. Do they simplify programming model?

My guess on all three is they don't. Async model (not based on OS kernel support) is broken in Java.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public interface LogReader extends AutoCloseable {
+
+  /**
+   * Seeks to the position before the record at the provided {@code offset} in the LogStream.
+   *
+   * @param offset A non-negative, offset in the LogStream
+   * @return A future for when the operation is completed.
+   */
+  void seek(long recordId) throws IOException;
```
VladRodionov: Seeking by record id is expensive to implement. One has to support separate index {recordid -> file offset}. This should be not ID, but generic "offset", which maps 1 -> 1 to file offset. The only use case, where we need positional reads (at least in logs) is to resume sequential reading from the last read offset (HBase replication as an example).
We have to separate logs from message queues - they are totally different by their purpose.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogService.java
VladRodionov: Again, why is this async operation and why we expose this to a client?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
VladRodionov: Do we really need this one? Use case?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
VladRodionov: Do we really need this? What is the use case?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
```diff
@@ -0,0 +1,109 @@
...
+   * to the configured archival storage.
+   *
+   * @param recordId A non-negative recordId for this LogStream
+   */
+  CompletableFuture<Void> archiveBefore(long recordId);
+
+  /**
+   * Returns the recordId, prior to which, all records in the LogStream are archived.
+   */
+  CompletableFuture<Long> getArchivalPoint();
```
VladRodionov: What does concept of log archive give us? We either need log (regardless its status: active/archive) or we do not.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogWriter.java
VladRodionov: Offset

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/RecordListener.java
```diff
@@ -0,0 +1,19 @@
...
+ * to that LogStream until it is removed.
+ */
+public interface RecordListener {
+
+  /**
+   * Processes the written record from the LogStream.
+   *
+   * @param record The record
+   */
+  CompletableFuture<Void> receiveRecord(ByteBuffer record);
```
VladRodionov: Again, do not break Listener contract - its synchronous.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogService.java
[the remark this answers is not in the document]
VladRodionov: Josh, how are you going to implement this w/o FileSystem-like API?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
ss77892: That's why most of the implementations of distributed logs (pub/sub model) actually are using segments/slices/whatever they call it and why usually they are relatively small. Usually, there is a separate index that maintains the map between id and segment OR it keeps start id for the segment in its name.  What do you mean by 'separate logs from message queues'?

ss77892: One more thing. Actually, such seek operation should be already implemented under the hood because it's required for quick sync up for lagging quorum nodes due network problems/restarts ( didn't dig into the code yet, but almost 100% sure why segments were implemented there). All we need is to provide a facade that can be used outside.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogService.java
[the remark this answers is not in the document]
ss77892: I would better ask whether this method should be here. Ratis is an embedded library that allows creating a SINGLE distributed state machine. That means a SINGLE log. A distributed log service that would allow creating a number of logs (i.e. state machines) sounds like an application that would include state machines creation/management. That still can be a part of the embedded library, but it would make everything more complicated and less flexible (i.e. different use cases may require special rules for quorum members like rack awareness, latency measurement, RS Groups in case of HBase, etc).

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
VladRodionov: I meant what I meant: logs are not a message queues. Period. There is no need for random access in log file to grab a particular log record, but there is such a need in message queue. As for seeking support in Raft log - there is no such requirement to implement this efficiently, quite contrary, afair, protocol describes very stupid and lazy algo of synching two logs: leader goes backward one record at a time and compares it with a lagging follower records. 
Do not bring please functionality we (Log service users) are not asking for.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
joshelser: A LogStream is a sequence of entries: <recordId, byte[]> and RecordId is monotonically increasing. Truncation is the removal of entries from the "head" of the stream.

e.g. if we have a log with recordIds from `[0,100]`, we might say `truncateBefore(10, false)` which would remove entries 0, 1, 2, ..., 9. The "head" of the log would then be `recordId=10`

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/ArchivedLog.java
joshelser: Dropping all of the async API for now. Might help later but we should do that with purpose.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
joshelser: Moving to a specific point in a LogStream is absolutely a necessary API call. We must be able to resume reading from a known point in the Log. HBase Replication needs that functionality.

A naive implementation can read all records from the LogStream to know where to start reading. A slightly more advanced implementation can use the underlying segment files to find the right segment file to read from. An advance implementation could consult an index which we explicitly maintain over all segment files.

We haven't touched on implementation of how the LogReader class would negotiate this call with the back-end StateMachine (that depends on what the StateMachine and/or Ratis APIs look like, I think).

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/RecordListener.java
joshelser: Dropping this with the other async stuff.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStream.java
joshelser: Yeah, need to rework this based on the doc I updated on RATIS-271 -- thanks!

joshelser:
Dropped a big update that tries to account for the implementation decisions.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
```diff
@@ -0,0 +1,82 @@
...
+   */
+  ByteBuffer readNext() throws IOException;
+
+  /**
+   * Reads the next record from the LogStream at the current position into the provided {@link buffer} and
+   * advances the current position to the point after the record just read.
+   *
+   * @param buffer A buffer to read the record into
+   */
+  void readNext(ByteBuffer buffer) throws IOException;
```
joshelser: @VladRodionov one concern I just realized is: what if the buffer's size is less than the size of the record we need to read? As written now, we would have to throw an Exception. I don't know if we have flexibility within ByteBuffer to fix this.

One way around it would be to build in some kind of ByteBufferPool to Ratis itself that would reuse buffers and reduce the work a client would have to do. WDYT?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
VladRodionov: >> Moving to a specific point in a LogStream is absolutely a necessary API call

I think everyone here agree with you. The issue is keeping separate indexes for records. I do not know how this was implemented in Ratis, but in any case, maintaining index eats CPU and IO and slow everything down. If it is there already - fine, we can reuse it. To resume HBase replication  - offset in a log stream (not record id) would suffice (in our use case, of course).

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
VladRodionov: No problem, exception should be thrown. It is up to user to decide what is the minimum safe buffer size is.

joshelser: Ok, sounds good. Let me update the javadoc at least to call this out.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogService.java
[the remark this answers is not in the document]
VladRodionov: Sergej, state machine can track more than one log. I have mentioned already very simple  design of a STATE: it is a key-value map, where keys are logs (fully qualified names), and values are current recordId or offset.  In this case, we will have to separate actual logs from a Raft log, otherwise all the data will be in a single Raft log, which makes everything more complex during RS failover. Raft protocol and implementation is not a Bible, we can add our own bits and pieces which suite our goals.

joshelser:
Rebased on master, updated javadoc, added a factory for LogService creation, and fixed a couple of other trivial cleanup things.

Good for a first draft, @VladRodionov @chrajeshbabu ?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogService.java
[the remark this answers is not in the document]
ss77892: @VladRodionov I completely agree with you that it's possible to implement more than one log using a single state machine. I was talking about the current Ratis implementation which has a single log per state machine. The benefit we have is that the replication of this log is a part of the implementation and we shouldn't implement anything on our own. If I understand correctly you suggest implementing multiple logs separately. That would mean 1) double writes (still all that data will go to RatisLog)  2) Special handling for those records to write it to the correct log. From my point of view, the performance of the write path is more critical for us compared to the read path.  
Just in case there is a concept of 'multi-raft' already implemented by a number of projects. We may get it implemented on Ratis level, but I'm not sure how it's suitable in the embedded library concept and how much efforts it would require.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogReader.java
joshelser: Ratis has a directory in which is stores many LogSegments. Each of these are ordered with an Index, per the Raft protocol. Ratis can find the LogSegment containing the record at an Index using binary search and then does a linear search over that LogSegment to find the specific ID. I think that's sufficient for what we need in a first-pass.

I don't think we need to support a "recordId" that is different than the Raft Index. Just trying to give it a better name.

joshelser:
I think this is close. @mukul1987 @szetszwo @anuengineer any input from y'all? Maybe a merge if you think this is OK :)

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/RecordListener.java
```diff
@@ -0,0 +1,18 @@
+package org.apache.ratis.logservice.api;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Interface that, when registered with a {@link LogStream}, will receive all records written
+ * to that LogStream until it is removed.
+ */
+public interface RecordListener {
```
liuml07: Will more methods to be added? If not, is a simple `Consumer` sufficient as listeners (consumers)?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogWriter.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Appends each entry of data as a new record in the LogStream. If this method returns
+   * successfully, all records can be considered persisted. Otherwise, none can be assumed
+   * to have been written.
+   *
+   * @param records Records to append
+   * @return The largest recordId assigned to the records written
+   */
+  long writeMany(List<ByteBuffer> records) throws IOException;
```
liuml07: Is default implementation (via for-write) possible/necessary?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStreamConfiguration.java
```diff
@@ -0,0 +1,47 @@
+package org.apache.ratis.logservice.api;
```
liuml07: Apache License

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogName.java
```diff
@@ -0,0 +1,48 @@
+package org.apache.ratis.logservice.api;
```
liuml07: Apache License

ratis-logservice/src/main/java/org/apache/ratis/logservice/LogServiceFactory.java
```diff
@@ -0,0 +1,27 @@
+package org.apache.ratis.logservice;
```
liuml07: Apache License

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogName.java
```diff
@@ -0,0 +1,48 @@
...
+  // It's pretty likely that what uniquely defines a LogStream
+  // to change over time. We should account for this by making an
+  // API which can naturally evolve.
+  private final String name;
+
+  private LogName(String name) {
+    this.name = requireNonNull(name);
+  }
+
+  // Implementation detail -- we want uses to use the LogName as identifiable, not to
```
liuml07: s/Implementation detail/`@implSpec`?

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/RecordListener.java
joshelser: Good point. Given what we know now, I don't think we have more complex requirements. Switching this over to a consumer is a good idea, I think.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStreamConfiguration.java
joshelser: oops! :D

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogWriter.java
joshelser: Just calling doing a `for (ByteBuffer record : records) write(record);`, you mean? Could do that.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogName.java
joshelser: Nice! Didn't know about these. Happy to add that.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogWriter.java
liuml07: Yes. I assume we don't provide any transaction support, i.e. if one record writing fails, previous written record in the batch will have no guarantee (succeeded or reverted).

joshelser: Exactly.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogName.java
joshelser: Hrm, seems like these are proposed but not yet implemented? The maven-javadoc-plugin warns me against defining them.

```
Note: Custom tags that could override future standard tags:  @implSpec, @implNote. To avoid potential overrides, use at least one period character (.) in custom tag names.
```

VladRodionov:
Overall looks good for the first cut. Just one question and one concern:

Q: May be I missed that, but I did not find truncate log method. The only way to say to Log Service that we do need anymore records with ID <= X. 
C: Still think that adopting hierarchical naming scheme for log directory is a right thing.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogStreamConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.Map;
+import java.util.Map.Entry;
+
+/**
+ * An encapsulation of configuration for a LogStream.
+ */
+public interface LogStreamConfiguration {
```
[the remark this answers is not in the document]
VladRodionov: OK, got it.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogWriter.java
joshelser: Right, not intrinsically, but I think that's OK. Raft is making sure that when the method returns successfully, a majority of the nodes has the writes which were acked.

For any other case, the client would have to re-read the log to see which were actually applied. Specifically for HBase, I think that's acceptable -- if the write to the HBase WAL fails, that flows back to the client and the client needs to just resubmit the update.

I'm open to suggestions on how we could make this API better; my only hesitation is that I think I know too little to give the problem adequate consideration right now. I remember reading a blog by Jay Kreps about how hard exactly once is.

joshelser:
> Q: May be I missed that, but I did not find truncate log method. The only way to say to Log Service that we do need anymore records with ID <= X.

Nope, you're right. The API only lets you archive (or delete) an entire LogStream. I think this flows a bit more naturally with what we're doing inside of HBase already. We _could_ support intra-LogStream truncation and archival, but I think that would complicate things quite a lot.

joshelser:
>  C: Still think that adopting hierarchical naming scheme for log directory is a right thing.

Let's keep this in mind as we work on the LogStream administrative/management side. I think some hierarchy would make sense -- the notion of a namespace to separate different LogStreams using the same "storage directory" would be nice (DistributedLog has this concept).

Right now we just have a String (inside of `LogName`) to uniquely identify one LogStream from another. This was intentional to let us either change how `LogName`s compute uniqueness, or overlay some "logical" hierarchy using that arbitrary String.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogWriter.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Guarantees that all previous data appended by {@link #write(ByteBuffer)} are persisted
+   * and durable in the LogStream.
+   *
+   * @return The recordId prior to which all records are durable
+   */
+  long sync() throws IOException;
```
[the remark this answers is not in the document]
joshelser: I'm not 100% sure, but, given what I remember from the paper, I don't think we'd have to take a new snapshot. I would have to re-read.

ratis-logservice/src/main/java/org/apache/ratis/logservice/api/LogName.java
liuml07: I see. This seems not enabled by default. I found a change in other projects to enable it via `maven-javadoc-plugin` see: https://github.com/CodeFX-org/demo-javadoc-8-tags/blob/master/pom.xml?ts=4#L110-L133

It's nice to have. No intention to block this PR. :)

joshelser:
Closed as this was submitted already.

PR #9, opened 2019-01-09
RATIS-473. Clarify size and length methods on LogStream

joshelser:
getLength() was meant to be the number of records in a log
but was implemented as the size in bytes of the records in
the log. This commit adds a getSize() which takes the old
implementation of getLength(), and then reimplements
getLength() as the number of records in the log.

### DIFF
--- a/ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
+++ b/ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
@@ -62,11 +62,10 @@ public class LogStateMachine extends BaseStateMachine {
   }
 
   /*
-   *  State is a pair log's length and state (closed/open);
+   *  State is a log's length, size, and state (closed/open);
    */
-
+  private long size;
   private long length;
-
   private State state = State.OPEN;
 
   private final SimpleStateMachineStorage storage = new SimpleStateMachineStorage();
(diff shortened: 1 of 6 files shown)


joshelser:
One more, @vladrodionov and @chrajeshbabu

chrajeshbabu:
It's looking good @joshelser .

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
```diff
@@ -246,14 +250,21 @@ public StateMachineStorage getStateMachineStorage() {
    * @param msg message
    * @return reply message
    */
+  private CompletableFuture<Message> processGetSizeRequest(LogServiceRequestProto proto) {
+    GetLogLengthRequestProto msgProto = proto.getLengthQuery();
```
VladRodionov: Is it correct? Should not it be photo.getSizeRequest()?

VladRodionov:
You probably have some minor issue (see my comment inline in files changed), otherwise is looking good, @joshelser

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
joshelser: Hrm, that does look wrong :). Let me double check.

joshelser: Yup, definitely wrong. Wasn't causing a test failure because we don't need to pull anything off of the request object. Fixed it nonetheless.

PR #14, opened 2019-02-21
RATIS-478. Support custom grpc interceptors.

elek:
See: https://issues.apache.org/jira/browse/RATIS-478


adoroszlai:
@elek resolved the issue as won't fix (almost a year ago), I guess this PR can be closed, too.

PR #20, opened 2019-05-14
RATIS-546. Configurable value sizes on VerificationTool

joshelser:
Pads the log entries that are written to the given size.

The read path of verification still works as the extra padding is trimmed off (the padding itself is not verified)


chrajeshbabu:
Looks good to me @joshelser. +1

PR #22, opened 2019-05-30
RATIS-574. Configured sleep interval in RetryPolicy should not be used on NotLeaderException

swagle:
While doing performance tests with Ozone, it was observed that the client always waits for retry timeout when a NotLeaderException is logged. This bears a performance penalty for the client who cannot try the next Ratis server in the ring.


swagle:
cc: @lokeshj1703 / @mukul1987

PR #25, opened 2019-06-12
RATIS-588 LogStream StateMachine export

ankitsinghal:


ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
```diff
@@ -139,7 +164,12 @@ public void initialize(RaftServer server, RaftGroupId groupId,
     this.syncRequesTimer = metricRegistry.timer("syncRequesTime");
     this.appendRequestTimer = metricRegistry.timer("appendRequestTime");
     this.getCloseLogTimer = metricRegistry.timer("getCloseLogTime");
+    //archiving request time not the actual archiving time
+    this.archiveLogRequesTimer= metricRegistry.timer("archiveLogRequestTime");
```
joshelser: nit `archiveLogRequestTime`

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
```diff
@@ -452,29 +503,166 @@ public void close() {
...
 
 
-  private CompletableFuture<Message> processCloseLog(LogServiceRequestProto logServiceRequestProto) {
-    CloseLogRequestProto closeLog = logServiceRequestProto.getCloseLog();
+  private CompletableFuture<Message> processChangeState(LogServiceRequestProto logServiceRequestProto) {
+    LogServiceProtos.ChangeStateLogRequestProto changeState = logServiceRequestProto.getChangeState();
     // Need to check whether the file is opened if opened close it.
     // TODO need to handle exceptions while operating with files.
+    Throwable t = verifyState(State.OPEN);
+    this.state= State.valueOf(changeState.getState().name());
```
joshelser: We should have validation on the previous state and the state being requested to be moved to. For example, `ARCHIVED` to `ARCHIVING` is invalid.

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
```diff
@@ -452,29 +503,166 @@ public void close() {
...
+                }
+                records = 0;
+              }
+              records++;
+            }
+            writer.close();
+            if (!isNoMoreLeader) {
+              sendChangeStateRequest(State.ARCHIVED);
+            } else {
+              sendArchiveLogrequestToNewLeader(writer.getLastWrittenRecordId(), logName, location);
```
joshelser: In the case that the active leader of a LogServer quorum bounces around a lot, we're going to be "chaining" query() RPCs, aren't we? While we wait on computation of archive in "this" instance, we might trigger another archive in other instance, which triggers another archive in a third instance, etc. I'd worry about this cascading through the cluster in a leader election storm.

Would having the metadata quorum drive this archiving logic be a way to try to prevent that? Maybe there are more problems in trying to do that which I'm not thinking about?

ratis-logservice/src/main/java/org/apache/ratis/logservice/impl/ArchiveHdfsLogWriter.java
```diff
@@ -0,0 +1,91 @@
...
+    return 0;
+  }
+
+  @Override public void close() throws IOException {
+    os.close();
+    renameIfNewData();
+    hdfs.close();
+  }
+
+  @Override public void rollWriter() throws IOException {
```
joshelser: Thinking through this method. Is the intent that this will help prevent us from re-writing the same data multiple times in an leader change?

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
```diff
@@ -452,29 +503,166 @@ public void close() {
...
+    if (x.size() == 1) {
+      return x.get(0);
+    } else {
+      throw new GroupMismatchException(x.size() + " are group found for group id:" + raftGroupId);
+    }
+  }
+
+  @Override public void notifyNotLeader(Collection<TransactionContext> pendingEntries)
+      throws IOException {
+    isNoMoreLeader = true;
```
joshelser: Is it possible to "interrupt" the transactions which are pending? Could we keep a `Map<TransactionContext,Callable>` where the callable is doing the archiving? That might be more clean?

ratis-logservice/src/main/java/org/apache/ratis/logservice/shell/CommandFactory.java
```diff
@@ -46,6 +48,8 @@
     commands.put("quit", exitCommand);
     commands.put("help", new HelpCommand());
     commands.put("list", new ListLogsCommand());
+    commands.put("archive", new ArchiveLogCommand());
```
joshelser: So nice!!

ratis-logservice/src/test/java/org/apache/ratis/logservice/LogServiceReadWriteBase.java
```diff
@@ -166,6 +174,37 @@ public void testLogServiceReadWrite() throws Exception {
         assertEquals(expected, actual);
       }
       testJMXMetrics(logStream);
+      assertEquals(logStream.getState(),State.OPEN);
+
+      //In absense of MetaStateMachine, calling raft directly
```
joshelser: This is a nice test!

Would it be good to make a new unit test which does have both metadata and logserver quorums?

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
joshelser: That's a good point on scalability. Sounds good to try to prevent the bouncing leader problem :)

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
```diff
@@ -452,29 +510,194 @@ public void close() {
...
+            Thread.sleep(10000);
+          } catch (InterruptedException e) {
+          }
+          sendArchiveLogrequestToNewLeader(writer.getLastWrittenRecordId(), logName, location);
+        }
+        return true;
+      };
+      archiveFuture = executorService.submit(callable);
+    }
+    return CompletableFuture.completedFuture(
```
joshelser: Shouldn't we send back the `archiveFuture`, waiting on its completion?

ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
```diff
@@ -123,6 +129,41 @@ public void testReadWritetoLog() throws IOException, InterruptedException {
         assert(res.array().length > 0);
     }
 
+    @Test
+    public void testLogArchival() throws IOException, InterruptedException {
```
joshelser: Can you add some more tests, please?

Should have an e2e UT how this would work for users. Could set a local filesystem path (e.g. file:///..) and then use the LogServiceClient to trigger the archive of the log. Then, validate that the file can read from the archive'd log path.

Same goes for the export use-case.

ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
```diff
@@ -123,6 +129,41 @@ public void testReadWritetoLog() throws IOException, InterruptedException {
...
+        LogReader reader = logStream.createReader();
+        List<ByteBuffer> data = reader.readBulk(records.size());
+        assertEquals(records.size(), data.size());
+        reader.seek(1);
+        data = reader.readBulk(records.size());
+        assertEquals(records.size() - 1, data.size());
+
+        //Test ArchiveLogStream
+        LogServiceConfiguration config = LogServiceConfiguration.create();
+        LogStream archiveLogStream = new ArchiveLogStreamImpl(logName, config);
```
joshelser: We should have some API wrapper around this and not have users hit this Impl classes (unless this is specifically just testing the archive reader/write code). It wouldn't be a bad idea to have some code around these classes, just validating that they work (in isolation of the LogService Ratis daemons).

ratis-logservice/src/test/resources/logservice.xml
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+-->
+<configuration>
+  <property>
+    <name>logservice.archival.location</name>
+    <value>target/tmp/archive_1/</value>
```
joshelser: Ah! Tricky.

Nice to have this set here -- should we have individual tests overriding this so we can have each test keep their data around for debugging purposes?

ratis-logservice/src/main/java/org/apache/ratis/logservice/client/LogServiceClient.java
```diff
@@ -149,20 +169,58 @@ public void close() throws Exception {
...
+     * exporting of the log will happen asynchronously from the client,
+     * The call will return immediately after adding a request for archiving log
+     * to the respective quorum
+     *
+     * Client can check the status of export by calling getState() Method
+     *
+     * @param logName The name of the log to archive.
+     */
+    public void exportLog(LogName logName, String location, long recordId) throws IOException {
+        RaftClientReply archiveLogReply = getRaftClient(getLogInfo(logName)).sendReadOnly(
```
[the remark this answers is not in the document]
ankitsinghal: Actually.  this is log client (not meta client), so we need to create one as per the logName every time user want to do any meta operation on the log. 
sure, let me close after the operation is finished.

ratis-logservice/src/main/java/org/apache/ratis/logservice/impl/ArchiveHdfsLogWriter.java
```diff
@@ -0,0 +1,95 @@
...
+  private long currentRecordId;
+  private long lastRollRecordId;
+
+  @Override
+  public void init(String location, LogName logName) throws IOException {
+    Configuration configuration = new Configuration();
+    hdfs = FileSystem.get(configuration);
+    Path loc = new Path(location);
+    if(!hdfs.exists(loc)){
+      hdfs.mkdirs(loc);
```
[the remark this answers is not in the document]
ankitsinghal: Actually, it's a bug, processArchiveLog should not start always from the beginning, it should take the recordId passed in the request. Let me fixed it

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
```diff
@@ -452,29 +510,194 @@ public void close() {
...
+    long recordId = archiveLog.getLastArchivedRaftIndex();
+    t = verifyState(State.CLOSED);
+    if (t == null) {
+      Callable<Boolean> callable = () -> {
+        sendChangeStateRequest(State.ARCHIVING);
+        updateArchivingInfo(recordId, logName, location, isArchival);
+        ArchiveLogWriter writer = new ArchiveHdfsLogWriter();
+        writer.init(archiveLocationForLog, logName);
+        LogServiceRaftLogReader reader = new LogServiceRaftLogReader(log);
+        reader.seek(0);
```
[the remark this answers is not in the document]
ankitsinghal: agreed, does it resolve your above doubt?

> Leader1 writes a segment 0->100.
> Leader1 starts writing the next segment 101->...
> Leader1 pauses, Leader2 takes over
> Leader1 sends archive request to Leader2
>

ratis-logservice/src/main/java/org/apache/ratis/logservice/impl/ArchiveHdfsLogReader.java
```diff
@@ -0,0 +1,194 @@
...
+  private void loadNext() throws IOException {
+    int length;
+    try {
+      length = readLength();
+    } catch (EOFException e) {
+      currentRecord = null;
+      return;
+    }
+    byte[] bytes = new byte[length];
+    is.read(bytes);
```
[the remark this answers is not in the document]
ankitsinghal: agreed.
Throwing this now:-
if (is.read(bytes) != length) {
      throw new EOFException(
          "File seems to be corrupted, Encountered EOF before reading the complete record");
    }

ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
ankitsinghal: yes, we already had the API for the same in LogServiceClient.
LogStream archiveLogStream = client.getArchivedLog(logName);

Let me add some unit test around this implementation.

ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
ankitsinghal: sure let me add some more tests, but this is actually e2e test, probably you have seen this config later.
https://github.com/apache/incubator-ratis/pull/25#discussion_r299662354

ratis-logservice/src/test/resources/logservice.xml
ankitsinghal: Let me see if I can do that, not sure if we need to restart the cluster if we change the config in the same test class.

ratis-logservice/src/main/java/org/apache/ratis/logservice/impl/ArchiveHdfsLogReader.java
```diff
@@ -0,0 +1,194 @@
...
+
+  @Override public byte[] next() throws IOException {
+    byte[] current = currentRecord;
+    currentRecord = null;
+    loadNext();
+    return current;
+  }
+
+  @Override public long getCurrentRaftIndex() {
+    return 0;
```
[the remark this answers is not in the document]
ankitsinghal: Done

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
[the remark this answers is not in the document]
joshelser: Yup! Thanks.

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
ankitsinghal: As archival is asynchronous, if we do so, woudn't the client still wait until the task completes?

ratis-logservice/src/test/resources/logservice.xml
ankitsinghal: I tested archival through some exports by providing different locations.

ankitsinghal:
Thanks a lot, @joshelser for the detailed review.  WDYT about the current state now?

ratis-logservice/src/test/resources/logservice.xml
```diff
@@ -0,0 +1,28 @@
...
+    <description>Log service meta-server port number</description>
```
joshelser: nit: invalid description

ratis-logservice/src/test/java/org/apache/ratis/logservice/impl/TestArchiveHdfsLogReaderAndWriter.java
```diff
@@ -0,0 +1,132 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class TestArchiveHdfsLogReaderAndWriter {
+  static MiniDFSCluster cluster;
+  static Configuration conf;
+
+  @BeforeClass public static void setup() throws IOException {
+    conf = new Configuration();
+    cluster = new MiniDFSCluster.Builder(conf).numDataNodes(3).build();
```
joshelser: Needs `@AfterClass` to stop the cluster

ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
```diff
@@ -126,6 +130,98 @@ public void testReadWritetoLog() throws IOException, InterruptedException {
         assert(res.array().length > 0);
     }
 
+    @Test
+    public void testLogArchival() throws IOException, InterruptedException {
```
joshelser: I really like this test (and `testLogExport`)!

joshelser:
UT's pass locally for me too. I think this is just about ready. If Ankit can get the rest of the minor stuff I covered in my last review updated, I think this is ready to go.

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/ArchivalInfo.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.apache.ratis.logservice.server;
+
+import org.apache.ratis.logservice.api.LogName;
+import org.apache.ratis.logservice.proto.LogServiceProtos;
+import org.apache.ratis.logservice.util.LogServiceProtoUtil;
+
+public class ArchivalInfo {
+
+  public enum ArchivalStatus{
```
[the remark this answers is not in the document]
ankitsinghal: At INTERRUPTED, we are submitting an archival request again to a leader and even if this call fails , we will change the archival status to FAILED and Log state to CLOSED again so that user can submit the request again.
Added these comments in the code as well

ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
ankitsinghal: Thanks :)

ratis-logservice/src/main/java/org/apache/ratis/logservice/client/LogServiceClient.java
```diff
@@ -149,20 +186,62 @@ public void close() throws Exception {
...
+     * The call will return immediately after adding a request for archiving log
+     * to the respective quorum
+     *
+     * Client can check the status of export by calling getState() Method
+     *
+     * @param logName The name of the log to archive.
      */
-    void archiveLog(LogName name) throws IOException {
-      // TODO: write me
+    public void exportLog(LogName logName, String location, long recordId) throws IOException {
```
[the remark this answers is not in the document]
ankitsinghal: An important difference is:-
Archival is allowed only once and the log is still readable (from the defined location ) using the same quorum even if the raft log is truncated. so the constraints like recordId =0 and location should be defined at server.

but export is allowed as many times (may have use-cases for user snapshot or backup). Exported data cannot be read from quorum and has to be read directly with the location.

ankitsinghal:
Thanks @joshelser , Last commit should have the review comments taken care of.

joshelser:
b618849 looks perfect! Thanks Ankit!

PR #27, opened 2019-07-11
RATIS-621. Metrics subproject is not included in the release tar

elek:


joshelser:
Good catch, Marton!

adoroszlai:
@lokeshj1703

PR #31, opened 2019-08-23
RATIS-667 Provide a visual for the LogService lifecycle

cbaenziger:
This adds a visual for the LogService [lifecycle](https://ratis.incubator.apache.org/logservice/lifecycle.html) documentation.


cbaenziger:
I provided an update where the log states are capitalized to match the documentation's capitalization.

PR #32, opened 2019-08-27
RATIS-667 Provide a visual for the LogService lifecycle

joshelser:
@cbaenziger had to tweak your PR a little. Does this look OK to you?

Co-authored-by: Josh Elser <elserj@apache.org>


joshelser:
> BTW should I add the dot command to the build.sh
Oh, that's not a bad idea (if the command exists). Let me try to add that.

joshelser:
> Let me try to add that.

Not super feasible because it will clobber the ASF license header. I don't see an obvious way to re-add the license via the DOT file itself, so I'm just skipping that part.

But, there is some good improvements coming into `build.sh` as a result of me messing around here.

joshelser:
Can you take a look again, @cbaenziger ?

static/lifecycle.svg
```diff
@@ -0,0 +1,179 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+ "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<!--
```
cbaenziger: Nit: To build the SVG in the build script, would you want to use `sed -i` (OS X needs a bit of [finagling](https://unix.stackexchange.com/a/131940) for `sed -i`) to inject the license looking for the string `Generated by graphviz version` for now? Or I could write an XSLT to be a bit more pedantic if you'd like?

joshelser: Ahh, tricky. I couldn't think of a way to do it via my normal text-replacement tools, but your sed trick sounds good to me.

If you have it handy, i'm happy to pull it in, else I'll try to come up with something myself.

joshelser:
OK, 261e8a1 can regenerate the license header on OSX. Let me test on Linux too.

joshelser:
15ca39f fixes the issue of building on Linux. Going to squash/rebase this one as I think this will be good for what you were looking for :)

If the user doesn't have `dot` on the `$PATH`, we won't re-generate the SVG and everything else should continue as expected.

build.sh
```diff
@@ -17,6 +17,14 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 set -e
 mkdir -p build
+
+# Usage
+if [[ $# -ne 1 ]]; then
+  echo "Usage: ./build.sh <website_output>"
```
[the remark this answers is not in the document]
joshelser: Just noticed I did this backwards in the commit. Fixing.

joshelser:
Merged!

ankitsinghal:
oh great!! I didn't know that a framework like Graphviz exists. Visualization looks great. Thanks for working on this.
https://ratis.incubator.apache.org/logservice/lifecycle.html

PR #37, opened 2019-09-27
RATIS-691. Fix checkstyle violations in ratis-logservice

dineshchitlangia:


ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
```diff
@@ -331,7 +331,7 @@ public StateMachineStorage getStateMachineStorage() {
     LogServiceProtos.GetExportInfoReplyProto.Builder exportBuilder =
         LogServiceProtos.GetExportInfoReplyProto.newBuilder();
     exportMap.values().stream().map(
-        archivalInfo -> exportBuilder.addInfo(LogServiceProtoUtil.toExportInfoProto(archivalInfo)))
+        archInfo -> exportBuilder.addInfo(LogServiceProtoUtil.toExportInfoProto(archInfo)))
```
joshelser: What's the checkstyle rule this is failing on?

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/MetaStateMachine.java
```diff
@@ -291,7 +290,8 @@ public TransactionContext preAppendTransaction(TransactionContext trx) throws IO
                 MetaServiceProtoUtil.toDeleteLogReplyProto().toByteString()));
     }
 
-//    private CompletableFuture<Message> processCloseLog(MetaServiceProtos.MetaServiceRequestProto logServiceRequestProto) {
```
joshelser: Please just delete commented out code instead of fixing it.

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
dineshchitlangia: {{archivalInfo}} is hiding another private member of the class

ratis-logservice/src/main/java/org/apache/ratis/logservice/server/MetaStateMachine.java
dineshchitlangia: Sure, I will post a revision shortly.

dineshchitlangia:
@joshelser Thanks for taking time to review this PR. There were a few comments from @szetszwo too. I have addressed review comments from both of you in the recent commit.

PR #40, opened 2019-10-16
RATIS-390. Fix java package structure.

cxorm:
This patch modifies the package of LogServiceClient as description of RATIS-390.
But the package of LogStateMachine is changed to server, so I don't modify it.


joshelser:
Thanks for the PR, @cxorm.

I'm seeing that unit tests are currently broken for the logservice, so i'm trying to figure out what broke those before merging this in.

PR #51, opened 2020-02-04
RATIS-807. ratis-tools subproject is not included in the release jar.

xiaoyuyao:
https://issues.apache.org/jira/browse/RATIS-807

How is this tested:

Test with the release script dev-support/make-rc.sh prepare_src and prepare_bin

prepare_bin will fail without the change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -260,6 +260,18 @@
        <scope>test</scope>
       </dependency>
       <dependency>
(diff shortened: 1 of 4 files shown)


xiaoyuyao:
Thanks @arp7 for the review. I will merge it shortly.

PR #52, opened 2020-02-05
RATIS-808. Ratis top plevel pom.xml misses distributionManagement and…

xiaoyuyao:
What is the issue:
Ratis top plevel pom.xml misses distributionManagement and repositories for release. As a result, the repository id does not match the maven server id specified in the ~/.m2/settings and always got HTTP 401 error returned even though the user/passwd are correct as found INFRA-19819. This ticket is opened to fix it. 

JIRA:
https://issues.apache.org/jira/browse/RATIS-808

How is this tested:
./dev-tools/make_rc.sh publish-mvn succeeded

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,6 +33,34 @@
   <url>https://ratis.apache.org/</url>
   <inceptionYear>2017</inceptionYear>
 
+  <distributionManagement>
+    <repository>
+      <id>${distMgmtStagingId}</id>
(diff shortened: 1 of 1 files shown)


xiaoyuyao:
Thanks @arp7  for the review. I've merged the change to master. Will open separate JIRAs to fix the ratis jenkins issue which have been broken for a while.

PR #56, opened 2020-03-02
RATIS-821. Fix high processor load for ScheduledThreadPoolExecutor with 0 core threads

runzhiwang:
1. This bug is related to the bug of JDK8: https://bugs.openjdk.java.net/browse/JDK-8129861. When create a thread pool with size 0,  it will causing 100% CPU load forever when schedule a shutdown task.
2. This bug was triggered in hadoop-ozone, the related PR is: https://github.com/apache/hadoop-ozone/pull/611. The triggered stack is as the image shows.
image


runzhiwang:
@mukul1987 Could you help review this patch ? Thank you very much.

PR #58, opened 2020-04-06
RATIS-832. Add Metrics for retry cache

runzhiwang:
I did not implement metric for cache size in bytes. Because `com.google.common.cache.Cache` does not  have the API to get the cache size. If I implement it myself,  it will not be accurate, because the cache size depends on the data structure used by `com.google.common.cache.Cache`.


runzhiwang:
@bshashikant Could you help review this patch ? Thank you very much.

bshashikant:
Thanks @runzhiwang . PR model doesn't yet work seamlessly in RATiS. Can you update the patch
on the jira itself?

runzhiwang:
@bshashikant Yes, I have updated the patch on the jira.

PR #61, opened 2020-04-24
RATIS-879. Fix TestRaftAsyncWithGrpc#testNoRetryWaitOnNotLeaderException

lokeshj1703:
Apache jira:
https://issues.apache.org/jira/browse/RATIS-879


elek:
Sorry for the noise. We had some issue with the pr build. I added the missing commits as an addendum to RATIS-697.

PR #65, opened 2020-04-27
RATIS-881. Fix Failed unit test because test before MiniRaftCluster ready

runzhiwang:
**What's the probelm ?**
image

**What's the reason ?**
The root cause is [RaftServerMetrics::getPeerCommitIndexGauge](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerMetrics.java#L141) happens before [RaftServerMetrics::addPeerCommitIndexGauge](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerMetrics.java#L122).  
When some RaftServerImpl [setRole(RaftPeerRole.LEADER, "changeToLeader")](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java#L345), the statement [waitForLeader](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java#L446) succ to get leader and test begin, but [role.startLeaderState](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java#L349) ->
 [new LeaderState](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/impl/RoleInfo.java#L94) ->
[LeaderState::addSenders](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderState.java#L409)->[RaftServerMetrics::addFollower](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerMetrics.java#L106) -> [RaftServerMetrics::addPeerCommitIndexGauge](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerMetrics.java#L122) has not finished.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -460,6 +460,20 @@ public class RaftServerImpl implements RaftServerProtocol, RaftServerAsynchronou
     return null;
   }
 
+  @VisibleForTesting
+  public boolean isLeaderReady() {
+    if (!isLeader()) {
+      return false;
+    }
+
+    final LeaderState leaderState = role.getLeaderState().orElse(null);
+    if (leaderState == null || !leaderState.isReady()) {
+      return false;
+    }
+
+    return true;
+  }
+
   NotLeaderException generateNotLeaderException() {
     if (lifeCycle.getCurrentState() != RUNNING) {
       return new NotLeaderException(getMemberId(), null, null);
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -447,8 +447,8 @@ public class RaftServerImpl implements RaftServerProtocol, RaftServerAsynchronou
       final RaftClientReply reply = new RaftClientReply(request, exception, getCommitInfos());
       return RetryCache.failWithReply(reply, entry);
     }
-    final LeaderState leaderState = role.getLeaderState().orElse(null);
-    if (leaderState == null || !leaderState.isReady()) {
+
+    if (!isLeaderReady()) {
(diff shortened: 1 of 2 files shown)


ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -460,6 +460,20 @@ public String toString() {
...
+    if (!isLeader()) {
+      return false;
+    }
+
+    final LeaderState leaderState = role.getLeaderState().orElse(null);
+    if (leaderState == null || !leaderState.isReady()) {
+      return false;
+    }
+
+    return true;
```
lokeshj1703, line 474: We can simplify this as.
`return leaderState != null && leaderState.isReady();`

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -460,6 +460,20 @@ public String toString() {
     return null;
   }
 
+  @VisibleForTesting
+  public boolean isLeaderReady() {
+    if (!isLeader()) {
```
lokeshj1703, line 465: We can remove this check as leaderState == null covers the case and use this function in checkLeaderState line 451 as well.

runzhiwang:
@lokeshj1703 Thanks for your review. I have updated the patch.

lokeshj1703:
@runzhiwang Thanks for the contribution! @bshashikant Thanks for the review! I have committed the PR to master branch. 
Made a minor change in the commit: Removed @VisibleForTesting annotation on RaftServerImpl@isLeaderReady.

PR #66, opened 2020-04-27
RATIS-884. Fix Failed UT: TestRaftLogMetrics.testRaftLogMetrics

runzhiwang:
**What's the problem ?**
image

**What's the reason ?**
As the images shows, `flushCount.incrementAndGet() `happens in the 1st statement: `stateMachine.flushStateMachineData(lastWrittenIndex)`, `ratisMetricRegistry.get(RAFT_LOG_FLUSH_TIME)` increase after the 2nd statement: `timerContext.stop()`. If the test `Assert.assertEquals(expectedFlush, tm.getCount())` happens between 1st and 2nd statement, then the expectedFlush will be tm.getCount() + 1, so the test fail.

image

**How to fix ?**
Retry check expectedFlush == tm.getCount()

### DIFF
--- a/ratis-test/src/test/java/org/apache/ratis/server/raftlog/TestRaftLogMetrics.java
+++ b/ratis-test/src/test/java/org/apache/ratis/server/raftlog/TestRaftLogMetrics.java
@@ -127,7 +127,11 @@ public class TestRaftLogMetrics extends BaseTest
     final MetricsStateMachine stateMachine = MetricsStateMachine.get(server);
     final int expectedFlush = stateMachine.getFlushCount();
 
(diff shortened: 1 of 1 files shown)


lokeshj1703:
@runzhiwang We don't see this failure in latest runs? Do you think this was fixed by RATIS-881?

runzhiwang:
@lokeshj1703  It can not be fixed by RATIS-881, it was caused by UT and Ratis Use different thread. And I test it again, it still happen.  To easy reproduce, you can add Thead.sleep(1000) to delay the timerContext.stop(), it will happen.
image

lokeshj1703:
@runzhiwang  Thanks for the contribution! I have merged the PR to master branch.

PR #76, opened 2020-04-30
RATIS-845. Fix memory leak of RaftServerImpl  for no unregister from reporter

runzhiwang:
**What's the problem ?**
The related ozone PR: https://github.com/apache/hadoop-ozone/pull/885

As the image shows, there are 1885 instances of  RaftServerImpl, most of them are Closed, and should be GC, but actually not. You can find from the image 
 1513 RaftServerImpl were held by ManagermentFactory->jxmMBeanServer->HashMap, 372 RaftServerImpl were held by Datanode ReportManager Thread -> prometheus -> HashMap. So 1513 RaftServerImpl leak in ratis, and 372 leak in ozone. If RaftServerImpl can not GC, there are a lot of related resource can not be GC, such as the [DirectByteBuffer](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogWorker.java#L150)  in SegmentRaftLogWorker, which result 1GB memory leak out of heap.

The reason RaftServerImpl held by ManagermentFactory->jxmMBeanServer->HashMap is ratis start [JmxReporter](https://github.com/apache/incubator-ratis/blob/master/ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java#L47), but does not stop it.

The reason RaftServerImpl held by Datanode ReportManager Thread -> prometheus -> HashMap is ozone call the ratis function to  [register](https://github.com/apache/hadoop-ozone/blob/master/hadoop-hdds/container-service/src/main/java/org/apache/hadoop/ozone/HddsDatanodeService.java#L189) metric in prometheus, but does not unregister it.

**1. 1885 instances of RaftServerImpl**
image

**2. 1513 RaftServerImpl were held by ManagermentFactory->jxmMBeanServer->HashMap, 372 RaftServerImpl were held by Datanode ReportManager Thread -> prometheus -> HashMap**
image

**3. 1513 RaftServerImpl were held by ManagermentFactory->jxmMBeanServer->HashMap**
image

**4. 372 RaftServerImpl were held by Datanode ReportManager Thread -> prometheus -> HashMap**
image

5. 2038 DirectByteBuffer, and 1885 held by RaftServerImpl
image

6. 1033 DirectByteBuffer were held by ManagermentFactory, 802 DirectByteBuffer were held by Datanode ReportManager Thread, total 1885.
image

### DIFF
--- a/ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
+++ b/ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
@@ -34,9 +34,21 @@ public final class MetricsReporting {
   }
 
   public static Consumer<RatisMetricRegistry> consoleReporter(TimeDuration rate) {
-    return ratisMetricRegistry -> ConsoleReporter.forRegistry(ratisMetricRegistry.getDropWizardMetricRegistry())
-        .convertRatesTo(TimeUnit.SECONDS).convertDurationsTo(TimeUnit.MILLISECONDS).build()
-        .start(rate.getDuration(), rate.getUnit());
+    return ratisMetricRegistry -> {
+      ConsoleReporter reporter = ConsoleReporter.forRegistry(ratisMetricRegistry.getDropWizardMetricRegistry())
+          .convertRatesTo(TimeUnit.SECONDS).convertDurationsTo(TimeUnit.MILLISECONDS).build();
+      reporter.start(rate.getDuration(), rate.getUnit());
+      ratisMetricRegistry.setConsoleReporter(reporter);
+    };
+  }
+
+  public static Consumer<RatisMetricRegistry> stopConsoleReporter() {
+    return ratisMetricRegistry -> {
+      ConsoleReporter reporter = ratisMetricRegistry.getConsoleReporter();
+      if (reporter != null) {
+        reporter.stop();
+      }
+    };
   }
 
   public static Consumer<RatisMetricRegistry> jmxReporter() {
@@ -44,7 +56,19 @@ public final class MetricsReporting {
       Builder builder =
           JmxReporter.forRegistry(registry.getDropWizardMetricRegistry());
       builder.inDomain(registry.getMetricRegistryInfo().getApplicationName());
-      builder.build().start();
+      JmxReporter reporter = builder.build();
+      reporter.start();
+
+      registry.setJmxReporter(reporter);
+    };
+  }
+
+  public static Consumer<RatisMetricRegistry> stopJmxReporter() {
+    return registry -> {
+      JmxReporter reporter = registry.getJmxReporter();
+      if (reporter != null) {
+        reporter.stop();
+      }
     };
   }
 }
--- a/ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
+++ b/ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
@@ -46,7 +46,7 @@ public final class MetricsReporting {
     return ratisMetricRegistry -> {
       ConsoleReporter reporter = ratisMetricRegistry.getConsoleReporter();
       if (reporter != null) {
-        reporter.stop();
+        reporter.close();
       }
     };
   }
@@ -67,7 +67,7 @@ public final class MetricsReporting {
     return registry -> {
       JmxReporter reporter = registry.getJmxReporter();
       if (reporter != null) {
-        reporter.stop();
+        reporter.close();
       }
     };
   }
--- a/ratis-metrics/src/main/java/org/apache/ratis/metrics/impl/RatisMetricRegistryImpl.java
+++ b/ratis-metrics/src/main/java/org/apache/ratis/metrics/impl/RatisMetricRegistryImpl.java
@@ -32,7 +32,6 @@ import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.MetricRegistry.MetricSupplier;
 import com.codahale.metrics.MetricSet;
 import com.codahale.metrics.Timer;
-import io.prometheus.client.Collector;
 import org.apache.ratis.metrics.MetricRegistryInfo;
 import org.apache.ratis.metrics.RatisMetricRegistry;
 import org.apache.ratis.thirdparty.com.google.common.annotations.VisibleForTesting;
@@ -47,7 +46,6 @@ public class RatisMetricRegistryImpl implements RatisMetricRegistry {
 
   private JmxReporter jmxReporter;
   private ConsoleReporter consoleReporter;
-  private Collector prometheusCollector;
 
   public RatisMetricRegistryImpl(MetricRegistryInfo info) {
     super();
@@ -142,15 +140,5 @@ public class RatisMetricRegistryImpl implements RatisMetricRegistry {
   @Override
   public ConsoleReporter getConsoleReporter() {
     return this.consoleReporter;
-  }
-
-  @Override
-  public void setPrometheusCollector(Collector prometheusCollector) {
-    this.prometheusCollector = prometheusCollector;
-  }
-
-  @Override
-  public Collector getPrometheusCollector() {
-    return this.prometheusCollector;
   }
 }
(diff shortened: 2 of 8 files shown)


bshashikant:
@avijayanhwx, can you plz look at this ?

ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
```diff
@@ -34,17 +34,41 @@ private MetricsReporting() {
...
-        .start(rate.getDuration(), rate.getUnit());
+    return ratisMetricRegistry -> {
+      ConsoleReporter reporter = ConsoleReporter.forRegistry(ratisMetricRegistry.getDropWizardMetricRegistry())
+          .convertRatesTo(TimeUnit.SECONDS).convertDurationsTo(TimeUnit.MILLISECONDS).build();
+      reporter.start(rate.getDuration(), rate.getUnit());
+      ratisMetricRegistry.setConsoleReporter(reporter);
+    };
+  }
+
+  public static Consumer<RatisMetricRegistry> stopConsoleReporter() {
```
avijayanhwx, line 45: Do we need to add a stop method for every reporter? Maybe we can rely on the generic close() method implemented by all reports (Closeable) and they already stop() in the close() method?

ratis-metrics/src/main/java/org/apache/ratis/metrics/impl/RatisMetricRegistryImpl.java
```diff
@@ -116,4 +123,34 @@ private String getMetricName(String shortName) {
...
+    this.consoleReporter = consoleReporter;
+  }
+
+  @Override
+  public ConsoleReporter getConsoleReporter() {
+    return this.consoleReporter;
+  }
+
+  @Override
+  public void setPrometheusCollector(Collector prometheusCollector) {
```
avijayanhwx, line 148: Are we doing this to support a prometheus collector purely in Ratis, instead of depending on an external client like Ozone?

avijayanhwx:
cc @elek since he added the reporter registrations.

ratis-metrics/src/main/java/org/apache/ratis/metrics/impl/RatisMetricRegistryImpl.java
runzhiwang: @avijayanhwx Thanks for review. Actually, I also do not want to add prometheus in Ratis.
But because ozone pass a lambda to ratis to register collector in prometheus, and if we want to unregister the collector when remove registry by `CollectorRegistry.defaultRegistry.unregister(collector)`, we must record the collector in Ratis, if you have better idea, plz share it.  The following code copy from ozone related PR: https://github.com/apache/hadoop-ozone/pull/885.
image

ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
runzhiwang: @avijayanhwx Yes, we can close() JmxReporter and ConsoleReporter. But for prometheus, it only has unregister, and also need a parameter i.e.collection. I think prometheus make the problem complicated.

image

ratis-metrics/src/main/java/org/apache/ratis/metrics/impl/RatisMetricRegistryImpl.java
runzhiwang: @avijayanhwx Thanks for review. I have remove prometheus in Ratis, I handle this in ozone.

ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
runzhiwang: @avijayanhwx For ConsoleReporter and JmxReporter we can call close. But for prometheus in ozone, there is only unregister() method. The following code copied from https://github.com/apache/hadoop-ozone/pull/885.
image

avijayanhwx: Thank you @runzhiwang. Can we use the close method now, given that you have removed the prometheus reporter registrations? Is that OK with you?

runzhiwang: @avijayanhwx Hi, thanks for review. Even though I have removed prometheus in ratis, but it exists in ozone. Such as the following code. ozone call `CollectorRegistry.defaultRegistry.register` to register registry in prometheus, and it also need call `CollectorRegistry.defaultRegistry.unregister` to unregister it. There is no close method for prometheus, and the unregister also need a parameter which differs from Reporter::close.

```
    MetricRegistries.global().addReporterRegistration(
        registry -> CollectorRegistry.defaultRegistry.register(
            new RatisDropwizardExports(
                registry.getDropWizardMetricRegistry())),
        registry -> CollectorRegistry.defaultRegistry.unregister(
            new RatisDropwizardExports(
                registry.getDropWizardMetricRegistry())));
```

avijayanhwx: @runzhiwang Thanks, but IMHO, given that the JMX and console reporters' implementation are present in Ratis code itself, we can expose an APIs that look like enableJmxReporter(), enableConsoleReporters(), and the current addReporterRegistration API can be used for external reporters like Prometheus. This keeps a client code simple, but provides scope for adding custom reporters.

runzhiwang: @avijayanhwx Thanks for explanation. I agree with you. I have exposed enableJmxReporter(), enableConsoleReporters().

avijayanhwx: Thanks @runzhiwang.

ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
```diff
@@ -34,17 +34,41 @@ private MetricsReporting() {
...
 
   public static Consumer<RatisMetricRegistry> consoleReporter(TimeDuration rate) {
-    return ratisMetricRegistry -> ConsoleReporter.forRegistry(ratisMetricRegistry.getDropWizardMetricRegistry())
-        .convertRatesTo(TimeUnit.SECONDS).convertDurationsTo(TimeUnit.MILLISECONDS).build()
-        .start(rate.getDuration(), rate.getUnit());
+    return ratisMetricRegistry -> {
+      ConsoleReporter reporter = ConsoleReporter.forRegistry(ratisMetricRegistry.getDropWizardMetricRegistry())
+          .convertRatesTo(TimeUnit.SECONDS).convertDurationsTo(TimeUnit.MILLISECONDS).build();
+      reporter.start(rate.getDuration(), rate.getUnit());
+      ratisMetricRegistry.setConsoleReporter(reporter);
```
avijayanhwx, line 41: Can you tell me why we need this change? Instead of returning a function that starts the console reporter for a given RatisMetricRegistry, why are we starting and using a setter?
      `ratisMetricRegistry.setConsoleReporter(reporter);`

ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
```diff
@@ -34,17 +34,41 @@ private MetricsReporting() {
...
   public static Consumer<RatisMetricRegistry> jmxReporter() {
     return registry -> {
       Builder builder =
           JmxReporter.forRegistry(registry.getDropWizardMetricRegistry());
       builder.inDomain(registry.getMetricRegistryInfo().getApplicationName());
-      builder.build().start();
+      JmxReporter reporter = builder.build();
+      reporter.start();
+
+      registry.setJmxReporter(reporter);
```
avijayanhwx, line 62: Same here.

ratis-metrics/src/main/java/org/apache/ratis/metrics/MetricsReporting.java
runzhiwang: @avijayanhwx Because when we stop the reporter, we must get it from registry, so we need to set it in registry when create. Such as the following code, we need to getConsoleReporter from registry.
```
return ratisMetricRegistry -> {
      ConsoleReporter reporter = ratisMetricRegistry.getConsoleReporter();
      if (reporter != null) {
        reporter.close();
      }
    };
```

avijayanhwx:
Thanks @runzhiwang. I am +1 for this patch.

avijayanhwx:
> Thanks @runzhiwang. I am +1 for this patch.

My vote is non-binding since I am not a committer. cc @bshashikant.

PR #78, opened 2020-04-30
RATIS-833. Add metrics for raft log cache count and size in bytes

runzhiwang:
@bshashikant Could you help review it ? Thank you very much.


bshashikant:
Thanks @runzhiwang for working on this. We can potentially large no of closed segments in ratis pipeline lifecycle if log purging is not enabled. I am just thinking if it would be better to maintain the total segment size in memory instead of iterating over all to figure the size every-time to fetch the metric value. What do you think?

bshashikant:
@runzhiwang , can you please update the PR?

runzhiwang:
@bshashikant updated

PR #81, opened 2020-04-30
RATIS-889. Fix TestRaftSnapshotWithGrpc and TestRaftSnapshotWithSimulatedRpc

adoroszlai:

Fix TestRaftSnapshotWithGrpc and TestRaftSnapshotWithSimulatedRpc: `StateMachineUpdater` [unregisters metrics from the registry](https://github.com/apache/incubator-ratis/blob/56e2073515804cc70f64b215f971d547070ed38f/ratis-server/src/main/java/org/apache/ratis/server/impl/StateMachineUpdater.java#L129) during stop, so the timer after restart is not the same instance as before.  Thus, to verify that snapshot is taken during stop, we need to obtain the pre-stop timer.

Also verify the metric right after snapshot file is created.

https://issues.apache.org/jira/browse/RATIS-889

## How was this patch tested?

Executed `TestRaftSnapshot*` multiple times.  Also passed in CI:

https://github.com/adoroszlai/incubator-ratis/runs/633049769


runzhiwang:
LGTM. +1

adoroszlai:
Thanks @runzhiwang for the review, and @bshashikant for reviewing and committing this.

PR #83, opened 2020-05-01
RATIS-901. Fix Failed UT: WatchRequestTests.testWatchRequestClientTimeout

runzhiwang:
**Why WatchRequestTests.testWatchRequestClientTimeout failed ?**
The test try to watch a log index i.e. 1000 which will never be committed, so client can not receive reply, and connection closed throw AlreadyClosedException, [trigger TimeoutIOException](https://github.com/apache/incubator-ratis/blob/master/ratis-grpc/src/main/java/org/apache/ratis/grpc/client/GrpcClientProtocolClient.java#L325).

But when throw AlreadyClosedException, suggested new leader is null, because connection has been closed, so RaftClientImpl::handleIOException will [random select](https://github.com/apache/incubator-ratis/blob/master/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java#L418) a new leader, so it maybe sometimes [throw NotLeaderException](https://github.com/apache/incubator-ratis/blob/master/ratis-grpc/src/main/java/org/apache/ratis/grpc/client/GrpcClientProtocolClient.java#L278) rather than TimeoutIOException.
企业微信截图_9fed55fa-5428-4ab2-a82f-f637791e9862

### DIFF
--- a/ratis-server/src/test/java/org/apache/ratis/WatchRequestTests.java
+++ b/ratis-server/src/test/java/org/apache/ratis/WatchRequestTests.java
@@ -23,6 +23,7 @@ import org.apache.ratis.client.RaftClientConfigKeys;
 import org.apache.ratis.conf.RaftProperties;
 import org.apache.ratis.proto.RaftProtos.CommitInfoProto;
 import org.apache.ratis.proto.RaftProtos.ReplicationLevel;
+import org.apache.ratis.protocol.NotLeaderException;
 import org.apache.ratis.protocol.NotReplicatedException;
 import org.apache.ratis.protocol.RaftClientReply;
 import org.apache.ratis.protocol.RaftRetryFailureException;
@@ -446,6 +447,8 @@ public abstract class WatchRequestTests<CLUSTER extends MiniRaftCluster>
     final Logger LOG = p.log;
 
     CompletableFuture<RaftClientReply> watchReply;
+    // watch 1000 which will never be committed
+    // so client can not receive reply, and connection closed, throw TimeoutException
     watchReply = p.sendWatchRequest(1000);
 
     try {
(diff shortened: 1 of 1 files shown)


ratis-server/src/test/java/org/apache/ratis/WatchRequestTests.java
```diff
@@ -457,8 +460,13 @@ static void runTestWatchRequestClientTimeout(TestParameters p) throws Exception
           ex.getCause().getClass());
       if (ex.getCause() != null) {
         if (ex.getCause().getCause() != null) {
-          Assert.assertEquals(TimeoutIOException.class,
-              ex.getCause().getCause().getClass());
+          // when client closed and throw TimeoutException,
+          // RaftClientImpl::handleIOException will random select a leader
+          // because suggested new leader is null.
```
bshashikant, line 465: I think probbaly, we should ensure there is no retry in the test by setting the attempt count to 1 in the retry policy.. In such case, it will always fail consistently with TimeoutException which is the goal here.

runzhiwang: @bshashikant Thanks for your review. I have updated.

ratis-server/src/test/java/org/apache/ratis/WatchRequestTests.java
```diff
@@ -446,14 +445,18 @@ static void runTestWatchRequestClientTimeout(TestParameters p) throws Exception
...
+    // then sometimes throw NotLeaderException.
+    watchReply = p.sendWatchRequest(1000, RetryPolicies.noRetry());
 
     try {
       watchReply.get();
       fail("runTestWatchRequestClientTimeout failed");
     } catch (Exception ex) {
       LOG.error("error occurred", ex);
-      Assert.assertEquals(RaftRetryFailureException.class,
+      Assert.assertEquals(AlreadyClosedException.class,
```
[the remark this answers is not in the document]
runzhiwang: @bshashikant Thanks for your review. I have updated.

PR #85, opened 2020-05-03
RATIS-639. Inconsistent pattern for variable names

adoroszlai:

1. Allow variable names like `v1` in expressions, too
2. Allow expressions like `0.1 + 0.2`
3. Fix `NumberFormatException` for expression like `-5`

https://issues.apache.org/jira/browse/RATIS-639

## How was this patch tested?

Added test cases in unit test.  Also tested manually via CLI.


adoroszlai:
Thanks @bshashikant for reviewing and committing it.

PR #86, opened 2020-05-04
RATIS-919. Exclude unwanted dependencies from hadoop dependency.

lokeshj1703:
Apache jira: https://issues.apache.org/jira/browse/RATIS-919


ratis-logservice/src/test/java/org/apache/ratis/logservice/impl/TestArchiveHdfsLogReaderAndWriter.java
```diff
@@ -39,11 +39,9 @@
 import org.apache.ratis.logservice.api.ArchiveLogWriter;
 import org.apache.ratis.logservice.api.LogName;
 import org.apache.ratis.logservice.util.LogServiceUtils;
-import org.junit.AfterClass;
-import org.junit.Assert;
-import org.junit.BeforeClass;
-import org.junit.Test;
+import org.junit.*;
```
mukul1987, line 42: The patch looks good to me. Can you please expand this wildcard import ?

PR #88, opened 2020-05-05
RATIS-873. Fix Install Snapshot Notification

hanishakoneru:
This Jira aims to fix the following:

- When Follower is in the process of installing snapshot and it gets an append entry, it replies with result INCONSISTENCY and the follower next index is updated to the snapshot index being installed. This should not happen as the snapshot installation is still in progress. Follower's next index on the leader should remain the same.

- After InstallSnapshot is done, Leader should update the commitIndex of the Follower to the installed snapshot index.

- After InstallSnapshot, when reloading StateMachine, any previously open LogSegment should be closed, if the last entry in the open log is already included in the snapshot.

- When Follower is notified to install snapshot through StateMachine, the reply should indicate the same. It would help with debugging if the Install Snapshot success reply and Install Snapshot notified replies are distinct.

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -376,6 +376,7 @@ public class GrpcLogAppender extends LogAppender {
           final long followerSnapshotIndex = reply.getSnapshotIndex();
           LOG.info("{}: set follower snapshotIndex to {}.", this, followerSnapshotIndex);
           getFollower().setSnapshotIndex(followerSnapshotIndex);
+          updateCommitIndex(followerSnapshotIndex);
           removePending(reply);
           break;
         case NOT_LEADER:
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -222,6 +222,10 @@ public class GrpcLogAppender extends LogAppender {
     }
   }
 
+  private void increaseNextIndex(final long installedSnapshotIndex) {
+    getFollower().increaseNextIndex(installedSnapshotIndex + 1);
+  }
+
   /**
    * StreamObserver for handling responses from the follower
    */
@@ -325,9 +329,15 @@ public class GrpcLogAppender extends LogAppender {
     private final String name = getFollower().getName() + "-" + getClass().getSimpleName();
     private final Queue<Integer> pending;
     private final AtomicBoolean done = new AtomicBoolean(false);
+    private final boolean isNotificationOnly;
 
     InstallSnapshotResponseHandler() {
+      this(false);
+    }
+
+    InstallSnapshotResponseHandler(boolean notifyOnly) {
       pending = new LinkedList<>();
+      this.isNotificationOnly = notifyOnly;
     }
 
     synchronized void addPending(InstallSnapshotRequestProto request) {
@@ -346,7 +356,9 @@ public class GrpcLogAppender extends LogAppender {
 
     void close() {
       done.set(true);
-      GrpcLogAppender.this.notifyAppend();
+      if (!isNotificationOnly) {
+        GrpcLogAppender.this.notifyAppend();
+      }
     }
 
     synchronized boolean hasAllResponse() {
@@ -369,11 +381,7 @@ public class GrpcLogAppender extends LogAppender {
 
       switch (reply.getResult()) {
         case SUCCESS:
-          LOG.info("{}: Completed InstallSnapshot. Reply: {}", this);
-          removePending(reply);
-          break;
-        case NOTIFIED:
-          LOG.info("{}: Notified to InstallSnapshot.", this);
+          LOG.info("{}: Completed InstallSnapshot. Reply: {}", this, reply);
           removePending(reply);
           break;
         case IN_PROGRESS:
@@ -385,6 +393,7 @@ public class GrpcLogAppender extends LogAppender {
           LOG.info("{}: Already Installed Snapshot Index {}.", this, followerSnapshotIndex);
           getFollower().setSnapshotIndex(followerSnapshotIndex);
           updateCommitIndex(followerSnapshotIndex);
+          increaseNextIndex(followerSnapshotIndex);
           removePending(reply);
           break;
         case NOT_LEADER:
@@ -409,12 +418,17 @@ public class GrpcLogAppender extends LogAppender {
       GrpcUtil.warn(LOG, () -> this + ": Failed InstallSnapshot", t);
       grpcServerMetrics.onRequestRetry(); // Update try counter
       resetClient(null);
+      close();
     }
 
     @Override
     public void onCompleted() {
-      if (!firstResponseReceived) {
-        LOG.info("{}: follower responded installSnapshot COMPLETED/NOTIFIED", this);
+      if (!isNotificationOnly) {
+        LOG.info("{}: follower responded installSnapshot COMPLETED", this);
+      } else {
+        if (LOG.isDebugEnabled()) {
+          LOG.info("{}: follower responded installSnapshot COMPLETED", this);
+        }
       }
       close();
     }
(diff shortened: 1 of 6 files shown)


hanishakoneru:
Failing unit tests pass locally.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -369,13 +369,22 @@ public void onNext(InstallSnapshotReplyProto reply) {
...
         case IN_PROGRESS:
+          LOG.info("{}: InstallSnapshot in progress.", this);
           removePending(reply);
           break;
         case ALREADY_INSTALLED:
           final long followerSnapshotIndex = reply.getSnapshotIndex();
-          LOG.info("{}: set follower snapshotIndex to {}.", this, followerSnapshotIndex);
+          LOG.info("{}: Already Installed Snapshot Index {}.", this, followerSnapshotIndex);
           getFollower().setSnapshotIndex(followerSnapshotIndex);
+          updateCommitIndex(followerSnapshotIndex);
```
lokeshj1703, line 387: Should we also update nextIndex?

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -394,17 +403,19 @@ public void onNext(InstallSnapshotReplyProto reply) {
...
+      GrpcUtil.warn(LOG, () -> this + ": Failed InstallSnapshot", t);
+      grpcServerMetrics.onRequestRetry(); // Update try counter
       resetClient(null);
-      close();
     }
 
     @Override
     public void onCompleted() {
-      LOG.info("{}: follower responses installSnapshot COMPLETED", this);
+      if (!firstResponseReceived) {
```
lokeshj1703, line 416: Why do we need this check?

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -394,17 +403,19 @@ public void onNext(InstallSnapshotReplyProto reply) {
...
       if (!isAppenderRunning()) {
-        LOG.info("{} is stopped", this);
+        LOG.info("{} is stopped", GrpcLogAppender.this);
         return;
       }
-      LOG.error("{}: Failed installSnapshot: {}", this, t);
+      GrpcUtil.warn(LOG, () -> this + ": Failed InstallSnapshot", t);
+      grpcServerMetrics.onRequestRetry(); // Update try counter
       resetClient(null);
-      close();
```
lokeshj1703, line 402: We can add a flag in close to mark it as closed without notifying append in this case.

hanishakoneru: Done.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
hanishakoneru: I had added it as otherwise the Log is flooded with "InstallSnapshot COMPLETED" messages even when the install snapshot is in progress. I updated it to not print in case of install snapshot notification.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
hanishakoneru: Done.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -369,13 +369,22 @@ public void onNext(InstallSnapshotReplyProto reply) {
 
       switch (reply.getResult()) {
         case SUCCESS:
+          LOG.info("{}: Completed InstallSnapshot. Reply: {}", this);
```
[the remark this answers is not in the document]
hanishakoneru: Done.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -369,13 +369,22 @@ public void onNext(InstallSnapshotReplyProto reply) {
...
+          removePending(reply);
+          break;
+        case NOTIFIED:
```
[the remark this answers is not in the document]
hanishakoneru: Agree. Done.

hanishakoneru:
Thank you @lokeshj1703 for the review. I addressed all your comments in commit [8ea4014](https://github.com/apache/incubator-ratis/pull/88/commits/8ea401453866d0d9e5a2b099fc6b8ecd6bdee33a).

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -222,6 +222,10 @@ private void increaseNextIndex(AppendEntriesRequestProto request) {
     }
   }
 
+  private void increaseNextIndex(final long installedSnapshotIndex) {
+    getFollower().increaseNextIndex(installedSnapshotIndex + 1);
```
lokeshj1703, line 226: FollowerInfo#increaseNextIndex will throw error if current next index is greater than installedSnapshotIndex + 1. I think we should do something like updateCommitIndex here.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -369,13 +381,19 @@ public void onNext(InstallSnapshotReplyProto reply) {
 
       switch (reply.getResult()) {
         case SUCCESS:
```
lokeshj1703, line 383: Do we need to update commitIndex and nextIndex here as well?

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -394,17 +412,24 @@ public void onNext(InstallSnapshotReplyProto reply) {
...
     @Override
     public void onCompleted() {
-      LOG.info("{}: follower responses installSnapshot COMPLETED", this);
+      if (!isNotificationOnly) {
+        LOG.info("{}: follower responded installSnapshot COMPLETED", this);
+      } else {
+        if (LOG.isDebugEnabled()) {
+          LOG.info("{}: follower responded installSnapshot COMPLETED", this);
+        }
+      }
```
lokeshj1703, line 432: We can combine the if statements using below condition.
`!isNotificationOnly || LOG.isDebugEnabled()`

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLog.java
```diff
@@ -492,6 +492,13 @@ public void syncWithSnapshot(long lastSnapshotIndex) {
     // if the last index in snapshot is larger than the index of the last
     // log entry, we should delete all the log entries and their cache to avoid
     // gaps between log segments.
+
+    // Close open log segment if entries are already included in snapshot
+    LogSegment openSegment = cache.getOpenSegment();
+    if (openSegment != null && openSegment.getEndIndex() <= lastSnapshotIndex) {
+      fileLogWorker.closeLogSegment(openSegment);
+    }
+    cache.clear();
```
lokeshj1703, line 501: We can potentially clear closed log segments with endIndex > snapshot index? Maybe we should move it inside the if statement?

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
hanishakoneru: Yes but I am not sure what all needs to be fixed in the regular InstallSnapshot. I can update the snapshotIndex, commitIndex and nextIndex here for now.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
hanishakoneru: Done.

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLog.java
hanishakoneru: Done.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
hanishakoneru: Done.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
lokeshj1703: Yes..Makes sense to do it in a separate jira.

hanishakoneru:
Thanks @lokeshj1703 for the reviews. Merged into master.

PR #93, opened 2020-05-07
RATIS-931. Fixed Add a raft group to all peers currently registered

juaby:
fixed In the log service, when creating a log, add a raft group to all group peers, where the group peers are not all peers currently registered


runzhiwang:
You can open a  jira at https://issues.apache.org/jira/projects/RATIS/issues, and add the jira number in the Pull Request title. 
Besides, the Pull Request title is also too long.

joshelser (changes requested):
Can we rename the member `peers` to be `allKnownPeers` or similar? Something to try to prevent such an error in the future.

What about a test? Did you try to write a unit test to catch this? I think having 4 available workers, create a log, and then look at the raft groups and see that it has 4 members instead of 3 would work.

juaby:
> Can we rename the member `peers` to be `allKnownPeers` or similar? Something to try to prevent such an error in the future.
> 
> What about a test? Did you try to write a unit test to catch this? I think having 4 available workers, create a log, and then look at the raft groups and see that it has 4 members instead of 3 would work.

Sorry, it took so long to reply to you, about the log service, it's okay to verify by documentation (because the members of available workers == the members of new raft groups), but when the working node starts more than 3, when a log is created, the metadata service broadcasts adding the raft group to an unnecessary out-of-raft-group node.

PR #98, opened 2020-05-11
RATIS-936. Fix ratis-hadoop with non shaded protobuf dependency.

mukul1987:
https://issues.apache.org/jira/browse/RATIS-936

With RATIS-932, dependency on shaded Ratis thirdparty has been removed. This has caused the ratis-hadoop protocol to break.

This jira purposes to fix this by using the protobuf 2.5.0 for the rpc communication and then converting to Protobuf 3.5.0 at rpc endpoint.


lokeshj1703:
@mukul1987 Thanks for the contribution! I have committed the PR to master branch.

PR #99, opened 2020-05-14
RATIS-921. Fix resource leak by closing thousands  of gRPC clients after use

runzhiwang:
**What's the problem ?**
Thousands of gRPC clients were not closed after use, in test and src. It will cause resource leak, and the unit test become very flaky. 

I think the following error is also caused by this problem.
image

**How to fix ?**
This patch do the following:
1. add `GrpcUtil.shutdownManagedChannel` method to ensure the ManagedChannel was closed.
2. change `RaftClient client = ...` to `try (RaftClient client = ...)`, so that client can be closed.
3. change `LogStream logStream1 = client.createLog...` to` try (LogStream logStream1 = client.createLog...)`, so that the client in `LogStreamImpl` can be closed.
4. close the filed client which forgot to be closed in close method. Such as `metaClient` in `LogServer`


runzhiwang:
@lokeshj1703 @bshashikant @mukul1987  Could you help review this patch ? Thank you very much.

bshashikant:
@runzhiwang , can you please update the PR?

runzhiwang:
@bshashikant updated.

bshashikant:

PR #100, opened 2020-05-14
RATIS-937. Loading of open segments always throws corrupted segment warning

hanishakoneru:
When an open log segment is loaded, it always logs a warning that the segment is corrupted.

```
Segment file is corrupted: expected to have -43 entries but only 8 entries read successfully
```
LogSegment#loadSegment() has the following check for corruption:
```
final int expectedEntryCount = Math.toIntExact(end - start + 1);
final boolean corrupted = entryCount != expectedEntryCount;
if (corrupted) {
  LOG.warn("Segment file is corrupted: expected to have {} entries but only {} entries read successfully",
      expectedEntryCount, entryCount);
}
```
But the end is always INVALID_LOG_INDEX (-1) for an open segment.
Before this check, entries are appended to the segment. So the end variable should be updated with the correct endIndex of the segment before checking for corruption.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/LogSegment.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/LogSegment.java
@@ -44,6 +44,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 
+import static org.apache.ratis.server.raftlog.RaftLog.INVALID_LOG_INDEX;
+
 /**
  * In-memory cache for a log segment file. All the updates will be first written
  * into LogSegment then into corresponding files in the same order.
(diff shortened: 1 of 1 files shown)


runzhiwang:
+1

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/LogSegment.java
```diff
@@ -145,6 +147,10 @@ static LogSegment loadSegment(RaftStorage storage, File file, long start, long e
     });
     LOG.info("Successfully read {} entries from segment file {}", entryCount, file);
 
+    if (isOpen && end == INVALID_LOG_INDEX) {
```
mukul1987, line 150: If the segment is open, the end should always be INVALID_LOG_INDEX?
Are there cases where it is not INVALID_LOG_INDEX ?

mukul1987: Also how is the segment's end index initialized after a node restart ?

hanishakoneru: Yes it should always be INVALID_LOG_INDEX. I will remove that check.

Segment endIndex is updated during readSegmentFile call. It in turn calls segment.append for each entry and updates the endIndex.

PR #101, opened 2020-05-15
RATIS-930. Fix failed to remove RaftStorageDirectory

runzhiwang:
**What's the problem ?**
When remove group, `RaftServerImpl` try to remove storage dir on disk. But sometimes leader election thread has not become dead, and still rename the file in storage dir. Then `FileUtils.deleteFully` will throw `NoSuchFileException`, because the file has been renamed by leader election thread. 

You can find from the following stack, `RaftServerImpl` try to delete raft-meta, but leader election thread try to rename raft-meta at the same time, then both fail.

**How to fix ?**
When `FileUtils.deleteFully` throw `NoSuchFileException`, we can retry.


runzhiwang:
@lokeshj1703 @bshashikant @mukul1987 Could you help review this patch ? Thank you very much.

PR #106, opened 2020-05-19
RATIS-942. Fix can not create raftLogMetrics in multi-raft

runzhiwang:
**What's the problem ?**
when create `RaftLogMetrics`, it use `memberId.getPeerId()` to identify, the `peerId` comes from `RaftServerProxy::getId`. So if one Ratis node belongs to different group, the `peerId` is same for different `RaftLogMetrics`. When create the second `RaftLogMetrics` for the second group, it will not create, because it's been created by the first group.

**How to fix ?**
`memberId.toString()` will return `peerId + "@" + groupId`.


runzhiwang:
@bshashikant @lokeshj1703 Could you help review it ? Thank you very much.

runzhiwang:
@dineshchitlangia Thanks for you review.

PR #107, opened 2020-05-23
RATIS-867. Fix failed UT: TestMetaServer#testListLogs

runzhiwang:
**What's the problem ?**

```
ava.lang.AssertionError: expected:<19> but was:<20>
	at org.apache.ratis.logservice.server.TestMetaServer.testJMXCount(TestMetaServer.java:339)
	at org.apache.ratis.logservice.server.TestMetaServer.testListLogs(TestMetaServer.java:331)
```

**What's the reason ?**
1. when create log, it will call [RaftClientImpl::sendRequestWithRetry](https://github.com/apache/incubator-ratis/blob/master/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java#L285), if throw TimeoutIOException, it will retry at [final RaftClientReply reply = sendRequest(request)](https://github.com/apache/incubator-ratis/blob/master/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java#L296), So JMXCount will increase many times at [timerContext = metricRegistry.timer(type.name()).time()](https://github.com/apache/incubator-ratis/blob/master/ratis-logservice/src/main/java/org/apache/ratis/logservice/server/MetaStateMachine.java#L224) when retry happens.  Then JMXCount i.e. 20 not equal to createCount i.e. 19

2. The TimeoutIOException is as follows:
`org.apache.ratis.protocol.TimeoutIOException: deadline exceeded after 2.999977899s. [buffered_nanos=1460409, remote_addr=localhost/127.0.0.1:9001]`

3.  The reason of TimeoutIOException is when create log, it will call [client.groupAdd](https://github.com/apache/incubator-ratis/blob/master/ratis-logservice/src/main/java/org/apache/ratis/logservice/server/MetaStateMachine.java#L325). When add a new group, the following code [file.write(jvmName.getBytes(StandardCharsets.UTF_8))](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageDirectory.java#L342) and [((FileOutputStream)out).getChannel().force(true)](https://github.com/apache/incubator-ratis/blob/master/ratis-common/src/main/java/org/apache/ratis/util/AtomicFileOutputStream.java#L65) will both force sync data to disk. The force sync to disk sometimes cost more than 2 seconds, but the timeout threshold of groupAdd is 3 seconds, so TimeoutIOException happens.

**How to fix ?**
Increase the timeout threshold from 3 seconds to 15 seconds.

### DIFF
--- a/ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
+++ b/ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
@@ -81,17 +81,20 @@ public class TestMetaServer {
             ((MetaStateMachine)master.getMetaStateMachine()).setProperties(properties));
 
         client = new LogServiceClient(cluster.getMetaIdentity(), properties) {
-          @Override public LogStream createLog(LogName logName) throws IOException {
+          @Override
+          public LogStream createLog(LogName logName) throws IOException {
             createCount.incrementAndGet();
             return super.createLog(logName);
           }
 
-          @Override public void deleteLog(LogName logName) throws IOException {
+          @Override
+          public void deleteLog(LogName logName) throws IOException {
             deleteCount.incrementAndGet();
             super.deleteLog(logName);
           }
 
-          @Override public List<LogInfo> listLogs() throws IOException {
+          @Override
+          public List<LogInfo> listLogs() throws IOException {
             listCount.incrementAndGet();
             return super.listLogs();
           }
--- a/ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
+++ b/ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
@@ -18,6 +18,8 @@
 
 package org.apache.ratis.logservice.server;
 
+import org.apache.ratis.client.RaftClientConfigKeys;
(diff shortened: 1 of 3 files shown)


runzhiwang:

ratis-logservice/src/test/java/org/apache/ratis/logservice/server/TestMetaServer.java
```diff
@@ -61,36 +63,52 @@
...
+
+        RaftProperties properties = new RaftProperties();
+        RaftClientConfigKeys.Rpc.setRequestTimeout(properties, TimeDuration.valueOf(15, TimeUnit.SECONDS));
+
+
+        cluster.getMasters().parallelStream().forEach(master ->
+            ((MetaStateMachine)master.getMetaStateMachine()).setProperties(properties));
+
+        client = new LogServiceClient(cluster.getMetaIdentity(), properties) {
+          @Override public LogStream createLog(LogName logName) throws IOException {
```
bshashikant, line 83: Can you plz fix the formatting here?

runzhiwang:
@bshashikant updated

bshashikant:

PR #112, opened 2020-05-25
RATIS-900. Fix Failed UT: RaftExceptionBaseTest.testHandleNotLeaderAndIOException

runzhiwang:
**What's the problem ?**
image

**What's the reason ?**
In the test, the cluster first select an oldLeader such as s1, then force change to newLeader such as s2, and kill s2. Then client connect to s1 and expect s1 reply the info of real leader i.e. s2 in the NotLeaderException. But the s2 has been killed, ratis will triger a leader election, leader will change from s2 to s0, then s0 was included in the NotLeaderException rather than s2. so the test failed.

**How to fix ?**
I think this UT is outdated. We can not expect the leader will not change if we kill the leader.


runzhiwang:

PR #115, opened 2020-05-28
RATIS-950. Handle Exceptions appropriately

dineshchitlangia:

Handled Exception appropriately.
1. InterruptedException should not be ignored, either re-interrupt or re-throw them
2. Catch Exception instead of Throwable

Occurrences in Native API were ignored as Native API will be removed via RATIS-757
Certain occurrences on Throwable were not replaced by Exception as they use the interfaces defined in ratis-common/src/main/java/org/apache/ratis/util/function package and those interfaces extend Throwable.
There needs to be a broader discussion on why they extend throwable and if we really need to change it.

https://issues.apache.org/jira/browse/RATIS-950

## How was this patch tested?

clean build, sonar, checkstyle in local

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/OrderedAsync.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/OrderedAsync.java
@@ -158,6 +158,7 @@ public final class OrderedAsync {
     try {
       requestSemaphore.acquire();
     } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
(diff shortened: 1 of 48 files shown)


runzhiwang:
@dineshchitlangia Could you help resolve the conflicts ? I will review this PR.

dineshchitlangia:
@runzhiwang Sure, I will update this soon.

runzhiwang:
@dineshchitlangia Thanks the patch. I have merged it.

dineshchitlangia:
Thanks @runzhiwang

PR #118, opened 2020-06-03
RATIS-964. Fix failed UT: testRestartLogAppender

runzhiwang:

**What's the problem ?**
Failed UT: testRestartLogAppender

**What's the reason ?**
[grpcServerMetrics.unregister()](https://github.com/apache/incubator-ratis/blob/master/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java#L130) was called when GrpcLogAppender become dead. If the order is: interrupt old GrpcLogAppender ->start new GrpcLogAppender -> grpcServerMetrics.onRequestInconsistency -> old GrpcLogAppender turn dead -> grpcServerMetrics.unregister when old GrpcLogAppender become dead -> unit test check the metric of grpcServerMetrics. Then the unit test failed, because grpcServerMetrics has been unregistered.

**How to fix ?**
grpcServerMetrics.unregister should be called before interrupt old GrpcLogAppender, then start new GrpcLogAppender will create a new grpcServerMetrics.

https://issues.apache.org/jira/browse/RATIS-964

## How was this patch tested?

Existed UT

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -127,7 +127,6 @@ public class GrpcLogAppender extends LogAppender {
     }
 
     Optional.ofNullable(appendLogRequestObserver).ifPresent(StreamObserver::onCompleted);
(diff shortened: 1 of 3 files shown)


runzhiwang:

PR #122, opened 2020-06-12
RATIS-975. Fix failed UT: testRaftLogMetrics

runzhiwang:

1. MetricsStateMachine is instance of DataApi, so `data()` return `this`, and `data().flush` (i.e. `this.flush`) call the method self, the dead recursion happens.
```
  default DataApi data() {
    return this instanceof DataApi? (DataApi)this : DataApi.DEFAULT;
  }
```
2. assertCommitCount always fail. @anshkhannasbu Could you explain why `stmCount + configCount + metaCount == expectedMsgs i.e. 2` ?  Sorry, I can not get the point.

https://issues.apache.org/jira/browse/RATIS-975

## How was this patch tested?

Existed tests.

### DIFF
--- a/ratis-test/src/test/java/org/apache/ratis/server/raftlog/TestRaftLogMetrics.java
+++ b/ratis-test/src/test/java/org/apache/ratis/server/raftlog/TestRaftLogMetrics.java
@@ -86,7 +86,7 @@ public class TestRaftLogMetrics extends BaseTest
     @Override
     public CompletableFuture<Void> flush(long index) {
       flushCount.incrementAndGet();
-      return super.data().flush(index);
+      return CompletableFuture.completedFuture(null);
     }
   }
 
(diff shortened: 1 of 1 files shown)


bshashikant:

bshashikant:
> 
> 1. assertCommitCount always fail. @anshkhannasbu Could you explain why `stmCount + configCount + metaCount == expectedMsgs i.e. 2` ?  Sorry, I can not get the point.
> 

@runzhiwang , u r right. The number client msgs should be equal to the stateMachine entries.
StateMachine + config + metadata should be equal to the total count of raft log entries.

runzhiwang:
@bshashikant Okay, I see, thanks for explanation.

anshkhannasbu:
@runzhiwang @bshashikant thanks for pointing this out. I assumed that in this case, the number of entries sent by client would be the total number of entries committed in the Log.

PR #123, opened 2020-06-12
RATIS-904. Fix Failed UT: testFileStoreAsync

runzhiwang:

1. It's hard to finish write one 10M file and hundred 1M file in 100 seconds, if the leader re-election happens when write, timeout will happens.

2. `if (offset < writeSize)`  we can not throw exception. For example,  if leader finish write data with (offset = 4096, length = 4096), then writeSize of leader become 8192, and 2 follower has not written the data with (offset = 4096, length = 4096), then start a leader re-election, and the old leader elected as a new leader . So client will retry send the data with (offset = 4096, length = 4096), then (offset = 4096) < (writeSize=8192) in leader.  If we throw exception, leader can not write successfully forever.

https://issues.apache.org/jira/browse/RATIS-904

## How was this patch tested?

Existed test.


runzhiwang:

PR #128, opened 2020-06-16
RATIS-978. Fix failed UT: testAddNewFollowers

runzhiwang:

When leader send install snapshot request to follower, leader will [wait](https://github.com/apache/incubator-ratis/blob/master/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java#L480), but when leader receive the response of install snapshot, leader does not notify.

https://issues.apache.org/jira/browse/RATIS-978

## How was this patch tested?

Existed tests

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -413,6 +413,7 @@ public class GrpcLogAppender extends LogAppender {
         default:
           break;
       }
+      notifyAppend();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


runzhiwang:

bshashikant:
Thanks @runzhiwang for working on this. The change looks ok. But , don't we need to check if request for install snapshot is just a notification but not a real snapshot install?

@lokeshj1703 , can you also check?

runzhiwang:
@lokeshj1703 Thanks for review. I have updated.

lokeshj1703:
@runzhiwang Thanks for the contribution! @bshashikant Thanks for review! I have committed the PR to master branch.

PR #129, opened 2020-06-17
RATIS-868. Fix Failed UT: TestRaftWithSimulatedRpc#testWithLoad

runzhiwang:
What's the problem ?
The [threads](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/test/java/org/apache/ratis/server/simulation/RequestHandler.java#L48) in one RequestHandler only process the requests which were sent to one server such as s2.
So request1 from s1 to s2, request2 from s3 to s2 are both processed by one RequestHandler. 
When change oldLeader such as s1, if [q.takeRequest()](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/test/java/org/apache/ratis/server/simulation/SimulatedRequestReply.java#L138) return request1, request1.getRequestorId() will return s1, because blockTakeRequestFrom was set true when [changeLeader](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/test/java/org/apache/ratis/RaftTestUtil.java#L386), so thread was blocked at  [RaftTestUtil.block(reqQ.blockTakeRequestFrom::get)](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/test/java/org/apache/ratis/server/simulation/SimulatedRequestReply.java#L145), if a lot of request from s1 to s2, then all the [threads](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/test/java/org/apache/ratis/server/simulation/RequestHandler.java#L48) in RequestHandler will be blocked, and the request from other server to s2 can not be sent by the blocked thread. So new leader can not be elected.

How to fix ?
When block oldLeader such as s1, we can discard all the request sent by s1 to other servers.

https://issues.apache.org/jira/browse/RATIS-868

## How was this patch tested?

Existed test.

### DIFF
--- a/ratis-server/src/test/java/org/apache/ratis/server/simulation/SimulatedRequestReply.java
+++ b/ratis-server/src/test/java/org/apache/ratis/server/simulation/SimulatedRequestReply.java
@@ -130,19 +130,23 @@ class SimulatedRequestReply<REQUEST extends RaftRpcMessage, REPLY extends RaftRp
       throw new IOException("The RPC of " + qid + " has already shutdown.");
     }
 
-    final REQUEST request;
+    REQUEST request;
     try {
       // delay request for testing
       RaftTestUtil.delay(q.delayTakeRequestTo::get);
+      while (true) {
+        request = q.takeRequest();
+        Preconditions.assertTrue(qid.equals(request.getReplierId()));
 
-      request = q.takeRequest();
-      Preconditions.assertTrue(qid.equals(request.getReplierId()));
-
-      // block request for testing
-      final EventQueue<REQUEST, REPLY> reqQ = queues.get(request.getRequestorId());
-      if (reqQ != null) {
-        RaftTestUtil.delay(reqQ.delayTakeRequestFrom::get);
-        RaftTestUtil.block(reqQ.blockTakeRequestFrom::get);
+        final EventQueue<REQUEST, REPLY> reqQ = queues.get(request.getRequestorId());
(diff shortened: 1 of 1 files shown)


runzhiwang:

ratis-server/src/test/java/org/apache/ratis/server/simulation/SimulatedRequestReply.java
```diff
@@ -130,19 +130,23 @@ public REQUEST takeRequest(String qid) throws IOException {
...
-      if (reqQ != null) {
-        RaftTestUtil.delay(reqQ.delayTakeRequestFrom::get);
-        RaftTestUtil.block(reqQ.blockTakeRequestFrom::get);
+      while (true) {
+        request = q.takeRequest();
+        Preconditions.assertTrue(qid.equals(request.getReplierId()));
+
+        final EventQueue<REQUEST, REPLY> reqQ = queues.get(request.getRequestorId());
+        if (reqQ != null) {
+          // discard request for testing
```
lokeshj1703, line 143: We can update the comment that we are doing a busy wait here.

runzhiwang: @lokeshj1703 Thanks for review. I have updated the PR.

lokeshj1703:
@runzhiwang Thanks for the contribution! I have committed it to master branch.

PR #138, opened 2020-06-29
RATIS-989. Avoid change state from CLOSING to EXCEPTION in LogAppender

runzhiwang:

1. allow EXCEPTION to CLOSING by 
image

2. check LifeCycle.State.isValid(from, to) before state transition.

3. add synchronized to lifeCycle

https://issues.apache.org/jira/browse/RATIS-989

## How was this patch tested?

Existed tests.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/LogAppender.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/LogAppender.java
@@ -73,7 +73,7 @@ public class LogAppender {
         if (!isRunning()) {
           return;
         }
-        transLifeCycle(RUNNING);
+        transitionLifeCycle(RUNNING);
       }
       try {
         runAppenderImpl();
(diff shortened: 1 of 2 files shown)


runzhiwang:
@lokeshj1703 Could you help review this patch ? Thank you very much.

runzhiwang:
CI can not be triggered, We should wait for CI.

ratis-server/src/main/java/org/apache/ratis/server/impl/LogAppender.java
```diff
@@ -117,6 +121,16 @@ void stop() {
     public String toString() {
       return name;
     }
+
+    private boolean transLifeCycle(LifeCycle.State to) {
```
lokeshj1703, line 125: NIT: Lets rename it to transitionLifeCycle.

runzhiwang: @lokeshj1703 Thanks for review. I have updated it.

lokeshj1703:
@runzhiwang Thanks for the contribution! I have committed the PR to master branch.

PR #146, opened 2020-07-10
RATIS-981: Step-down stale leader in case of split-brain

GlenGeng:

If a leader can not hear from majority (including itself)
during last election timeout, it should step down and
become a follower of next term.
 

https://issues.apache.org/jira/projects/RATIS/issues/RATIS-981

## How was this patch tested?

CI


GlenGeng:
The implementation is ready for review. Right now I am working on several unit test, hope I can fix them soon.

PR #150, opened 2020-07-13
RATIS-981: Step-down stale leader in case of split-brain

GlenGeng:

  /**
    * See the thesis section 6.2: A leader in Raft steps down
    * if an election timeout elapses without a successful
    * round of heartbeats to a majority of its cluster.
    */

https://issues.apache.org/jira/projects/RATIS/issues/RATIS-981?filter=allissues

## How was this patch tested?

CI


ratis-server/src/main/java/org/apache/ratis/server/impl/PeerConfiguration.java
```diff
@@ -90,11 +90,8 @@ boolean hasMajority(Collection<RaftPeerId> others, RaftPeerId selfId) {
       if (contains(other)) {
         num++;
       }
-      if (num > size() / 2) {
-        return true;
-      }
     }
-    return false;
+    return num > size() / 2;
```
runzhiwang, line 94: why change this ? looks like equivalent with origin code.

GlenGeng: After this change, it can handle single-server cluster.
Previous one can not handle such case, since its `others` is empty.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderState.java
```diff
@@ -787,6 +789,46 @@ private void replicateNewConf() {
...
+    // immediate round of AppendEntries request. To ensure that leader
+    // is stable, just ignore the check if become leader soon.
+    if (server.getRole().getRoleElapsedTimeMs() < 3 * server.getMaxTimeoutMs()) {
+      return;
+    }
+
+    final List<RaftPeerId> activePeers = senders.stream()
+        .filter(sender -> sender.getFollower()
+                                .getLastRpcResponseTime()
+                                .elapsedTimeMs() <= server.getMaxTimeoutMs())
```
lokeshj1703, line 809: Should we have a separate config for the timeout value? I think it might be better to have a separate config.

ratis-examples/src/test/java/org/apache/ratis/examples/ParameterizedBaseTest.java
```diff
@@ -123,6 +126,11 @@ public boolean hasNext() {
     final RaftProperties prop = new RaftProperties();
     prop.setClass(MiniRaftCluster.STATEMACHINE_CLASS_KEY,
         stateMachineClass, StateMachine.class);
+
+    // make sure leadership check won't affect the test
+    RaftServerConfigKeys.Rpc.setTimeoutMin(prop, TimeDuration.valueOf(2, TimeUnit.SECONDS));
+    RaftServerConfigKeys.Rpc.setTimeoutMax(prop, TimeDuration.valueOf(4, TimeUnit.SECONDS));
```
lokeshj1703, line 132: We should not ideally require setting timeout in the tests. Are we experiencing flakiness in the tests due to the change?

ratis-server/src/main/java/org/apache/ratis/server/impl/PeerConfiguration.java
lokeshj1703: `if (contains(selfId)) {
      num++;
  }`
This check should handle single server case.

GlenGeng: For single-server cluster,  `others` will be empty, `for` loop will not be entered, 
```
       if (num > size() / 2) {		
         return true;		
       }	
```
will be un-reachable.

Even `num = 1` will return `false`

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderState.java
GlenGeng: I am afraid not. Since the criterial is "A leader in Raft steps down if an election timeout elapses without a successful round of heartbeats to a majority of its cluster"

ratis-examples/src/test/java/org/apache/ratis/examples/ParameterizedBaseTest.java
GlenGeng: Yes, some test assume that leader will not step down even it can not hear from followers. 

Such as `runTestUpdateViaHeartbeat` in `TestRaftWithGrpc`, after leader is elected, it just block followers's append requests for 5 seconds, then unblock them

```
    waitForLeader(cluster);
    try (final RaftClient client = cluster.createClient()) {
      // block append requests
      cluster.getServerAliveStream()
          .filter(impl -> !impl.isLeader())
          .map(SimpleStateMachine4Testing::get)
          .forEach(SimpleStateMachine4Testing::blockWriteStateMachineData);

      CompletableFuture<RaftClientReply>
          replyFuture = client.sendAsync(new RaftTestUtil.SimpleMessage("abc"));
      TimeDuration.valueOf(5 , TimeUnit.SECONDS).sleep();
      // replyFuture should not be completed until append request is unblocked.
      Assert.assertTrue(!replyFuture.isDone());
      // unblock append request.
      cluster.getServerAliveStream()
          .filter(impl -> !impl.isLeader())
          .map(SimpleStateMachine4Testing::get)
          .forEach(SimpleStateMachine4Testing::unblockWriteStateMachineData);
```

Such case will failed due to `checkLeadership`

lokeshj1703: blockWriteStateMachineData - it only blocks StateMachine$DataApi#write call. Heartbeats should still work.

GlenGeng: appendEntries in Follower is blocked by StateMachine$DataApi#write, thus won't reply append entries response, which will cause leader to step down.

GlenGeng: Hi @lokeshj1703 
raft use bi-direction streaming grpc, it will maintain sequence between requests and sequence between responses. 

Leader do send heartbeats after sending the append entries request which will be blocked by StateMachine$DataApi#write, but it won't receive replies of these heartbeats until receiving reply of that append entries response.

This sequence assurance is important for maintain log match property of raft.

ratis-test/src/test/java/org/apache/ratis/grpc/TestRetryCacheWithGrpc.java
```diff
@@ -45,6 +47,11 @@
   public void testRetryOnResourceUnavailableException()
       throws InterruptedException, IOException {
     RaftProperties properties = new RaftProperties();
+
+    // make sure leadership check won't affect the test
```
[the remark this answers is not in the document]
GlenGeng: test cases invoking `SimpleStateMachine4Testing::blockWriteStateMachineData()` or `DelayLocalExecutionInjection::setDelayMs()` will be affected, which is meet expectation. We need enlarge election timeout to make sure leader won't step down before hearing append entries response from majority followers.

Another case is related to the membership change, especially the one with 9-servers. To avoid flaky behavior in CI environment, we need enlarge election timeout a little bit, say from 300s to 1s, since leader is vulnerable during such situation.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderState.java
```diff
@@ -787,6 +789,46 @@ private void replicateNewConf() {
...
+      return;
+    }
+
+    List<FollowerInfo> followers = senders.stream()
+        .map(LogAppender::getFollower).collect(Collectors.toList());
+
+    LOG.warn(this + ": Lost leadership on term: " + currentTerm
+        + ". Election timeout: " + server.getMaxTimeoutMs() + "ms"
+        + ". In charge for: " + server.getRole().getRoleElapsedTimeMs() + "ms"
+        + ". Conf: " + conf + ". Followers: " + followers);
```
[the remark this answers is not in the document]
GlenGeng: Yes, it works well. Relies on `List::toString` and `FollowerInfo::toString`.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderState.java
```diff
@@ -787,6 +789,46 @@ private void replicateNewConf() {
...
+   * See the thesis section 6.2: A leader in Raft steps down
+   * if an election timeout elapses without a successful
+   * round of heartbeats to a majority of its cluster.
+   */
+  private void checkLeadership() {
+    // The initial value of lastRpcResponseTime in FollowerInfo is set by
+    // LeaderState::addSenders(), which is fake and used to trigger an
+    // immediate round of AppendEntries request. To ensure that leader
+    // is stable, just ignore the check if become leader soon.
+    if (server.getRole().getRoleElapsedTimeMs() < 3 * server.getMaxTimeoutMs()) {
```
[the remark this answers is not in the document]
GlenGeng: Thanks for pointing this. I will remove this magic number 3. 

Basically, candidates collect votes from majority before becoming leader. Since response time of request votes is recoded, If leader just take charge, without seeing higher term, ideally it should be legal for election timeout.

runzhiwang:
+1

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderState.java
```diff
@@ -792,6 +794,47 @@ private void replicateNewConf() {
...
+    List<FollowerInfo> followers = senders.stream()
+        .map(LogAppender::getFollower).collect(Collectors.toList());
+
+    LOG.warn(this + ": Lost leadership on term: " + currentTerm
+        + ". Election timeout: " + server.getMaxTimeoutMs() + "ms"
+        + ". In charge for: " + server.getRole().getRoleElapsedTimeMs() + "ms"
+        + ". Conf: " + conf + ". Followers: " + followers);
+
+    // become follower of next term.
+    stepDown(currentTerm + 1);
```
lokeshj1703, line 835: I think this should be currentTerm.

ratis-server/src/test/java/org/apache/ratis/RaftAsyncTests.java
```diff
@@ -402,6 +407,46 @@ void runTestAppendEntriesTimeout(CLUSTER cluster) throws Exception {
     RaftServerConfigKeys.RetryCache.setExpiryTime(getProperties(), oldExpiryTime);
   }
 
+  @Test
+  public void testCheckLeadershipFailure() throws Exception {
+    runWithNewCluster(NUM_SERVERS, this::runTestCheckLeadershipFailure);
+  }
+
+  void runTestCheckLeadershipFailure(CLUSTER cluster) throws Exception {
```
lokeshj1703, line 415: This test can also pass when followers turn to candidate state and trigger an election. How about we delay REQUEST_VOTE in the test? And then check the state of current leader before and after timeout?

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderState.java
GlenGeng: ok

ratis-server/src/test/java/org/apache/ratis/RaftAsyncTests.java
GlenGeng: "This test can also pass when followers turn to candidate state and trigger an election."
We don't need to worry about this, the `outstandingOp` will prevent follower to timeout.

lokeshj1703:
@GlenGeng Thanks for the contribution! @runzhiwang Thanks for the review! I have committed the PR to master branch.

PR #151, opened 2020-07-14
RATIS-1001: add currentTerm to LeaderInfoProto for supporting SCM-HA.

GlenGeng:

During SCM-HA, SCM not only needs to know whether it is a leader, but also needs to know which term it is in charge of.

Assume such a case: underlying raft node was leader on term 1, then step down as follower on term 2, then init election and become leader again on term 3. If term is not exposed together with leader information, SCM can not distinguish a leader of term 1 from that of term 3.
 
BTW, according to nanda's design, leader SCM need propagate its term to Datanode, `RaftServerImpl::getRoleInfoProto()` will be a good place to expose term from Ratis to SCM

https://issues.apache.org/jira/browse/RATIS-1001

## How was this patch tested?

CI

### DIFF
--- a/ratis-proto/src/main/proto/Raft.proto
+++ b/ratis-proto/src/main/proto/Raft.proto
@@ -363,6 +363,7 @@ message ServerRpcProto {
 
 message LeaderInfoProto {
   repeated ServerRpcProto followerInfo = 1;
+  uint64 term = 2;
 }
 
 message FollowerInfoProto {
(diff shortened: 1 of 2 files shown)


runzhiwang:
+1

GlenGeng:
@lokeshj1703 

The concern here is we need consider the atomicity of reading role and reading term.

For leader case, `LeaderState` has its member `private final long currentTerm;`, which is read only, meanwhile `RoleInfo` has its member `AtomicReference<LeaderState> leaderState`, which is an atomic variable, thus current `getRoleInfoProto()` can ensure the atomicity of reading role and reading term.

But for candidate and follower, if we want to ensure of same atomicity, we need lock entire `RaftServerImpl`, reading role and call `getState().getCurrentTerm()`. But I didn't see the requirement of return candidate and follower with its term.

lokeshj1703:
@GlenGeng Thanks for the contribution! I have committed the PR to master branch.

PR #154, opened 2020-07-23
RATIS-1008: SimpleStateMachine4Testing#blockWriteStateMachineData should not block heartbeats

GlenGeng:

If raft server use Grpc, when append entries blocked at follower side (e.g., by calling SimpleStateMachine4Testing::blockWriteStateMachineData), leader should still be able to send heartbeat (size of entries of append entries request be 0) to followers, and receive response of heartbeat from follower side.

But test case "testUpdateViaHeartbeat" shows that after calling SimpleStateMachine4Testing::blockWriteStateMachineData, leader do send heartbeats, but ServerRequestStreamObserver::onNext at GrpcServerProtocolService was not called, until follower finish handling the append entries request that blocked by blockWriteStateMachineData.

https://issues.apache.org/jira/projects/RATIS/issues/RATIS-1008

## How was this patch tested?

CI

### DIFF
--- a/ratis-server/src/test/java/org/apache/ratis/statemachine/SimpleStateMachine4Testing.java
+++ b/ratis-server/src/test/java/org/apache/ratis/statemachine/SimpleStateMachine4Testing.java
@@ -61,8 +61,6 @@ import java.util.TreeMap;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Executor;
-import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
@@ -79,7 +77,6 @@ public class SimpleStateMachine4Testing extends BaseStateMachine {
       = "raft.test.simple.state.machine.take.snapshot";
   private static final boolean RAFT_TEST_SIMPLE_STATE_MACHINE_TAKE_SNAPSHOT_DEFAULT = false;
   private boolean notifiedAsLeader;
-  private final Executor executor = Executors.newCachedThreadPool();
 
   public static SimpleStateMachine4Testing get(RaftServerImpl s) {
     return (SimpleStateMachine4Testing)s.getStateMachine();
@@ -148,9 +145,13 @@ public class SimpleStateMachine4Testing extends BaseStateMachine {
       future.complete(null);
     }
 
+    CompletableFuture<Void> getFuture(Type type) {
+      return maps.getOrDefault(type, CompletableFuture.completedFuture(null));
+    }
+
     void await(Type type) {
       try {
-        maps.getOrDefault(type, CompletableFuture.completedFuture(null)).get();
+        getFuture(type).get();
       } catch(InterruptedException | ExecutionException e) {
         throw new IllegalStateException("Failed to await " + type, e);
       }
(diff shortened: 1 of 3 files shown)


GlenGeng:
@lokeshj1703  could you please have a look ?

ratis-server/src/test/java/org/apache/ratis/statemachine/SimpleStateMachine4Testing.java
```diff
@@ -358,20 +361,26 @@ public TransactionContext startTransaction(RaftClientRequest request) {
 
   @Override
   public CompletableFuture<Void> write(LogEntryProto entry) {
-    blocking.await(Blocking.Type.WRITE_STATE_MACHINE_DATA);
-    return CompletableFuture.completedFuture(null);
+    return CompletableFuture.supplyAsync(() -> {
+      blocking.await(Blocking.Type.WRITE_STATE_MACHINE_DATA);
+      return null;
+    }, executor);
```
lokeshj1703, line 367: I think we can use another api in SimpleStateMachine4Testing#Blocking class which just returns a future which is not yet completed. This api would be called here to return a future for  write(LogEntryProto entry) call. unblockWriteStateMachineData can then complete that future. This would avoid a blocking call.

ratis-test/src/test/java/org/apache/ratis/server/raftlog/segmented/TestSegmentedRaftLog.java
```diff
@@ -522,19 +522,20 @@ public void testSegmentedRaftLogStateMachineData() throws Exception {
...
+      Assert.assertTrue(t.isAlive());
+      sm.unblockWriteStateMachineData();
+
       assertIndices(raftLog, flush, next);
       TimeUnit.SECONDS.sleep(1);
       assertIndices(raftLog, flush, next);
       sm.unblockFlushStateMachineData();
       assertIndicesMultipleAttempts(raftLog, flush + 2, next);
+
+      t.join();
```
lokeshj1703, line 538: Why do we need this change?

ratis-server/src/test/java/org/apache/ratis/statemachine/SimpleStateMachine4Testing.java
GlenGeng: Good point! Done.

ratis-test/src/test/java/org/apache/ratis/server/raftlog/segmented/TestSegmentedRaftLog.java
GlenGeng: It is because `raftLog.appendEntry(entry).get();` won't return until `sm.unblockFlushStateMachineData();` called.

GlenGeng:
@lokeshj1703  Thanks for the review, the comments are helpful. 

I've fixed the comments and replied the inline comments, please take a look!
The title of PR and Jira has been changed to  "SimpleStateMachine4Testing#blockWriteStateMachineData should not block heartbeats".

ratis-test/src/test/java/org/apache/ratis/server/raftlog/segmented/TestSegmentedRaftLog.java
lokeshj1703: Got it. I think the old changes were good. Can you please bring it back?

lokeshj1703:
@GlenGeng Thanks for the contribution! I have committed the PR to master branch.

PR #159, opened 2020-07-29
RATIS-805 - Metrics for GRPC messages

anshkhannasbu:

Metrics for GRPC messages. Utilized interceptors for GRPC server.

https://issues.apache.org/jira/browse/RATIS-805

## How was this patch tested?

Verified that metrics were populating by triggering data creation in single node ozone cluster.

Screen Shot 2020-07-29 at 5 40 08 PM


ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
```diff
@@ -111,13 +113,21 @@ private GrpcService(RaftServer raftServer, Supplier<RaftPeerId> idSupplier, int
 
     this.clientProtocolService = new GrpcClientProtocolService(idSupplier, raftServer);
 
+    MetricServerInterceptor monitoringInterceptor = new MetricServerInterceptor(
```
anshkhannasbu, line 116: Better identifier for the Server proxy.

bshashikant: i think we can add the client id and raftserverImpl id...

ratis-grpc/src/main/java/org/apache/ratis/grpc/metrics/MessageMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public void rpcStarted(String rpcType){
+    registry.counter(rpcType + "_started_total").inc();
+  }
+
+  public void rpcCompleted(String rpcType){
+    registry.counter(rpcType + "_completed_total").inc();
+  }
+
+  public void rpcHandled(String rpcType){
```
bshashikant, line 47: why do we need 2 different counters one for rpc handled and one more rpc completed? Can you add some details?

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
anshkhannasbu: that is also created after the server is started.

anshkhannasbu: fixed.

bshashikant:
Thanks @anshkhannasbu . Can you add some unit tests?

anshkhannasbu:
@bshashikant , i added the tests. Could you please check once and let me know your review?

anshkhannasbu:
@bshashikant build passed.

bshashikant:
Thanks @anshkhannasbu for the contribution. I have committed this.

PR #163, opened 2020-08-04
RATIS-1018. Fix Failed UT: testGroupInfo

runzhiwang:

remove following assert
```
if (RaftPeerId.valueOf(i.getServer().getId()).equals(killedFollower)) {
   Assert.assertTrue(i.getCommitIndex() <= maxCommit);
}
```

As the image shows, `killedFollower` is s0, after client receive RaftClientReply, s0 can still increase commit index if it receive metadata entry, then the commit index of s0 maybe greater than the max index in RaftClientReply.

image

image

https://issues.apache.org/jira/browse/RATIS-1018

## How was this patch tested?

No need to add unit tests.


bshashikant:

PR #166, opened 2020-08-05
RATIS-1021. Fix failed UT: testMultipleLinearRandomRetry

runzhiwang:
`randomized.getDuration()` got from the following code, i.e. `rounded`. `multiplier` is a double value less than 1.5 in the test, such as 1.4999999, `duration` is 1000, then `product` will be 1499.9999, so `rounded` will be 1500, which is not less than `d*1.5`, i.e. 1500.

image

https://issues.apache.org/jira/browse/RATIS-1021

## How was this patch tested?

No need to add ut.


runzhiwang:

bshashikant:

amaliujia:
Hello! Can you tell me what is UT for Ratis? Is it integration tests tool?

amaliujia:
I just realized UT means unit tests :-)

PR #167, opened 2020-08-05
RATIS-1022. Clear servers when shutdown MiniRaftCluster

runzhiwang:

clear `servers` when shutdown MiniRaftCluster.
Otherwise, if we need to run test, for example `testMultiRaftGroup`, several times to debug, the second time will fail at `Preconditions.assertTrue(servers.put(id, s) == null);`.

image

image

https://issues.apache.org/jira/browse/RATIS-1022

## How was this patch tested?

No need to add test.


runzhiwang:

PR #168, opened 2020-08-05
RATIS-1023.During installing snapshot, when snapshot file is corrup…

amaliujia:
…ted, rename the temp snapshot file to .corrupt.

It is possible that snapshot file contains corrupted data, in this case within `installSnapshot`, we can rename the temp snapshot file with `.corrupt` before throwing an exception.

https://issues.apache.org/jira/browse/RATIS-1023

## How was this patch tested?
unit test

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
@@ -29,7 +29,6 @@ import org.apache.ratis.proto.RaftProtos.FileChunkProto;
 import org.apache.ratis.proto.RaftProtos.InstallSnapshotRequestProto;
 import org.apache.ratis.statemachine.SnapshotInfo;
 import org.apache.ratis.statemachine.StateMachine;
-import org.apache.ratis.thirdparty.com.google.common.annotations.VisibleForTesting;
 import org.apache.ratis.util.FileUtils;
 import org.apache.ratis.util.IOUtils;
 import org.apache.ratis.util.MD5FileUtil;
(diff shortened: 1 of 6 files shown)


amaliujia:
This is my first time to submit a patch to Ratis. Any suggestion is much appreciated.

amaliujia:
@bshashikant could you take a look?

runzhiwang:
@amaliujia Please change your title to **RATIS-1023. During installing snapshot, when snapshot file is corrup**

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
```diff
@@ -131,4 +133,12 @@ public void installSnapshot(StateMachine stateMachine,
       tmpDir.renameTo(dir.getStateMachineDir());
     }
   }
+
+  // Rename a file by appending .corrupt to file name. This function does not guarantee
+  // that the rename operation is successful.
+  @VisibleForTesting
+  static void renameFileToCorrupt(File tmpSnapshotFile) {
```
runzhiwang, line 140: why static ?

amaliujia: Because this function does not access any class member. Any suggestion or style guide that says better to not use static?

runzhiwang: I think it should be a common util, and should not placed here.

amaliujia: Thanks for your reminder.

I didn't find such function implemented in common util, but indeed I found something useful so I pushed to adopt those. I also moved the rename-to-corrupt to FileUtils.

bshashikant:
Thanks Rui Wang for starting working on this. Can you please add some details as to why renaming temp snapshot file to .corrupt is required?

amaliujia:
Thanks @bshashikant! I updated the JIRA and we can discuss there.

bshashikant:
Thanks @amaliujia for the contribution and @runzhiwang for the review.

PR #174, opened 2020-08-09
RATIS-1029. NPE at MetricServerInterceptor.interceptCall

amaliujia:
When identifier is not NULL, `metrics` variable will not be initialized: https://github.com/apache/incubator-ratis/blob/349c365e81d76b210df7a97704217178e3f7f826/ratis-grpc/src/main/java/org/apache/ratis/grpc/metrics/intercept/server/MetricServerInterceptor.java#L71

Screen Shot 2020-08-08 at 10 51 24 PM

https://issues.apache.org/jira/browse/RATIS-1029

## How was this patch tested?

UT

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/metrics/intercept/server/MetricServerInterceptor.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/metrics/intercept/server/MetricServerInterceptor.java
@@ -69,10 +69,8 @@ public class MetricServerInterceptor implements ServerInterceptor {
         identifier = defaultIdentifier;
       }
     }
-    else{
-         if(metrics == null){
-            metrics = new MessageMetrics(identifier, "server");
-         }
+    if (metrics == null) {
+      metrics = new MessageMetrics(identifier, "server");
     }
     String metricNamePrefix = getMethodMetricPrefix(method);
     ServerCall<R,S> monitoringCall = new MetricServerCall<>(call, metricNamePrefix, metrics);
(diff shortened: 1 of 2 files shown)


amaliujia:
R: @anshkhannasbu @bshashikant

amaliujia:
This PR is related to #159.

anshkhannasbu:
@amaliujia, if the identifier is not null then metrics should be set. Since there is no other way to set the identifier. Is there some case where metrics cannot be created?

ratis-grpc/src/main/java/org/apache/ratis/grpc/metrics/intercept/server/MetricServerInterceptor.java
```diff
@@ -62,14 +62,14 @@ private String getMethodMetricPrefix(MethodDescriptor<?, ?> method){
...
+    if (identifier == null) {
       try {
         identifier = peerIdSupplier.get().toString();
-      } catch (Exception e){
+      } catch (Exception e) {
         identifier = defaultIdentifier;
       }
-      metrics = new MessageMetrics(identifier, "server");
     }
+    metrics = new MessageMetrics(identifier, "server");
```
anshkhannasbu, line 72: This will probably create a new metrics info for every invocation. I think we should move it to an else in case identifier is not null and check if metrics are set or not.
```suggestion
    else{
         if(metrics == null){
            metrics = new MessageMetrics(identifier, "server");
         }
    }
```

amaliujia:
@anshkhannasbu 

I accepted your suggestion and then give another look.  I think it is better to have a separate `if (metrics == null)` check (see my last commit). WDYT?

anshkhannasbu:
> @anshkhannasbu
> 
> I accepted your suggestion and then give another look. I think it is better to have a separate `if (metrics == null)` check (see my last commit). WDYT?

@amaliujia sure looks good.

PR #181, opened 2020-08-25
RATIS-1032. Change proto and RaftConfiguration to support priority

runzhiwang:

Change proto and RaftConfiguration to support priority

https://issues.apache.org/jira/browse/RATIS-1032

## How was this patch tested?

add ut to test it.


runzhiwang:

PR #187, opened 2020-08-31
RATIS-1047. Fix can not elect leader when higher priority server crash

runzhiwang:

For example, s1, s2, s3's priority is (1, 2, 3). If s1 askForVotes, s2 reply success, but s3 crash, so exception happens.
Then election rejected at [return logAndReturn(Result.REJECTED, responses, exceptions, -1)](https://github.com/runzhiwang/incubator-ratis/blob/052e109bf33bedaf6754c6a26f55c2eb1d9dbae9/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderElection.java#L351)

https://issues.apache.org/jira/browse/RATIS-1047

## How was this patch tested?

add test.


runzhiwang:

PR #190, opened 2020-09-08
RATIS-1051. Use try-resource-block when saveMd5File

maobaolong:

Use try-resource-block when saveMd5File

RATIS-1051

## How was this patch tested?

No need.


maobaolong:
@runzhiwang Please take a look at this PR. no try-resource-block cannot ensure the `afos` will close when exception happen.

runzhiwang:
+1
@maobaolong Thanks for the patch, I have merged it.

PR #192, opened 2020-09-08
RATIS-1053. Math operands should be cast before assignment

maobaolong:

Math operands should be cast before assignment

RATIS-1052

## How was this patch tested?

It is already a test.


runzhiwang:
+1
@maobaolong Thanks the patch, I have merged it.

amaliujia:
This makes sense!

PR #193, opened 2020-09-08
RATIS-1055. Switch cases should end with an unconditional "break" statement

maobaolong:

Switch cases should end with an unconditional "break" statement

image

RATIS-1055

## How was this patch tested?

No need.

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -409,6 +409,8 @@ public class GrpcLogAppender extends LogAppender {
               getServer().getId(), installSnapshotEnabled, getFollowerId(), !installSnapshotEnabled);
           break;
         case UNRECOGNIZED:
+          LOG.error("Unrecongnized the reply result {}: Leader is {}, follower is {}",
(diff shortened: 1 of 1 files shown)


ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -407,7 +407,9 @@ public void onNext(InstallSnapshotReplyProto reply) {
           LOG.error("{}: Configuration Mismatch ({}): Leader {} has it set to {} but follower {} has it set to {}",
               this, RaftServerConfigKeys.Log.Appender.INSTALL_SNAPSHOT_ENABLED_KEY,
               getServer().getId(), installSnapshotEnabled, getFollowerId(), !installSnapshotEnabled);
+          break;
         case UNRECOGNIZED:
+          break;
```
amaliujia, line 412: Should there be a LOG printed when hitting `UNRECOGNIZED`?

maobaolong: @amaliujia Yeah, thanks you the suggestion, done.

amaliujia:
LGTM

runzhiwang:
+1
@maobaolong Thanks the patch, I have merged it.

PR #194, opened 2020-09-09
RATIS-1058. Avoid NPE when response is null in CombinedClientProtocolServerSideTranslatorPB.sendClient

amaliujia:

Avoid NPE when response is null in CombinedClientProtocolServerSideTranslatorPB.sendClient.

Detailed analysis see https://issues.apache.org/jira/browse/RATIS-1058.

https://issues.apache.org/jira/browse/RATIS-1058

## How was this patch tested?

N/A

### DIFF
--- a/ratis-hadoop/src/main/java/org/apache/ratis/hadooprpc/client/CombinedClientProtocolServerSideTranslatorPB.java
+++ b/ratis-hadoop/src/main/java/org/apache/ratis/hadooprpc/client/CombinedClientProtocolServerSideTranslatorPB.java
@@ -79,7 +79,7 @@ public class CombinedClientProtocolServerSideTranslatorPB
     }
     return ClientReplyProto.newBuilder()
         .setType(type)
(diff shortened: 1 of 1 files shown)


amaliujia:
@runzhiwang  can you take a look? Thanks!

amaliujia:
@runzhiwang 

Can you take a look at this PR please?

runzhiwang:
@amaliujia Thanks the patch. But how response == null happens ?

amaliujia:
> @amaliujia Thanks the patch. But how response == null happens ?
This is more like a pre-defence check. The NULL can happen when a new value added to the ENUM but the code in the SWITCH is not updated. I think it is always good to eliminate potential NPE.  

Alternately, we can choose to throw an exception for a new type, WDYT?

runzhiwang:
@amaliujia I think we fix NPE only when it could happen sometimes, otherwise we need check too much in the code.

amaliujia:
That's ok. This is a very minor change anyway.

PR #195, opened 2020-09-09
RATIS-1059. Avoid NPE when exportInfo is NULL in LogStateMachine.processArchiveLog

amaliujia:

Avoid NPE when exportInfo is NULL in LogStateMachine.processArchiveLog. 

Detailed analysis see https://issues.apache.org/jira/browse/RATIS-1059

https://issues.apache.org/jira/browse/RATIS-1059

## How was this patch tested?

N/A

### DIFF
--- a/ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
+++ b/ratis-logservice/src/main/java/org/apache/ratis/logservice/server/LogStateMachine.java
@@ -623,12 +623,15 @@ public class LogStateMachine extends BaseStateMachine {
         loc = archiveLog.getLocation();
         ArchivalInfo exportInfo =
             exportMap.putIfAbsent(loc, new ArchivalInfo(loc));
-        if (exportInfo != null && exportInfo.getLastArchivedIndex() == archiveLog
-            .getLastArchivedRaftIndex()) {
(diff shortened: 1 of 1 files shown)


amaliujia:
@joshelser  Can you take a look? Thanks!

runzhiwang:
+1

runzhiwang:
@amaliujia Thanks for the patch, I have merged it.

amaliujia:
@runzhiwang 

that sounds good. I do pay attention to instable tests. If I can find something easy to fix, I will cut PRs.

runzhiwang:
@amaliujia  Thanks. You can pay attenation to RATIS-916 and RATIS-1017, they all caused by java.net.BindException: Address already in use. This is the top one reason cause ut flaky. Maybe the root cause is some connection not closed when finish ut.

amaliujia:
@runzhiwang 
The current problem is I cannot reproduce such problems locally (even keep running tests locally). Do you know whether we can download the full log from CI?

runzhiwang:
@amaliujia 

> The current problem is I cannot reproduce such problems locally (even keep running tests locally)

It's not easy to reproduce, you can run CI in your github ratis repo many times by commit a lot of empty commit.  I think when you run CI 20 times, it will reproduce.

> Do you know whether we can download the full log from CI ?

image
image

PR #197, opened 2020-09-14
RATIS-1063. Failed UT: testOldLeaderCommit: expected leader is different than actual leader

amaliujia:
## What changes were proposed in this pull request?`
Screen Shot 2020-09-13 at 10 09 13 PM

The test failed because the actual elected leader is not the expected leader.

Analysis:
The original test keeps only one follower and kill other follower, then write a new message to Ratis ring. So the only left follower will have one entry that other closed follower do not have.

Then the test kills the leader. The intention here is to expect that follower with higher entry index be selected as leader. However, because only one follower has higher entry index, so Ratis does not guarantee that follower is elected.

The resolution is to keep at least half of followers alive to preserve that higher entry index. 

https://issues.apache.org/jira/browse/RATIS-1063

## How was this patch tested?

UT

### DIFF
--- a/ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
+++ b/ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
@@ -19,6 +19,7 @@ package org.apache.ratis;
 
 import com.codahale.metrics.Gauge;
 
+import java.util.Set;
 import org.apache.log4j.Level;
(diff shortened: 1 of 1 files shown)


amaliujia:
R: @bshashikant @runzhiwang 

Can you take a look please?

runzhiwang:
@amaliujia Thanks the patch. I have merged it.

PR #198, opened 2020-09-14
RATIS-1060. Fix TestRaftReconfiguration.testRestorePriority

amaliujia:
Screen Shot 2020-09-13 at 11 33 19 PM

Analysis:
testRestorePriority failed due to no leader selected exception. Based on the log, two of three server in the cluster was stuck at STARTING state so they could not participate into leader election, no matter how long waiting time.

So this PR proposes to restart the cluster when no leader is elected when have reached half of the number of attempts. Because at this moment it is likely that the servers are stuck somehow so keep waiting won't change anything.  

https://issues.apache.org/jira/browse/RATIS-1060

## How was this patch tested?

UT

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/util/JavaUtils.java
+++ b/ratis-common/src/main/java/org/apache/ratis/util/JavaUtils.java
@@ -17,6 +17,7 @@
  */
 package org.apache.ratis.util;
 
+import org.apache.ratis.util.function.CheckedFunction;
 import org.apache.ratis.util.function.CheckedRunnable;
 import org.apache.ratis.util.function.CheckedSupplier;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


amaliujia:
R @bshashikant @runzhiwang  Can you take a look please? Thanks!

runzhiwang:
>  Based on the log, two of three server in the cluster was stuck at STARTING state so they could not participate into leader election

@amaliujia Thanks the patch. Could you find the root cause of cluster was stuck at STARTING state ?

amaliujia:
@runzhiwang 
Unfortunately the log was truncated and also I only caught this situation once locally.

If I can see the similar situation again with full logs I might be able to dig into the root cause.

runzhiwang:
@amaliujia Okay, I would prefer to the root cause fix.

PR #200, opened 2020-09-15
RATIS-1066. Fixed example filestore client run failed

coolbeevip:

The docs[1] that describe client command will added duplicate peers parameters.
 
```
${BIN}/client.sh filestore loadgen --size 1024000 --numFiles 10 --peers ${PEERS}
Found xxx/incubator-ratis/ratis-examples/target/ratis-examples-1.1.0-SNAPSHOT.jar
Wrong parameters: Can only specify option --peers once.
 ```

I think the QUORUM_OPTS parameter should not be used in client.sh, just like server.sh.
 
[1] https://github.com/apache/incubator-ratis/blob/master/ratis-examples/README.md#filestore-client

https://issues.apache.org/jira/browse/RATIS-1066

## How was this patch tested?

manual tests

### DIFF
--- a/ratis-examples/src/main/bin/client.sh
+++ b/ratis-examples/src/main/bin/client.sh
@@ -32,4 +32,4 @@ shift
 subcommand="$1"
 shift
 
-java ${LOGGER_OPTS} -jar $ARTIFACT "$example" "$subcommand" $QUORUM_OPTS "$@"
+java ${LOGGER_OPTS} -jar $ARTIFACT "$example" "$subcommand" "$@"

### CODE REVIEW

**Comment by amaliujia** [comment]
LGTM

File: ratis-examples/src/main/bin/client.sh [hunk 1]
```diff
@@ -32,4 +32,4 @@ shift
 subcommand="$1"
 shift
 
-java ${LOGGER_OPTS} -jar $ARTIFACT "$example" "$subcommand" $QUORUM_OPTS "$@"
```
amaliujia, line 35: Alternatively we can keep `QUORUM_OPTS` and then have a default setup for client.sh/server.sh so people can start client/server without adding the peers.

But that will be overlapped with the purpose of `start-all.sh` so I think remove `QUORUM_OPTS` is ok.

coolbeevip: Thank you, I also think it’s better to let users set their own peers parameters

runzhiwang:
@coolbeevip Thanks the patch, sorry for the late merge. @amaliujia Thanks for review. I have merged the patch.

PR #201, opened 2020-09-15
RATIS-1064. Fix TestGrpcOutputStream.testSimpleWrite

amaliujia:
`TestGrpcOutputStream` has failed both `testKillLeader` and `testSimpleWrite`. This PR fixes `testSimpleWrite`.

For the failure of `testSimpleWrite`:
Screen Shot 2020-09-15 at 3 07 40 PM

It is because the CLUSTER hasn't elect a leader.

https://issues.apache.org/jira/browse/RATIS-1064

## How was this patch tested?

UT

### DIFF
--- a/ratis-server/src/test/java/org/apache/ratis/OutputStreamBaseTest.java
+++ b/ratis-server/src/test/java/org/apache/ratis/OutputStreamBaseTest.java
@@ -26,6 +26,7 @@ import org.apache.ratis.server.raftlog.RaftLog;
 import org.apache.ratis.util.SizeInBytes;
 import org.apache.ratis.util.StringUtils;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.IOException;
@@ -75,6 +76,7 @@ public abstract class OutputStreamBaseTest<CLUSTER extends MiniRaftCluster>
   private void runTestSimpleWrite(CLUSTER cluster) throws Exception {
     final int numRequests = 5000;
     final int bufferSize = 4;
+    RaftTestUtil.waitForLeader(cluster);
     try (OutputStream out = newOutputStream(cluster, bufferSize)) {
       for (int i = 0; i < numRequests; i++) { // generate requests
         out.write(toBytes(i));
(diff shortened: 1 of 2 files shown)


ratis-server/src/test/java/org/apache/ratis/OutputStreamBaseTest.java
```diff
@@ -259,6 +261,7 @@ private void runTestWriteWithOffset(CLUSTER cluster) throws Exception {
   /**
    * Write while leader is killed
    */
+  @Ignore
```
amaliujia, line 264: Will tackle this one separately: https://issues.apache.org/jira/browse/RATIS-1065

amaliujia:
R: @runzhiwang Can you take a look please?

ratis-server/src/test/java/org/apache/ratis/OutputStreamBaseTest.java
runzhiwang: We do not need add Ignore here, you can still track it.

ratis-test/src/test/java/org/apache/ratis/grpc/TestGrpcOutputStream.java
```diff
@@ -31,9 +32,7 @@
 
 /**
  * Test {@link GrpcOutputStream}
- * TODO: {@link GrpcOutputStream} current has some bugs.
  */
-@Ignore
```
runzhiwang, line 36: Please run it many times to make sure there is no bug.

amaliujia: @runzhiwang I wasn't clear enough before:

both `testKillLeader` and `testSimpleWrite` are not a case of transient failures. They failed deterministic. 

This PR fixed the root cause of `testSimpleWrite` and RATIS-1065 will track the fix for `testKillLeader`

I will fix both RATIS-1065 and RATIS-1064 to move them out of the category of transient failures to make it more clear.

amaliujia: I also ran `TestGrpcOutputStream.testSimpleWrite` 30 times locally and without seeing a problem.

ratis-server/src/test/java/org/apache/ratis/OutputStreamBaseTest.java
amaliujia: This is a case that fails deterministically so I added a `@Ignore`. Will try to go to root case and document that in RATIS-1065.

amaliujia:
Also @bshashikant to see if you have a chance to check this PR?

runzhiwang:
@amaliujia Thanks the patch, I have merged it.

PR #203, opened 2020-09-16
RATIS-455. Move exception classes to org.apache.ratis.protocol.exceptions

amaliujia:

Move exception classes to  `org.apache.ratis.protocol.exceptions` package

https://issues.apache.org/jira/browse/RATIS-455

## How was this patch tested?

UT


amaliujia:
R: @szetszwo 

Can you take a look?

szetszwo (approved):
+1 the change looks good.

PR #208, opened 2020-09-25
RATIS-1075. Classes that implement AutoCloseable should call "close()" when it should be terminated

amaliujia:

https://sonarcloud.io/project/issues?id=apache_incubator-ratis&issues=AXJWdiRzwsm5D8rduwS_&open=AXJWdiRzwsm5D8rduwS_

https://issues.apache.org/jira/browse/RATIS-1075

## How was this patch tested?

UT

### DIFF
--- a/ratis-logservice/src/main/java/org/apache/ratis/logservice/tool/VerificationTool.java
+++ b/ratis-logservice/src/main/java/org/apache/ratis/logservice/tool/VerificationTool.java
@@ -124,60 +124,61 @@ public class VerificationTool {
           return;
         }
         System.out.println(tool.metaQuorum);
-        LogServiceClient client = new LogServiceClient(tool.metaQuorum);
-        ExecutorService executor = Executors.newCachedThreadPool();
-        List<Future<?>> futures = new ArrayList<Future<?>>(tool.numLogs);
+        try (LogServiceClient client = new LogServiceClient(tool.metaQuorum)) {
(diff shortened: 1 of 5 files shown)


ratis-logservice/src/main/java/org/apache/ratis/logservice/tool/VerificationTool.java
```diff
@@ -176,6 +176,7 @@ public static void main(String[] args) throws IOException {
           }
           waitForCompletion(futures);
         }
+        client.close();
```
amaliujia, line 179: Because client implements `AutoCloseable`, so it has a `close()` function should be called.

runzhiwang: we can use `try (LogServiceClient client = new LogServiceClient(tool.metaQuorum)) { ... }`

amaliujia: @runzhiwang Yes. Pushed a commit to use try-with-resources pattern.

amaliujia:
R: @runzhiwang 

I totally fix 5 cases that need be closed finally in this PR.

runzhiwang:
I will merge it once CI passed

amaliujia:
Thanks @runzhiwang . Tests passed.

runzhiwang:
@amaliujia Thanks the patch,  I have merged it.

PR #209, opened 2020-09-26
RATIS-1077. Unify metric name convention

amaliujia:

There are two convention of metric naming convention 
`public static final String RAFT_LOG_WORKER_QUEUE_SIZE = "workerQueueSize";` (for example, in [RaftLogMetrics](https://github.com/apache/incubator-ratis/blob/master/ratis-server/src/main/java/org/apache/ratis/server/metrics/RaftLogMetrics.java#L44))
`public static final String RATIS_GRPC_METRICS_REQUEST_RETRY_COUNT = "num_retries";`

We can have a unified convention.

https://issues.apache.org/jira/browse/RATIS-1077

## How was this patch tested?

UT


amaliujia:
R: @bshashikant @runzhiwang

runzhiwang:
I think there is other name need to unify.

runzhiwang:
Besides, we should do more important work.

PR #223, opened 2020-10-13
Revert "Revert "RATIS-672. Vagrant test's Maven 404's""

Parigyan:
This reverts commit 89abd139

./dev-support/vagrant/run_all_tests.sh build command was failing.
There was previous patch available for the legacy issue , concerning Maven 404 response from apache mirror for maven installation.
The fix has to do with reverting the patch to original fix, and merging the new changes with the concerned files.

https://issues.apache.org/jira/browse/RATIS-672

## How was this patch tested?
$./dev-support/vagrant/run_all_tests.sh build , command ran successfully
Screenshot 2020-10-13 at 11 57 46 PM

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLog.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLog.java
@@ -326,11 +326,13 @@ public final class SegmentedRaftLog extends RaftLogBase {
 
   private void checkAndEvictCache() {
     if (cache.shouldEvict()) {
-      // TODO if the cache is hitting the maximum size and we cannot evict any
-      // segment's cache, should block the new entry appending or new segment
-      // allocation.
-      cache.evictCache(server.getFollowerNextIndices(), fileLogWorker.getSafeCacheEvictIndex(),
-          server.getLastAppliedIndex());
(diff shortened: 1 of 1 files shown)


lokeshj1703 (approved):
@Parigyan Thanks for working on this! 
./dev-support/vagrant/run_all_tests.sh build completed successfully for me.

lokeshj1703 (commented):
@Parigyan I was going through the commit messages for the reason why RATIS-672 was reverted.
It appears that docker build is failing with the PR. Can you please take a look?
./dev-support/docker Try building the Dockerfile in this folder for reproducing the issue.

lokeshj1703:
The new Dockerfile would always require some arguments I guess. Can we give some default values to the ARG defined in the Dockerfile?

Parigyan:
@lokeshj1703 Thanks for the feedback, I'll take a look into dockerfile build.

Parigyan:
Found https://yetus.apache.org/documentation/0.12.0/RELEASENOTES/ 
YETUS-919 was specifically created to address the issue of build failure in YETUS.

add_docker_build_arg should get the work done.
in Dockerfile the argument for protobuf is already there, and it picks the value from source /dev-support/ci/common.sh 
but it seems in case of yetus it fails as the common.sh is not sufficient in getting executed.

I'll try to use the new mechanism of passing the args for docker build in yetus.

PR #229, opened 2020-10-21
RATIS-1101. Fix Failed UT: RaftBasicTests.runTestOldLeaderNotCommit:IndexOutOfBounds

amaliujia:

Seeing a flaky instance of `RaftBasicTests.runTestOldLeaderNotCommit` due to `IndexOutOfBounds` (https://github.com/apache/incubator-ratis/pull/225/checks?check_run_id=1281324912).

The exception is from 
```
      for (int i = 1; i < NUM_SERVERS - 1; i++) {
        RaftServerImpl follower = followers.get(i);
        cluster.killServer(follower.getId());
      }
```
where it turned out i exceeded the range of followers size. This might be due to slow machine/process so followers are not enough.

Per suggestion from @szetszwo  we can use `org.junit.AssumptionViolatedException` to not fail this test in this special case.

https://issues.apache.org/jira/browse/RATIS-1101

## How was this patch tested?
UT


amaliujia:
R: @szetszwo can you take a look?

szetszwo (approved):
+1 the change looks good.

PR #241, opened 2020-10-30
RATIS-1120. Terminate election in advance if majority peers reject

dengziming:

In the leader election approach, the candidate will change to leader in advance if the majority voters vote for it. we can also terminate the approach if majority voters reject it.

https://issues.apache.org/jira/browse/RATIS-1120

## How was this patch tested?

unit test.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/PeerConfiguration.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/PeerConfiguration.java
@@ -94,6 +94,16 @@ class PeerConfiguration {
     return num > size() / 2;
   }
 
+  boolean majorityRejectVotes(Collection<RaftPeerId> rejected) {
+    int num = size();
+    for (RaftPeerId other : rejected) {
+      if (contains(other)) {
+        num --;
+      }
+    }
+    return num <= size() / 2;
+  }
+
   @Override
   public boolean equals(Object obj) {
     if (this == obj) {
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftConfiguration.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftConfiguration.java
@@ -206,6 +206,12 @@ public final class RaftConfiguration {
         (oldConf == null || oldConf.hasMajority(others, selfId));
   }
 
+  /** @return true if the rejects are in the majority(maybe half is enough in some cases). */
+  boolean majorityRejectVotes(Collection<RaftPeerId> rejects) {
+    return conf.majorityRejectVotes(rejects) ||
+            (oldConf != null && oldConf.majorityRejectVotes(rejects));
+  }
+
   @Override
   public String toString() {
     return logEntryIndex + ": " + conf + ", old=" + oldConf;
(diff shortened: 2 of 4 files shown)


runzhiwang:
@dengziming Thanks for the first PR.  Please fill What changes were proposed in this pull request?,  What is the link to the Apache JIRA?, How was this patch tested?.

dengziming:
@runzhiwang ,Hi, I changed the comment, please take a look.

ratis-server/src/main/java/org/apache/ratis/server/impl/PeerConfiguration.java
```diff
@@ -94,6 +94,16 @@ boolean hasMajority(Collection<RaftPeerId> others, RaftPeerId selfId) {
     return num > size() / 2;
   }
 
+  boolean majorityImpossibleWithout(Collection<RaftPeerId> rejected) {
```
runzhiwang, line 97: Do you think rename majorityImpossibleWithout to majorityRejectVote maybe better ?

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftConfiguration.java
```diff
@@ -206,6 +206,12 @@ boolean hasMajority(Collection<RaftPeerId> others, RaftPeerId selfId) {
         (oldConf == null || oldConf.hasMajority(others, selfId));
   }
 
+  /** @return true if the rejects are in the half(impossible for others in majority). */
+  boolean majorityImpossibleWithout(Collection<RaftPeerId> rejects) {
+    return conf.majorityImpossibleWithout(rejects) &&
+            (oldConf == null || oldConf.majorityImpossibleWithout(rejects));
+  }
```
runzhiwang, line 213: I think this check it not totally reverse  of hasMajority.  The reverse of hasMajority is `!conf.hasMajority(others, selfId) || !(oldConf == null || oldConf.hasMajority(others, selfId))` i.e.` !conf.hasMajority(others, selfId) || (oldConf != null && !oldConf.hasMajority(others, selfId))``. i.e. ` conf.majorityImpossibleWithout(rejects) || (oldConf != null && !oldConf.majorityImpossibleWithout(rejects))``

ratis-server/src/main/java/org/apache/ratis/server/impl/PeerConfiguration.java
dengziming: OK, firstly I assume rejected votes is not necessarily a majority, maybe half is enough. But I think majorityRejectVote is more self-explanatory even it is not very accurate.

runzhiwang: If you have better name, I also can accept it.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftConfiguration.java
dengziming: This is right! I just forgot this.

runzhiwang:
reopen pr to trigger ci.

runzhiwang:
@dengziming Thanks the first patch. I have merged it. Looking forward to your more contribution.

PR #253, opened 2020-11-03
RATIS-1128. Update Configuration on InstallSnapshot

hanishakoneru:

If a Follower is lagging behind and needs to install snapshot to catch up, it can miss any configuration changes that were applied on the Leader. When installing snapshot, the configuration change must also be propagated and applied on the follower. 

https://issues.apache.org/jira/browse/RATIS-1128

## How was this patch tested?

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerProtoUtils.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerProtoUtils.java
@@ -38,6 +38,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static org.apache.ratis.server.impl.RaftServerConstants.DEFAULT_CALLID;
+import static org.apache.ratis.server.impl.RaftServerConstants.DEFAULT_TERM;
 
 /** Server proto utilities for internal use. */
 public interface ServerProtoUtils {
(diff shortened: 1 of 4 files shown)


ratis-server/src/main/java/org/apache/ratis/server/impl/ServerProtoUtils.java
```diff
@@ -384,22 +385,29 @@ static InstallSnapshotRequestProto toInstallSnapshotRequestProto(
             .addAllFileChunks(chunks)
             .setTotalSize(totalSize)
             .setDone(done);
+    // Set term to DEFAULT_TERM as this term is not going to used by installSnapshot to update the RaftConfiguration
+    final LogEntryProto confLogEntryProto = toLogEntryProto(raftConfiguration, DEFAULT_TERM,
+        raftConfiguration.getLogEntryIndex());
```
hanishakoneru, line 390: The term of this LogEntryProto is not used to update the in memory RaftConfiguration (or ConfigurationManager). But it is updated in the Raft Meta conf file (raft-meta.conf).
@szetszwo , @lokeshj1703, would it be better to send the correct term corresponding to the last conf entry while sending the installSnapshot?

szetszwo (approved):
+1 the change looks good.

PR #266, opened 2020-11-10
RATIS-1144. Default disable datastream when start a MiniRaftCluster

runzhiwang:

Default disable datastream when start a MiniRaftCluster, so that the other test no need to start data stream server, to make the CI run faster.

https://issues.apache.org/jira/browse/RATIS-1144

## How was this patch tested?

no need to add new ut.

### DIFF
--- a/ratis-test/src/test/java/org/apache/ratis/datastream/TestDataStreamNetty.java
+++ b/ratis-test/src/test/java/org/apache/ratis/datastream/TestDataStreamNetty.java
@@ -96,6 +96,7 @@ public class TestDataStreamNetty extends DataStreamBaseTest {
       RaftPeerId peerId = RaftPeerId.valueOf("s" + i);
       RaftProperties properties = new RaftProperties();
       NettyConfigKeys.DataStream.setPort(properties, NetUtils.createLocalServerAddress().getPort());
+      RaftConfigKeys.DataStream.setType(properties, SupportedDataStreamType.NETTY);
 
       if (i == leaderIndex) {
         raftClientReply = expectedClientReply;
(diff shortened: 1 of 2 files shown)


szetszwo (approved):
+1

amaliujia:
+1

ratis-test/src/test/java/org/apache/ratis/datastream/TestDataStreamNetty.java
```diff
@@ -96,6 +96,7 @@ private void testCloseStream(int leaderIndex, int numServers) throws Exception {
       RaftPeerId peerId = RaftPeerId.valueOf("s" + i);
       RaftProperties properties = new RaftProperties();
       NettyConfigKeys.DataStream.setPort(properties, NetUtils.createLocalServerAddress().getPort());
+      RaftConfigKeys.DataStream.setType(properties, SupportedDataStreamType.NETTY);
```
amaliujia, line 99: This line is redundant. If you check line 61 this config is already be set.

amaliujia: hmmm I see `RaftProperties properties = new RaftProperties();` is not using the existing `properties`. 

I will take a look and maybe fix this in my branch.

runzhiwang: @amaliujia It's a mock RaftServer, will not use properties set at  line 61.

amaliujia: O I see. thanks for clarification

PR #271, opened 2020-11-11
RATIS-1146. Does not build with newer JDK

adoroszlai:

1. Add `javax.annotation-api` dependency, required for compilation with Java 9+.
2. Improve _compile_ CI check to build Ratis with both Java 8 and Java 11.

https://issues.apache.org/jira/browse/RATIS-1146

## How was this patch tested?

Compile checks:
https://github.com/adoroszlai/incubator-ratis/runs/1385567749#step:4:10
https://github.com/adoroszlai/incubator-ratis/runs/1385567774#step:4:1729

Compiled Ratis locally with the same Java versions, and Java 14, too:

```
$ mvn -DskipTests -V clean package
...
Java version: 14.0.1
...
[INFO] BUILD SUCCESS
```

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -19,9 +19,19 @@ jobs:
   build:
     name: compile
     runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        java: [ 8, 11 ]
+      fail-fast: false
     steps:
-      - uses: actions/checkout@master
-      - run: ./dev-support/checks/build.sh
+      - name: Checkout project
+        uses: actions/checkout@v2
+      - name: Setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Run a full build
+        run: ./dev-support/checks/build.sh
   rat:
     name: rat
     runs-on: ubuntu-18.04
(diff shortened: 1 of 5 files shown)


.github/workflows/post-commit.yml
```diff
@@ -19,9 +19,19 @@ jobs:
   build:
     name: compile
     runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        java: [ 8, 11 ]
```
amaliujia, line 24: Out of curiosity: is this the setup to have post commits runs build by java 11? If so, is there a way to trigger the post commit in this PR so this change can be verified?

runzhiwang (approved):
LGTM

.github/workflows/post-commit.yml
adoroszlai: > is this the setup to have post commits runs build by java 11?

Correct.

> If so, is there a way to trigger the post commit in this PR so this change can be verified?

It was [already triggered](https://github.com/adoroszlai/incubator-ratis/runs/1385567774) in my own fork.

amaliujia: Thank you for the clarification!

runzhiwang:
@adoroszlai Thanks the patch. @amaliujia Thanks for review. I have merged it.

adoroszlai:
Thanks @amaliujia for the review, and @runzhiwang for reviewing and merging this.

PR #272, opened 2020-11-11
RATIS-1130. Add E2E test for Ratis streaming based on MiniRaftCluster

amaliujia:

Add E2E test for Ratis streaming based on MiniRaftCluster

https://issues.apache.org/jira/browse/RATIS-1130

## How was this patch tested?

UT


amaliujia:
R: @szetszwo @runzhiwang

ratis-client/src/main/java/org/apache/ratis/client/api/DataStreamApi.java
```diff
@@ -36,5 +38,5 @@
  */
 public interface DataStreamApi {
   /** Create a stream to write data. */
-  DataStreamOutput stream();
+  DataStreamOutputRpc stream();
```
szetszwo, line 41: DataStreamApi is user API.  We want to hide the RPC API from users.  So, please don't make this change.

We may cast it to DataStreamOutputRpc in the test.

ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyServerStreamRpc.java
```diff
@@ -326,8 +327,11 @@ private StreamInfo newStreamInfo(ByteBuf buf) {
   }
 
   static long writeTo(ByteBuf buf, DataStream stream) {
-    final WritableByteChannel channel = stream.getWritableByteChannel();
     long byteWritten = 0;
+    if (stream == null) {
```
szetszwo, line 331: Why the stream could be null?

ratis-server/src/main/java/org/apache/ratis/server/DisabledDataStreamServerFactory.java
```diff
@@ -33,6 +34,11 @@ public DataStreamServerRpc newDataStreamServerRpc(RaftServer server) {
       @Override
       public void start() {}
 
+      @Override
+      public InetSocketAddress getInetSocketAddress() {
+        return InetSocketAddress.createUnresolved("0.0.0.0", 0);
```
szetszwo, line 39: Is it okay to return null?

ratis-server/src/test/java/org/apache/ratis/DataStreamTestUtils.java
```diff
@@ -0,0 +1,41 @@
...
+        final ByteBuffer buffer = ByteBuffer.allocateDirect(size);
+        final int length = buffer.capacity();
+        buffer.position(0).limit(length);
+        for (int j = 0; j < length; j++) {
+            buffer.put(pos2byte(offset + j));
+        }
+        buffer.flip();
+        Assert.assertEquals(length, buffer.remaining());
+        return buffer;
+    }
```
szetszwo, line 40: 2-space indentations please.

ratis-server/src/test/java/org/apache/ratis/DataStreamTests.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.Test;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ThreadLocalRandom;
+
+public abstract class DataStreamTests <CLUSTER extends MiniRaftCluster> extends BaseTest
```
szetszwo, line 43: 2-space indentation please.

szetszwo (changes requested):
Thanks for working on this.  The change looks good.  Just some simple comments.

ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyServerStreamRpc.java
amaliujia: In current test I am using BaseStateMachine, in which the stream api returns:

https://github.com/apache/incubator-ratis/blob/f91ef63809b21ed65445f4a60d8005ce0a1af8ca/ratis-server/src/main/java/org/apache/ratis/statemachine/StateMachine.java#L91

ratis-server/src/main/java/org/apache/ratis/server/DisabledDataStreamServerFactory.java
amaliujia: hmm good point. Let me try.

amaliujia: `null` works. Have updated to using `null`

ratis-server/src/test/java/org/apache/ratis/DataStreamTestUtils.java
amaliujia: Done.

ratis-server/src/test/java/org/apache/ratis/DataStreamTests.java
amaliujia: Done.

I tried to update my IDE config again to match the 2-space indentation. I have three devices so always mess up with which IDE is using which style.

ratis-client/src/main/java/org/apache/ratis/client/api/DataStreamApi.java
amaliujia: Got it. Changed this back.

amaliujia:
@szetszwo comment addressed.

ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyServerStreamRpc.java
szetszwo: We should override DataApi.stream(RaftClientRequest request) in the test.  Otherwise, how could it test the correctness?  We may use MultiDataStreamStateMachine in DataStreamBaseTest.

amaliujia: Yeah that what I was thinking in https://issues.apache.org/jira/browse/RATIS-1130?focusedCommentId=17228841&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17228841

Sure I can move `MultiDataStreamStateMachine` to make it reusable. Do you prefer I do that in another PR (as this PR is pretty large now) or just add that refactoring in this PR?

Also adding a `NULL` is still useful. I found that without this NULL check, this code will hit a NPE and the program will hang there. If don't add this NULL check, then at least we should check the NPE and solve why program will hang.

ratis-netty/src/test/java/org/apache/ratis/netty/MiniRaftClusterWithNetty.java
```diff
@@ -47,6 +48,20 @@ public MiniRaftClusterWithNetty newCluster(String[] ids, RaftProperties prop) {
     }
   }
 
+  public interface FactoryGetWithDataStreamEnabled extends Factory.Get<MiniRaftClusterWithNetty> {
```
runzhiwang, line 51: try the following code, and delete `FactoryGetWithDataStreamEnabled`.
```
public class TestDataStreamWithNetty extends DataStreamTests<MiniRaftClusterWithNetty>
    implements MiniRaftClusterWithNetty.FactoryGet {
  {
    RaftConfigKeys.DataStream.setType(getProperties(), SupportedDataStreamType.NETTY);
  }
}
```

ratis-server/src/test/java/org/apache/ratis/DataStreamTests.java
```diff
@@ -0,0 +1,114 @@
...
+
+      for (int i = 0; i < futures.size(); i++) {
+        futures.get(i).join();
+      }
+
+      // send close
+      dataStreamOutputRpc.closeAsync().join();
+
+      // send start transaction
+      dataStreamOutputRpc.startTransactionAsync().join();
```
runzhiwang, line 93: should remove this, closeAsync includes startTransactionAsync.

ratis-server/src/test/java/org/apache/ratis/DataStreamTests.java
```diff
@@ -0,0 +1,114 @@
...
+      for(int i = 0; i < bufferNum; i++) {
+        final int size = halfBufferSize + ThreadLocalRandom.current().nextInt(halfBufferSize);
+        sizes.add(size);
+
+        final ByteBuffer bf = initBuffer(dataSize, size);
+        futures.add(dataStreamOutputRpc.writeAsync(bf));
+        dataSize += size;
+      }
+
+      for (int i = 0; i < futures.size(); i++) {
```
runzhiwang, line 85: should remove these join.

ratis-server/src/test/java/org/apache/ratis/DataStreamTests.java
```diff
@@ -0,0 +1,114 @@
...
+      final List<Integer> sizes = new ArrayList<>();
+
+      dataStreamOutputRpc.getHeaderFuture().join();
+
+      // send data
+      final int halfBufferSize = bufferSize / 2;
+      int dataSize = 0;
+      for(int i = 0; i < bufferNum; i++) {
+        final int size = halfBufferSize + ThreadLocalRandom.current().nextInt(halfBufferSize);
+        sizes.add(size);
```
runzhiwang, line 78: why need sizes ?

ratis-server/src/test/java/org/apache/ratis/DataStreamTests.java
```diff
@@ -0,0 +1,114 @@
...
+    Assert.assertEquals(NUM_SERVERS, peers.size());
+    RaftPeer raftPeer = peers.iterator().next();
+
+    try (RaftClient client = cluster.createClient(raftPeer)) {
+      // send header
+      DataStreamOutputRpc dataStreamOutputRpc = (DataStreamOutputRpc) client.getDataStreamApi().stream();
+      final List<CompletableFuture<DataStreamReply>> futures = new ArrayList<>();
+      final List<Integer> sizes = new ArrayList<>();
+
+      dataStreamOutputRpc.getHeaderFuture().join();
```
runzhiwang, line 71: Do not need this.

ratis-server/src/test/java/org/apache/ratis/DataStreamTests.java
```diff
@@ -0,0 +1,114 @@
...
+  // TODO: change bufferSize and bufferNum configurable
+  private static int bufferSize = 1_000_000;
+  private static int bufferNum =  10;
+
+  @Test
+  public void testStreamWrites() throws Exception {
+    runWithNewCluster(NUM_SERVERS, this::testStreamWrites);
+  }
+
+  void testStreamWrites(CLUSTER cluster) throws Exception {
```
runzhiwang, line 56: Should test multi-client and multi-stream.

ratis-netty/src/test/java/org/apache/ratis/netty/MiniRaftClusterWithNetty.java
```diff
@@ -59,6 +74,7 @@ protected RaftServerProxy newRaftServer(
       RaftPeerId id, StateMachine.Registry stateMachineRegistry , RaftGroup group,
       RaftProperties properties) throws IOException {
     NettyConfigKeys.Server.setPort(properties, getPort(id, group));
+    NettyConfigKeys.DataStream.setPort(properties, getDataStreamPort(id, group));
```
runzhiwang, line 77: What's the relationship between this propertis and the propertis in the following code, why need setPort twice?
image

runzhiwang:
@amaliujia I think this PR related to the failed CI.

ratis-server/src/test/java/org/apache/ratis/DataStreamTests.java
amaliujia: Agreed. But I will prefer to add that support in another PR.

I am trying to make this PR as one that have a working MiniRaftCluster with a simple test to prove it. Then we can add more tests afterwards.

ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyServerStreamRpc.java
amaliujia: Another thought is, with MiniRaftCluster based test, do we still need existing test cases? Can we migrate all tests to MiniRaftCluster based?

ratis-netty/src/test/java/org/apache/ratis/netty/MiniRaftClusterWithNetty.java
amaliujia: In `TestDataStreamNetty`, MiniRaftCluster is not used. This line is setup the port for MiniRaftCluster. 

I am asking above for whether we want to merge current tests to only use MiniRaftCluster.

amaliujia:
@runzhiwang 
Tests can pass locally so the failed CL might not be related to this PR. If there is still issues after rerun I will definitely take a deeper look.

ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyServerStreamRpc.java
runzhiwang: We need existing test case. We need both unit test and integration Test.

runzhiwang: I  do not think we need to add a NULL check. if NULL happen in test, we should change the test to avoid this. Please think about it.

runzhiwang:
> @runzhiwang
> Tests can pass locally so the failed CL might not be related to this PR. If there is still issues after rerun I will definitely take a deeper look.

@amaliujia Now we merge PR  when it pass CI.

amaliujia:
@runzhiwang  yes agreed. What I meant was this PR will be updated and CI will rerun. If there are problems at that time I will take a look. Right now as this PR is not finalized I will focus on code changes first.

runzhiwang:
> @runzhiwang yes I agree. What I meant was this PR will be updated and CI will rerun. If there are problems at that time I will take a look. Right now as this PR is not finalized I will focus on code changes first.

Sure, thanks

ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyServerStreamRpc.java
amaliujia: I will hold off if we need the `NULL` check for now because it is not related to test itself. 


Back to whether we need `MultiDataStreamStateMachine` for this MiniRaftCluster test, here is my question:

If we enable `MultiDataStreamStateMachine` for MiniRaftCluster test, what is the difference between this MiniRaftCluster test and the existing `TestDataStreamNetty`? From the test coverage MiniRaftCluster will be a super set over `TestDataStreamNetty`, why we want to keep both?

amaliujia: Done

amaliujia:
@szetszwo @runzhiwang 

Comments addressed. Can you take a look?

ratis-test/src/test/java/org/apache/ratis/datastream/DataStreamTests.java
```diff
@@ -0,0 +1,104 @@
...
+  private static int bufferNum =  10;
+
+  @Test
+  public void testStreamWrites() throws Exception {
+    runWithNewCluster(NUM_SERVERS, this::testStreamWrites);
+  }
+
+  void testStreamWrites(CLUSTER cluster) throws Exception {
+    final RaftServerImpl leader = waitForLeader(cluster);
+    Assert.assertTrue(leader != null);
```
szetszwo, line 59: This can be removed since leader is never null.

ratis-server/src/test/java/org/apache/ratis/MiniRaftCluster.java
```diff
@@ -269,6 +283,19 @@ public MiniRaftCluster initServers() {
...
+    LOG.info("Setting up data stream servers");
+    for (RaftServerProxy serverProxy : servers.values()) {
+      serverProxy.getDataStreamServerRpc().addRaftPeers(getOtherRaftPeers(serverProxy.getId()));
+    }
+  }
+
+  private Collection<RaftPeer> getOtherRaftPeers(RaftPeerId id) {
+    HashMap peerMap = new HashMap<>(peers);
+    peerMap.remove(id);
+    return peerMap.values();
```
szetszwo, line 296: Replace it with the following.
```suggestion
    return peers.values().stream().filter(p -> !p.getId().equals(id)).collect(Collectors.toList());
```

ratis-test/src/test/java/org/apache/ratis/datastream/DataStreamTests.java
```diff
@@ -0,0 +1,104 @@
...
+
+    final RaftGroup raftGroup = cluster.getGroup();
+    final Collection<RaftPeer> peers = raftGroup.getPeers();
+    Assert.assertEquals(NUM_SERVERS, peers.size());
+    RaftPeer raftPeer = peers.iterator().next();
+
+    try (RaftClient client = cluster.createClient(raftPeer)) {
+      // send header
+      DataStreamOutputRpc dataStreamOutputRpc = (DataStreamOutputRpc) client.getDataStreamApi().stream();
+      final List<CompletableFuture<DataStreamReply>> futures = new ArrayList<>();
```
szetszwo, line 69: futures is not used.  Please remove it.

ratis-test/src/test/java/org/apache/ratis/datastream/TestDataStreamWithNettyMiniRaftCluster.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.apache.ratis.RaftConfigKeys;
+import org.apache.ratis.datastream.DataStreamBaseTest.MultiDataStreamStateMachine;
+import org.apache.ratis.netty.MiniRaftClusterWithNetty;
+
+public class TestDataStreamWithNettyMiniRaftCluster extends DataStreamTests<MiniRaftClusterWithNetty>
+    implements MiniRaftClusterWithNetty.FactoryGet {
+  {
+    RaftConfigKeys.DataStream.setType(getProperties(), SupportedDataStreamType.NETTY);
+    setStateMachine(MultiDataStreamStateMachine.class);
```
szetszwo, line 28: Move this to DataStreamTests since all subclasses need it.

szetszwo (commented):

runzhiwang (approved):
LGTM, waiting CI passed.

amaliujia:
@runzhiwang CI has passed.

runzhiwang:
@szetszwo Could you help review it again ?

szetszwo (approved):
+1 the change looks good.

PR #297, opened 2020-11-24
RATIS-1177. NPE when RaftServerJmxAdapter queries getLeaderId.

hanishakoneru:

When a follower is in candidate state but unable to get votes and also unable to get new leader info, the JMXBeanServer query to getLeaderId can throw a NPE. 

This happens because RaftServerJmxAdapter#getLeaderId() does not check whether leaderId is null before calling toString() on it.


szetszwo (approved):
+1

PR #315, opened 2020-12-02
RATIS-1194. SegmentedRaftLog syncWithSnapshot should not create a new open segment.

avijayanhwx:
In SegmentedRaftLog#syncWithSnapshot, if the snapshot index is the last open segment index, then the open segment is finalized and closed (which releases the SegmentedRaftLogOutputStream). In the next step, while rolling the open segment in the cache, a new open segment is created. However, the output stream has already been closed. This causes the peer to terminate when a transaction comes in after the syncWithSnapshot call.

Instead of creating a new open segment here, we should let any new transaction coming in to the system create a new open segment through the StartOpenSegment call which re-initializes the output stream.

https://issues.apache.org/jira/browse/RATIS-1194

## How was this patch tested?
Manually tested.


avijayanhwx:
cc @lokeshj1703 / @hanishakoneru for review.

lokeshj1703 (approved):
@avijayanhwx Thanks for working on this! The changes look good to me. +1.

lokeshj1703:
@avijayanhwx Thanks for the contribution! I have committed the PR to master branch.

PR #329, opened 2020-12-07
RATIS-1212. Remove steams from StreamMap when close stream

runzhiwang:

Remove steams from StreamMap when close stream

https://issues.apache.org/jira/browse/RATIS-1212

## How was this patch tested?

No need to add ut.


szetszwo (approved):
+1 thanks for working on this.

amaliujia:
+1

PR #335, opened 2020-12-08
RATIS-1217. Fix filestore failed readStateMachineData when use AsyncApi

runzhiwang:

1. There are two usage for FileStore#read, the first is for FileStoreStateMachine#query which read  data has been committed, the second if for FileStoreStateMachine#read which read data has been written by leader, and leader read the data to appendEntry to follower. So in the second case, we should not check `offset + length > committedSize`, we should check `offset + length > writeSize`

image

2. data length should use ByteBuffer#remaining()
```
     final WriteRequestHeaderProto.Builder header = WriteRequestHeaderProto.newBuilder()
         .setPath(ProtoUtils.toByteString(path))
         .setOffset(offset)
-        .setLength(data.position())
+        .setLength(data.remaining())
         .setClose(close);
```

https://issues.apache.org/jira/browse/RATIS-1217

## How was this patch tested?

manually execute following command many times.
${BIN}/client.sh filestore loadgen --size 100000000 --numFiles 10 --bufferSize 4000000 --peers ${PEERS}

### DIFF
--- a/ratis-examples/src/main/java/org/apache/ratis/examples/filestore/FileInfo.java
+++ b/ratis-examples/src/main/java/org/apache/ratis/examples/filestore/FileInfo.java
@@ -62,7 +62,12 @@ abstract class FileInfo {
     return relativePath;
   }
 
-  long getSize() {
+  long getWriteSize() {
+    throw new UnsupportedOperationException(
+        "File " + getRelativePath() + " size is unknown.");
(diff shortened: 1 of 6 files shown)


runzhiwang:

szetszwo (approved):
+1 the change looks good.

PR #353, opened 2020-12-14
RATIS-1238. Seperate request and write executor

runzhiwang:

Seperate the thread pool of request and write disk 

https://issues.apache.org/jira/browse/RATIS-1238

## How was this patch tested?

No need to add new ut.


runzhiwang:
@szetszwo In my test, if seperate the thread pool, different clients's cost are almost same, but if use only one thread pool, different clients' cost vary a lot.

szetszwo (approved):
+1 the change looks good.

PR #364, opened 2020-12-19
RATIS-1248. Support network topology

runzhiwang:

Not only support primary -> peer1 && peer2, but also support primary -> peer1 -> peer2.

https://issues.apache.org/jira/browse/RATIS-1248

## How was this patch tested?

TODO

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/api/DataStreamApi.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/api/DataStreamApi.java
@@ -17,7 +17,11 @@
  */
 package org.apache.ratis.client.api;
 
+import org.apache.ratis.protocol.RaftPeerId;
+
 import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Stream data asynchronously to all the servers in the {@link org.apache.ratis.protocol.RaftGroup}.
@@ -44,4 +48,7 @@ public interface DataStreamApi {
 
   /** Create a stream by providing a customized header message. */
   DataStreamOutput stream(ByteBuffer headerMessage);
+
+  /** Create a stream by providing a customized header message and route table. */
+  DataStreamOutput stream(ByteBuffer headerMessage, Map<RaftPeerId, List<RaftPeerId>> routingTable);
 }
--- a/ratis-client/src/main/java/org/apache/ratis/client/api/DataStreamApi.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/api/DataStreamApi.java
@@ -17,11 +17,9 @@
  */
 package org.apache.ratis.client.api;
 
-import org.apache.ratis.protocol.RaftPeerId;
+import org.apache.ratis.protocol.RoutingTable;
 
 import java.nio.ByteBuffer;
-import java.util.List;
-import java.util.Map;
 
 /**
  * Stream data asynchronously to all the servers in the {@link org.apache.ratis.protocol.RaftGroup}.
@@ -50,5 +48,5 @@ public interface DataStreamApi {
   DataStreamOutput stream(ByteBuffer headerMessage);
 
   /** Create a stream by providing a customized header message and route table. */
-  DataStreamOutput stream(ByteBuffer headerMessage, Map<RaftPeerId, List<RaftPeerId>> routingTable);
+  DataStreamOutput stream(ByteBuffer headerMessage, RoutingTable routingTable);
 }
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/ClientProtoUtils.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/ClientProtoUtils.java
@@ -115,16 +115,25 @@ public interface ClientProtoUtils {
     }
   }
 
+  static RoutingTable getRoutingTable(RaftClientRequestProto p) {
+    if (!p.hasRoutingTable()) {
+      return null;
+    }
+
+    RoutingTable.Builder builder = RoutingTable.newBuilder();
+    for (RouteProto routeProto : p.getRoutingTable().getRoutesList()) {
+      RaftPeerId from = RaftPeerId.valueOf(routeProto.getPeerId().getId());
+      List<RaftPeerId> to = routeProto.getSuccessorsList().stream()
+          .map(v -> RaftPeerId.valueOf(v.getId())).collect(Collectors.toList());
+      builder.addSuccessors(from, to);
+    }
+
+    return builder.build();
+  }
+
   static RaftClientRequest toRaftClientRequest(RaftClientRequestProto p) {
     final RaftClientRequest.Type type = toRaftClientRequestType(p);
     final RaftRpcRequestProto request = p.getRpcRequest();
-    RoutingTable.Builder routingTablebuilder = new RoutingTable.Builder();
-    for (RouteProto routeProto : p.getRoutingTable().getRoutingTableList()) {
-      RaftPeerId from = RaftPeerId.valueOf(routeProto.getFrom().getId());
-      List<RaftPeerId> to = routeProto.getToList().stream()
-          .map(v -> RaftPeerId.valueOf(v.getId())).collect(Collectors.toList());
-      routingTablebuilder.addSuccessors(from, to);
-    }
 
     return new RaftClientRequest(
         ClientId.valueOf(request.getRequestorId()),
@@ -134,7 +143,7 @@ public interface ClientProtoUtils {
         toMessage(p.getMessage()),
         type,
         request.getSlidingWindowEntry(),
-        routingTablebuilder.build());
+        getRoutingTable(p));
   }
 
   static ByteBuffer toRaftClientRequestProtoByteBuffer(RaftClientRequest request) {
@@ -148,7 +157,9 @@ public interface ClientProtoUtils {
       b.setMessage(toClientMessageEntryProtoBuilder(request.getMessage()));
     }
 
-    b.setRoutingTable(request.getRoutingTable().toProto());
+    if (request.getRoutingTable() != null) {
+      b.setRoutingTable(request.getRoutingTable().toProto());
+    }
 
     final RaftClientRequest.Type type = request.getType();
     switch (type.getTypeCase()) {
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/ClientProtoUtils.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/ClientProtoUtils.java
@@ -18,6 +18,8 @@
 package org.apache.ratis.client.impl;
 
 import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 
 import org.apache.ratis.datastream.impl.DataStreamReplyByteBuffer;
@@ -93,6 +95,12 @@ public interface ClientProtoUtils {
         request.getSlidingWindowEntry());
   }
 
+  static RoutingTableProto.Builder toRoutingTableProtoBuilder(
+      RaftClientRequest request) {
+    return RoutingTableProto.newBuilder()
+        .addAllRoutingTable(ProtoUtils.toRouteProtos(request.getRoutingTable()));
+  }
+
   static RaftClientRequest.Type toRaftClientRequestType(RaftClientRequestProto p) {
     switch (p.getTypeCase()) {
       case WRITE:
@@ -118,6 +126,14 @@ public interface ClientProtoUtils {
   static RaftClientRequest toRaftClientRequest(RaftClientRequestProto p) {
     final RaftClientRequest.Type type = toRaftClientRequestType(p);
     final RaftRpcRequestProto request = p.getRpcRequest();
+    Map<RaftPeerId, List<RaftPeerId>> routingTable = new HashMap<>();
+    for (RouteProto routeProto : p.getRoutingTable().getRoutingTableList()) {
+      RaftPeerId from = RaftPeerId.valueOf(routeProto.getFrom().getId());
+      List<RaftPeerId> to = routeProto.getToList().stream()
+          .map(v -> RaftPeerId.valueOf(v.getId())).collect(Collectors.toList());
+      routingTable.put(from, to);
+    }
+
     return new RaftClientRequest(
         ClientId.valueOf(request.getRequestorId()),
         RaftPeerId.valueOf(request.getReplyId()),
@@ -125,7 +141,8 @@ public interface ClientProtoUtils {
         request.getCallId(),
         toMessage(p.getMessage()),
         type,
-        request.getSlidingWindowEntry());
+        request.getSlidingWindowEntry(),
+        routingTable);
   }
 
   static ByteBuffer toRaftClientRequestProtoByteBuffer(RaftClientRequest request) {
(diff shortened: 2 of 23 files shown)


runzhiwang:
@szetszwo Hi, sorry, now I have one problem, where should I put the route table in proto ? Now I put route table in DataStreamInitProto, and encode DataStreamInitProto as  RaftClientRequestProto#message. But this can not work for filestore, because filestore encode FileStoreRequestProto as RaftClientRequestProto#message, I can only put route table in StreamWriteRequestProto. But this seems a little complicated, maybe we should put route table in a same proto, no matter whether it's filestore. Do you have any suggestion ?

ratis-proto/src/main/proto/Raft.proto
```diff
@@ -280,6 +280,19 @@ message WatchRequestTypeProto {
...
 }
 
+message RoutePairProto {
+  RaftPeerProto from = 1;
+  RaftPeerProto to = 2;
+}
+
+message DataStreamRouteTableProto {
+  repeated RoutePairProto routeTable = 1;
+}
```
szetszwo, line 290: Let's call the RouteProto and RouteTableProto.  It may be useful outside DataStream in the future.

Let's use RaftPeerId instead of RaftPeerProto.  It may be a good time to create RaftPeerIdProto which is currently missing.

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -231,23 +232,48 @@ StreamInfo remove(Key key) {
     return f;
   }
 
+
+  private IdentityHashMap<RaftPeer, RaftPeer> getRouteTable(RaftClientRequest request)
```
szetszwo, line 236: Let's create a org.apache.ratis.protocol.RouteTable interface/class in ratis-common.
```
 interface RoutingTable {
    List<RaftPeerId> getSuccessors(RaftPeerId peer);
  }
```
I think we need RaftPeerId since the RaftPeer can be looked up locally.

BTW, why using IdentityHashMap but not just a Map in the implementation?

szetszwo (commented):
We probably should add the RoutingTableProto to RaftClientRequestProto.  Or we may add RoutingTableProto to DataStreamRequestTypeProto.  What do you think?

szetszwo:
BTW, let's separate the FileStore change to another JIRA since RoutingTable is an important change.  We should focus on it.

Thanks a lot for working on this.

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
runzhiwang: IdentityHashMap allows duplicated key, such as primary -> peer1, primary -> peer2, we can put them in IdentityHashMap, but can not put them in a map which do not allow duplicated key.

runzhiwang:
> We probably should add the RoutingTableProto to RaftClientRequestProto

@szetszwo I agree.

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
szetszwo: IdentityHashMap may have unexpected behaviour since it uses == but not equals(..).  Let's use 
```
Map<RaftPeerId, List<RaftPeerId>>
```
Or, we may use guava Multimap https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Multimap.html

runzhiwang: Got it, I will use Map<RaftPeerId, List<RaftPeerId>>. Thanks a lot.

runzhiwang:
@szetszwo I have updated the patch, could you help review it again ?

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -145,6 +150,22 @@ LocalStream getLocal() {
...
       return JavaUtils.getClassSimpleName(getClass()) + ":" + request;
     }
+
+    @Override
+    public List<RaftPeerId> getSuccessors(RaftPeerId peerId) throws IOException {
+      Map<RaftPeerId, List<RaftPeerId>> routingTable = request.getRoutingTable();
+      if (routingTable.containsKey(peerId)) {
+        return routingTable.get(peerId);
+      } else if (isPrimary()) {
+        // TODO(runzhiwang): if filestore support route table, we can remove this branch
```
szetszwo, line 160: We should keep this branch as the default for the applications which does not support RoutingTable.

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -145,6 +150,22 @@ LocalStream getLocal() {
...
+        final RaftGroupId groupId = request.getRaftGroupId();
+        return server.getDivision(groupId).getRaftConf().getCurrentPeers().stream()
+            .filter(p -> !p.getId().equals(server.getId())).map(RaftPeer::getId)
+            .collect(Collectors.toList());
+      } else {
+        return new ArrayList<>();
```
szetszwo, line 166: Use `Collections.emptyList()`.

szetszwo (commented):
The change looks good.  Just some minor comments.  Thanks.

ratis-client/src/main/java/org/apache/ratis/client/api/DataStreamApi.java
```diff
@@ -44,4 +48,7 @@ default DataStreamOutput stream() {
 
   /** Create a stream by providing a customized header message. */
   DataStreamOutput stream(ByteBuffer headerMessage);
+
+  /** Create a stream by providing a customized header message and route table. */
+  DataStreamOutput stream(ByteBuffer headerMessage, Map<RaftPeerId, List<RaftPeerId>> routingTable);
```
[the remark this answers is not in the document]
runzhiwang: @szetszwo RoutingTable has been defined as a interface as follows, maybe we need to defined another class ?
```
 interface RoutingTable {
    List<RaftPeerId> getSuccessors(RaftPeerId peer);
  }
```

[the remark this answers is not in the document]
szetszwo: Let's add a Builder.
```
public interface RoutingTable {
  Set<RaftPeerId> getSuccessors(RaftPeerId peerId);

  class Builder {
    private final Map<RaftPeerId, Set<RaftPeerId>> map = new HashMap<>();

    private Set<RaftPeerId> computeIfAbsent(RaftPeerId peerId) {
      return map.computeIfAbsent(peerId, key -> new HashSet<>());
    }

    public Builder addSuccessor(RaftPeerId peerId, RaftPeerId successor) {
      computeIfAbsent(peerId).add(successor);
      return this;
    }

    public Builder addSuccessors(RaftPeerId peerId, List<RaftPeerId> successors) {
      computeIfAbsent(peerId).addAll(successors);
      return this;
    }

    public Builder addSuccessors(RaftPeerId peerId, RaftPeerId... successors) {
      return addSuccessors(peerId, Arrays.asList(successors));
    }

    public RoutingTable build() {
      return peerId -> Optional.ofNullable(map.get(peerId)).orElseGet(Collections::emptySet);
    }
  }
}
```
In the code, we should keeping using RoutingTable instead of Map.  It will give us more flexibility.

Btw, it should return a Set instead of List.  My bad!

szetszwo:
RoutingTable is a small interface to encapsulate the getSuccessors(..) method.  We want to use it instead of Map since we don't have to build a Map in some cases.  E.g. the default uses a RaftConfigurable to return the other peer ids.

BTW, StreamInfo does not have to implement RoutingTable.  It may just provide a getSuccessors method as below.
```
    // StreamInfo
    private Set<RaftPeer> getSuccessors(RaftPeerId peerId, RaftConfiguration conf) {
      final RoutingTable routingTable = ...;
      if (routingTable != null) {
        return routingTable.getSuccessors(peerId).stream().map(conf::getPeer).collect(Collectors.toSet());
      } if (isPrimary()) {
        return conf.getCurrentPeers().stream()
            .filter(p -> !p.getId().equals(server.getId()))
            .collect(Collectors.toSet());
      } else {
        return Collections.emptySet();
      }
    }
```

ratis-client/src/main/java/org/apache/ratis/client/api/DataStreamApi.java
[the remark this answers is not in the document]
runzhiwang: @szetszwo Maybe we also need to expose the map to user, because we need encode map to proto, the lambda `peerId -> Optional.ofNullable(map.get(peerId)).orElseGet(Collections::emptySet)` is not enough.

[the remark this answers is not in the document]
szetszwo: That is a good point.  We should add a toProto() method to the interface.
```
  public interface RoutingTable {
    Set<RaftPeerId> getSuccessors(RaftPeerId peerId);

    RoutingTableProto toProto();
  }
```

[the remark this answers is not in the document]
runzhiwang: @szetszwo Sorry, not familar with Java, but the following code compiled error.
```
public interface RoutingTable {
  Set<RaftPeerId> getSuccessors(RaftPeerId peerId);

  RoutingTableProto toProto();

  class Builder {
    private final Map<RaftPeerId, Set<RaftPeerId>> map = new HashMap<>();

    private Set<RaftPeerId> computeIfAbsent(RaftPeerId peerId) {
      return map.computeIfAbsent(peerId, key -> new HashSet<>());
    }

    public Builder addSuccessor(RaftPeerId peerId, RaftPeerId successor) {
      computeIfAbsent(peerId).add(successor);
      return this;
    }

    public Builder addSuccessors(RaftPeerId peerId, List<RaftPeerId> successors) {
      computeIfAbsent(peerId).addAll(successors);
      return this;
    }

    public Builder addSuccessors(RaftPeerId peerId, RaftPeerId... successors) {
      return addSuccessors(peerId, Arrays.asList(successors));
    }

    public RoutingTable build() {
      return peerId -> Optional.ofNullable(map.get(peerId)).orElseGet(Collections::emptySet);
    }
  }
}
```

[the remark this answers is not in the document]
szetszwo: The build() method needs to be updated.  We should also change the Builder to use a reference so that it cannot be built more than once.
```
public interface RoutingTable {
  Set<RaftPeerId> getSuccessors(RaftPeerId peerId);

  RoutingTableProto toProto();

  class Builder {
    private final AtomicReference<Map<RaftPeerId, Set<RaftPeerId>>> ref = new AtomicReference<>(new HashMap<>());

    private Set<RaftPeerId> computeIfAbsent(RaftPeerId peerId) {
      return Optional.ofNullable(ref.get())
          .map(map -> map.computeIfAbsent(peerId, key -> new HashSet<>()))
          .orElseThrow(() -> new IllegalStateException("Already built"));
    }

    public Builder addSuccessor(RaftPeerId peerId, RaftPeerId successor) {
      computeIfAbsent(peerId).add(successor);
      return this;
    }

    public Builder addSuccessors(RaftPeerId peerId, Collection<RaftPeerId> successors) {
      computeIfAbsent(peerId).addAll(successors);
      return this;
    }

    public Builder addSuccessors(RaftPeerId peerId, RaftPeerId... successors) {
      return addSuccessors(peerId, Arrays.asList(successors));
    }

    public RoutingTable build() {
      return Optional.ofNullable(ref.getAndSet(null))
          .map(RoutingTable::newRoutingTable)
          .orElseThrow(() -> new IllegalStateException("Already built"));
    }
  }

  static RoutingTable newRoutingTable(Map<RaftPeerId, Set<RaftPeerId>> map){
    return new RoutingTable() {
      @Override
      public Set<RaftPeerId> getSuccessors(RaftPeerId peerId) {
        return Optional.ofNullable(map.get(peerId)).orElseGet(Collections::emptySet);
      }

      @Override
      public RoutingTableProto toProto() {
        return RoutingTableProto.newBuilder().addAllRoutingTable(ProtoUtils.toRouteProtos(map)).build();
      }
    };
  }
}
```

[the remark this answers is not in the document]
runzhiwang: @szetszwo Thanks, really learn a lot.

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -145,6 +148,26 @@ LocalStream getLocal() {
...
       return JavaUtils.getClassSimpleName(getClass()) + ":" + request;
     }
+
+    private Set<RaftPeer> getSuccessors(RaftPeerId peerId) throws IOException {
+      final RoutingTable routingTable = request.getRoutingTable();
+      final RaftGroupId groupId = request.getRaftGroupId();
+      final RaftConfiguration conf = server.getDivision(groupId).getRaftConf();
+      Set<RaftPeer> successors = routingTable.getSuccessors(peerId).stream().map(conf::getPeer)
+          .collect(Collectors.toSet());
+      if (successors.size() > 0) {
```
szetszwo, line 158: If successors.size() == 0, either there is no routingTable, or there is no successors in the route.  We need to handle this two cases separately.

When there is no routing table, we should change request.getRoutingTable() to return null.

ratis-client/src/main/java/org/apache/ratis/client/impl/ClientProtoUtils.java
```diff
@@ -118,14 +118,23 @@
   static RaftClientRequest toRaftClientRequest(RaftClientRequestProto p) {
     final RaftClientRequest.Type type = toRaftClientRequestType(p);
     final RaftRpcRequestProto request = p.getRpcRequest();
+    RoutingTable.Builder routingTablebuilder = new RoutingTable.Builder();
```
szetszwo, line 121: Check `p.hasRoutingTable()`.  Pass null if `p.hasRoutingTable() == false`.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftClientRequest.java
```diff
@@ -235,17 +235,28 @@ public static RaftClientRequest toWriteRequest(RaftClientRequest r, Message mess
...
+  @SuppressWarnings("parameternumber")
+  public RaftClientRequest(
+      ClientId clientId, RaftPeerId serverId, RaftGroupId groupId,
+      long callId, Message message, Type type, SlidingWindowEntry slidingWindowEntry,
+      RoutingTable routingTable) {
     super(clientId, serverId, groupId, callId);
     this.message = message;
     this.type = type;
     this.slidingWindowEntry = slidingWindowEntry != null? slidingWindowEntry: SlidingWindowEntry.getDefaultInstance();
+    this.routingTable = routingTable != null ? routingTable : new RoutingTable.Builder().build();
```
szetszwo, line 259: Just set this.routingTable to null if there is no routingTable.

ratis-common/src/main/java/org/apache/ratis/protocol/RoutingTable.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
+
+public interface RoutingTable {
+  Set<RaftPeerId> getSuccessors(RaftPeerId peerId);
+
+  RoutingTableProto toProto();
+
+  class Builder {
```
szetszwo, line 38: I forgot to add a `newBuilder()` method and a private constructor.
```
      private Builder() {}
```
Please add them.

szetszwo (commented):
Thanks for the update.  Some comments inlined.

szetszwo (approved):
+1 the change looks good.

PR #368, opened 2020-12-21
RATIS-1253. FileStore support routing table

runzhiwang:

FileStore support routing table

https://issues.apache.org/jira/browse/RATIS-1253

## How was this patch tested?

no need new ut


szetszwo (approved):
+1 the change looks good.

PR #384, opened 2020-12-29
RATIS-1274. Close DataStreamClient when close RaftClient

runzhiwang:

When create dataStreamApi in RaftClientImpl, we need to create DataStreamClient, so we need to  close DataStreamClient when close RaftClient. 

https://issues.apache.org/jira/browse/RATIS-1274

## How was this patch tested?

no need new ut


szetszwo (approved):
+1 the change looks good.

runzhiwang:
@szetszwo Thanks for review. Because CI passed, so I merge the patch.

PR #387, opened 2020-12-30
RATIS-1276. Use ClientInvocationId as key instead of StreamMap.Key

runzhiwang:

Use ClientInvocationId as key instead of StreamMap.Key

https://issues.apache.org/jira/browse/RATIS-1276

## How was this patch tested?

no need new ut


szetszwo (approved):
+1 the change looks good.

runzhiwang:
@szetszwo Thanks for review.

PR #389, opened 2020-12-31
RATIS-1277. Fix FileStore write failed because reply out of order

runzhiwang:

Fix FileStore write failed because out of order

https://issues.apache.org/jira/browse/RATIS-1277

## How was this patch tested?

test it manually


szetszwo (approved):
+1 the change looks good.

PR #393, opened 2021-01-08
RATIS-1281. TestRetryCacheWithGrpc#testRetryOnResourceUnavailableException fails

ferhui:

(Please fill in changes proposed in this fix)

## How was this patch tested?

### DIFF
--- a/ratis-test/src/test/java/org/apache/ratis/grpc/TestRetryCacheWithGrpc.java
+++ b/ratis-test/src/test/java/org/apache/ratis/grpc/TestRetryCacheWithGrpc.java
@@ -81,9 +81,12 @@ public class TestRetryCacheWithGrpc
     while (failure.get()) {
       try {
         // retry until the request failed with ResourceUnavailableException succeeds.
-        leaderProxy.submitClientRequestAsync(r).get();
+        RaftClientReply reply = leaderProxy.submitClientRequestAsync(r).get();
+        if (reply.isSuccess()) {
+          failure.set(false);
+        }
       } catch (Exception e) {
-        failure.set(false);
+        // Ignore the exception
       }
     }
     cluster.shutdown();


ferhui:
@lokeshj1703 Hi, i see you add this test, could you please take a look? Thanks

ferhui:
Detail about this PR, I have post it to jira, and paste it here.

The first while clause in TestRetryCacheWithGrpc#testRetryOnResourceUnavailableException
```
while (!failure.get()) {
  long cid = callId;
  r = cluster.newRaftClientRequest(clientId, leaderProxy.getId(), callId++,
      new RaftTestUtil.SimpleMessage("message"));
  CompletableFuture<RaftClientReply> f = leaderProxy.submitClientRequestAsync(r);
  f.exceptionally(e -> {
      if (e.getCause() instanceof ResourceUnavailableException) {
        RetryCacheTestUtil.isFailed(RetryCacheTestUtil.get(leader, clientId, cid));
        failure.set(true);
      }
      return null;
  });
}
```

There are 2 requests here, and because followers call blockWriteStateMachineData, the first request will be blocked, and ELEMENT_LIMIT_KEY is 1, the second request will fail and receive ResourceUnavailableException.

After quittng the first while clause. Followers call unblockWriteStateMachineData

In the second while clause
```
while (failure.get()) {
  try {
  // retry until the request failed with ResourceUnavailableException succeeds.
  leaderProxy.submitClientRequestAsync(r).get();
  } catch (Exception e) {
    failure.set(false);
  }
}
```

There are 2 cases: 1. If the first request has been done, the 3rd request will be handled, and later requests will hit retrycache, and server will always return success. It couldn't get Exception and enter infinite loop. 2. If the first request hasn't been done, and the 3rd request has been reach the server, it will get   ResourceUnavailableException and set failure true, and quit the loop.

Mostly the case 1 occurs, and this UT nearly always fails

So i suggest that quit the loop when request handled successfully as the comments say

> // retry until the request failed with ResourceUnavailableException succeeds.

lokeshj1703 (approved):
@ferhui Thanks for working on this! The changes look good to me. +1.

lokeshj1703:
@ferhui Can you please update the PR description as well?

lokeshj1703:
@ferhui Thanks for the contribution! I have committed the PR to master branch. The test failures in CI are unrelated as PR only has changes in a test file.

ferhui:
@lokeshj1703 Thanks for review!

PR #394, opened 2021-01-13
RATIS-1286. Move setConfiguration and transferLeadership to admin GRPC calls

elek:

Discussed [here](https://lists.apache.org/thread.html/r51e13b73d6f3de8ffe2fe5db497535d816e3b1d5ff6f9256f954bf05%40%3Cdev.ratis.apache.org%3E) on dev@ mailing list.

Ratis GRPC has multiple GRPC services:

  * RaftClientProtocolService
  * RaftServerProtocolService
  * AdminProtocolService

If all of them use the same port / netty server a malicious client/user 
may send admin/setConfiguration requests to the servers adding more 
groups or re-configuring existing ones.

For example if somebody implements any authentication/authorization on 
the StateMachine level, it can be ignored by adding more nodes and 
replicating the raw Ratis data.

I suggest improving the GrpcService.java and add only the configured 
services. With this approach mTLS can be turned on for admin and 
server2server communication but not for the client.

The only problem is the RaftClientProtocolService.setConfiguration. It's 
an un-safe operation and more-like an admin command.

This patch moves admin related methods from client to admin protocol.

https://issues.apache.org/jira/browse/RATIS-1286

## How was this patch tested?

Logic has not been changed just the organization of a few methods. Existing unit tests should cover it.

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/protocol/AdminAsynchronousProtocol.java
+++ b/ratis-common/src/main/java/org/apache/ratis/protocol/AdminAsynchronousProtocol.java
@@ -17,6 +17,7 @@
  */
 package org.apache.ratis.protocol;
 
+import java.io.IOException;
 import java.util.concurrent.CompletableFuture;
 
 /** Asynchronous version of {@link AdminProtocol}. */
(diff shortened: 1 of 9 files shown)


runzhiwang (approved):
LGTM

PR #396, opened 2021-01-19
RATIS-1289. Support skip the storagedir with multiple volumes which cause by bad disk

cchenax:

Support skip the storagedir with multiple volumes which cause by bad disk
the problem statement:
when ratis created the storagedir on the bad disk,it would throw java.nio.file.AccessDeniedException

https://issues.apache.org/jira/browse/RATIS-1289

## How was this patch tested?

I started the Ozone server which the ratis was replaced by my local changed ratis, when the one of volumes thrown the IOException caused by bad disk ,this volume would be skipped

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java
@@ -104,15 +104,16 @@ class ServerState implements Closeable {
     LOG.info("{}: {}", getMemberId(), configurationManager);
 
     List<File> directories = RaftServerConfigKeys.storageDir(prop);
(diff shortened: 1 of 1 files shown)


amaliujia:
I think such permission issue shouldn't be handled by code. It is better to fix permission outside of Ratis.

runzhiwang (approved):
LGTM

runzhiwang:
@amaliujia cc, this mainly fixed for bad disk, not permission, but the exception is same.

amaliujia:
+1 for bad disk it is ok.

runzhiwang:
@cchenax Thanks the patch. @amaliujia Thanks for review. I have merged it.

PR #398, opened 2021-01-20
RATIS-1291. Send heartbeat when there is no reply

runzhiwang:

Currently, when there always exist log, leader only send log, and will not send heartbeat. The problem is if follower need a 
long time to process log, leader maybe receives response from follower after a long time.
```
  default long getHeartbeatRemainingTimeMs() {
    return getServer().properties().minRpcTimeoutMs()/2 - getFollower().getLastRpcTime().elapsedTimeMs();
  }
```

In this pr, I record lastRpcSendTimeWithResponse, i.e. when leader receive response of request1, leader record the send time of request1. Besides, I record lastHeartBeatSendTime, i.e. the send time of heartbeat.

Use this two timestamp, leader send heartbeat when can not receive response after minRpcTimeoutMs()/2. If log cost a 
short time from request to reply, leader also does not need to send heartbeat.

```
  default long getHeartbeatRemainingTimeMs() {
    return getServer().properties().minRpcTimeoutMs()/2 -
       Math.min(getFollower().getLastRpcSendTimeWithResponse().elapsedTimeMs(),
           getFollower().getLastHeartBeatSendTime().elapsedTimeMs());
  }
```

https://issues.apache.org/jira/browse/RATIS-1291

## How was this patch tested?

no need new ut

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -143,7 +143,6 @@ public class GrpcLogAppender extends LogAppenderBase {
         }
 
         appendLog(installSnapshotRequired || haveTooManyPendingRequests());
-
       }
       getLeaderState().checkHealth(getFollower());
     }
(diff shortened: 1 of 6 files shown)


runzhiwang:
will fix the failed ut.

amaliujia:
cc @amaliujia  I can help review when the UT is fixed.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -143,7 +143,6 @@ public void run() throws IOException {
         }
 
         appendLog(installSnapshotRequired || haveTooManyPendingRequests());
-
```
amaliujia, line 146: nit: undo this change.

spacemonkd:
Hi @runzhiwang I see that this PR has been open for a long time and https://github.com/apache/ratis/pull/383 is dependent on this PR.

Were you able to get time to fix the UT failures?
In general RATIS-1273 would be good to have.

PR #401, opened 2021-01-21
RATIS-1290. Allow separate admin and client service

adoroszlai:

Make it possible to expose client, server and admin services on different ports and with different TLS configuration.

https://issues.apache.org/jira/browse/RATIS-1290

## How was this patch tested?

* unit test coverage: updated `MiniRaftCluster` to use different ports
* examples: tested with arithmetic and filestore
* integrated with Ozone (locally for now)

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
+++ b/ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
@@ -56,9 +56,23 @@ public final class RaftPeer {
     return new Builder();
   }
 
+  /** @return a new builder initialized from {@code peer} */
+  public static Builder newBuilder(RaftPeer peer) {
+    Objects.requireNonNull(peer, "peer == null");
+    return newBuilder()
+        .setId(peer.getId())
+        .setAddress(peer.getAddress())
+        .setAdminAddress(peer.getAdminAddress())
+        .setClientAddress(peer.getClientAddress())
+        .setDataStreamAddress(peer.getDataStreamAddress())
+        .setPriority(peer.getPriority());
+  }
+
   public static class Builder {
     private RaftPeerId id;
     private String address;
+    private String adminAddress;
+    private String clientAddress;
     private String dataStreamAddress;
     private int priority;
 
@@ -84,6 +98,24 @@ public final class RaftPeer {
       return setAddress(NetUtils.address2String(address));
     }
 
+    public Builder setAdminAddress(String addr) {
+      this.adminAddress = addr;
+      return this;
+    }
+
+    public Builder setAdminAddress(InetSocketAddress addr) {
+      return setAdminAddress(NetUtils.address2String(addr));
+    }
+
+    public Builder setClientAddress(String addr) {
+      this.clientAddress = addr;
+      return this;
+    }
+
+    public Builder setClientAddress(InetSocketAddress addr) {
+      return setClientAddress(NetUtils.address2String(addr));
+    }
+
     public Builder setDataStreamAddress(String dataStreamAddress) {
       this.dataStreamAddress = dataStreamAddress;
       return this;
@@ -104,7 +136,7 @@ public final class RaftPeer {
     public RaftPeer build() {
       return new RaftPeer(
           Objects.requireNonNull(id, "The 'id' field is not initialized."),
-          address, dataStreamAddress, priority);
+          address, adminAddress, clientAddress, dataStreamAddress, priority);
     }
   }
 
@@ -112,6 +144,8 @@ public final class RaftPeer {
   private final RaftPeerId id;
   /** The RPC address of the peer. */
   private final String address;
+  private final String adminAddress;
+  private final String clientAddress;
   /** The DataStream address of the peer. */
   private final String dataStreamAddress;
   /** The priority of the peer. */
@@ -119,10 +153,14 @@ public final class RaftPeer {
 
   private final Supplier<RaftPeerProto> raftPeerProto;
 
-  private RaftPeer(RaftPeerId id, String address, String dataStreamAddress, int priority) {
+  private RaftPeer(RaftPeerId id,
+      String address, String adminAddress, String clientAddress, String dataStreamAddress,
+      int priority) {
     this.id = Objects.requireNonNull(id, "id == null");
     this.address = address;
     this.dataStreamAddress = dataStreamAddress;
+    this.adminAddress = adminAddress;
+    this.clientAddress = clientAddress;
     this.priority = priority;
     this.raftPeerProto = JavaUtils.memoize(this::buildRaftPeerProto);
   }
@@ -132,6 +170,8 @@ public final class RaftPeer {
         .setId(getId().toByteString());
     Optional.ofNullable(getAddress()).ifPresent(builder::setAddress);
     Optional.ofNullable(getDataStreamAddress()).ifPresent(builder::setDataStreamAddress);
+    Optional.ofNullable(getClientAddress()).ifPresent(builder::setClientAddress);
+    Optional.ofNullable(getAdminAddress()).ifPresent(builder::setAdminAddress);
     builder.setPriority(priority);
     return builder.build();
   }
@@ -141,9 +181,19 @@ public final class RaftPeer {
     return id;
   }
 
-  /** @return The RPC address of the peer. */
+  /** @return The RPC address of the peer for server-server communication. */
   public String getAddress() {
     return address;
+  }
+
+  /** @return The RPC address of the peer for admin operations. */
+  public String getAdminAddress() {
+    return adminAddress != null ? adminAddress : address;
+  }
+
+  /** @return The RPC address of the peer for client operations. */
+  public String getClientAddress() {
+    return clientAddress != null ? clientAddress : address;
   }
 
   /** @return The data stream address of the peer. */
@@ -163,9 +213,13 @@ public final class RaftPeer {
   @Override
   public String toString() {
     final String rpc = address != null? "|rpc:" + address: "";
+    final String admin = adminAddress != null && !Objects.equals(address, adminAddress)
+        ? "|admin:" + adminAddress : "";
+    final String client = clientAddress != null && !Objects.equals(address, clientAddress)
+        ? "|client:" + clientAddress : "";
     final String data = dataStreamAddress != null? "|dataStream:" + dataStreamAddress: "";
     final String p = "|priority:" +  priority;
-    return id + rpc + data + p;
+    return id + rpc + admin + client + data + p;
   }
 
   @Override
--- a/ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
+++ b/ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
@@ -186,14 +186,16 @@ public final class RaftPeer {
     return address;
   }
 
-  /** @return The RPC address of the peer for admin operations. */
+  /** @return The RPC address of the peer for admin operations.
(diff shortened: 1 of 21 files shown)


ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
```diff
@@ -132,6 +170,8 @@ private RaftPeerProto buildRaftPeerProto() {
         .setId(getId().toByteString());
     Optional.ofNullable(getAddress()).ifPresent(builder::setAddress);
     Optional.ofNullable(getDataStreamAddress()).ifPresent(builder::setDataStreamAddress);
+    Optional.ofNullable(getClientAddress()).ifPresent(builder::setClientAddress);
+    Optional.ofNullable(getAdminAddress()).ifPresent(builder::setAdminAddress);
```
amaliujia, line 174: Is there a need in build() to check whether these address are different or it is ok to reuse the same address for these fields (I am not sure, just asking)

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
```diff
@@ -82,8 +103,10 @@ public static Builder newBuilder() {
     return new Builder();
   }
 
-  private final Server server;
+  private final Set<Server> servers = new LinkedHashSet<>(3);
```
szetszwo, line 106: The Server class does not override hashCode and equal so that the behavior may be unexpected.  Let's use a prot-to-Server Map instead?

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
```diff
@@ -118,23 +144,76 @@ private GrpcService(RaftServer raftServer, Supplier<RaftPeerId> idSupplier, int
...
 
+    final int port = serverConfig.getPort();
     this.serverInterceptor = new MetricServerInterceptor(
         idSupplier,
         JavaUtils.getClassSimpleName(getClass()) + "_" + port
     );
 
-    NettyServerBuilder nettyServerBuilder = NettyServerBuilder.forPort(port)
+    final boolean separateAdminServer = !serverConfig.equals(adminServerConfig);
+    final boolean separateClientServer = !serverConfig.equals(clientServerConfig);
```
szetszwo, line 154: Let's compare getPort(), i.e.
```
    final boolean separateAdminServer = adminServerConfig.getPort() != port;
    final boolean separateClientServer = clientServerConfig.getPort() != port;
```

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
```diff
@@ -118,23 +144,76 @@ private GrpcService(RaftServer raftServer, Supplier<RaftPeerId> idSupplier, int
...
+
+    final NettyServerBuilder serverBuilder =
+        startBuildingNettyServer(serverConfig, grpcMessageSizeMax, flowControlWindow);
+    serverBuilder.addService(ServerInterceptors.intercept(
+        new GrpcServerProtocolService(idSupplier, raftServer), serverInterceptor));
+    if (!separateAdminServer) {
+      addAdminService(raftServer, serverBuilder);
```
szetszwo, line 161: Let's move this line down to the else-clause (line 179) below?

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
```diff
@@ -118,23 +144,76 @@ private GrpcService(RaftServer raftServer, Supplier<RaftPeerId> idSupplier, int
...
+    }
+    if (!separateClientServer) {
+      addClientService(serverBuilder);
```
szetszwo, line 164: Similarly, let's move this line down to the else-clause (line 190) below?

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
```diff
@@ -231,4 +323,40 @@ public StartLeaderElectionReplyProto startLeaderElection(StartLeaderElectionRequ
...
+        return false;
+      }
+      ServerConfig that = (ServerConfig) o;
+      return port == that.port && port > 0;
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(port);
+    }
```
szetszwo, line 360: We may compare getPort() directly as mentioned above.  Then, we don't have to override equals(..) and hashCode() here.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
```diff
@@ -141,11 +181,21 @@ public RaftPeerId getId() {
...
 
-  /** @return The RPC address of the peer. */
+  /** @return The RPC address of the peer for server-server communication. */
   public String getAddress() {
     return address;
   }
 
+  /** @return The RPC address of the peer for admin operations. */
+  public String getAdminAddress() {
+    return adminAddress != null ? adminAddress : address;
```
szetszwo, line 191: getAdminAddress() should just return adminAddress.  The callers should fallback to use getAddress() when getAdminAddress() returns null.  

image

In the current usages, we already use Optional.ofNullable(..) to handle the case getAdminAddress() returning null.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
```diff
@@ -141,11 +181,21 @@ public RaftPeerId getId() {
...
+  }
+
+  /** @return The RPC address of the peer for client operations. */
+  public String getClientAddress() {
+    return clientAddress != null ? clientAddress : address;
```
szetszwo, line 196: Same as getAdminAddress(), getClientAddress() should just return clientAddress.

szetszwo (changes requested):
@adoroszlai , thanks a lot for working on this.  Some comments inlined.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
```diff
@@ -163,9 +213,13 @@ public RaftPeerProto getRaftPeerProto() {
   @Override
   public String toString() {
     final String rpc = address != null? "|rpc:" + address: "";
+    final String admin = adminAddress != null && !Objects.equals(address, adminAddress)
+        ? "|admin:" + adminAddress : "";
+    final String client = clientAddress != null && !Objects.equals(address, clientAddress)
+        ? "|client:" + clientAddress : "";
```
amaliujia, line 219: Nit: I think it is more readable to move "|admin:" and "|client:" before the `?` thus these string will always be generated.

ratis-examples/src/main/bin/common.sh
```diff
@@ -38,7 +38,7 @@ fi
 
 echo "Found ${ARTIFACT}"
 
-QUORUM_OPTS="--peers n0:localhost:6000,n1:localhost:6001,n2:localhost:6002"
+QUORUM_OPTS="--peers n0:localhost:6000:6010:6020:6030,n1:localhost:6001:6011:6021:6031,n2:localhost:6002:6012:6022:6032"
```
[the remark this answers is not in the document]
amaliujia: +1

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
amaliujia: I guess this is answered before: client and admin can reuse address.

adoroszlai: Yes, they share address by default for backward compatibility.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
adoroszlai: Should I get rid of `ServerConfig` class completely?  Doesn't seem to add much value after removing `equals()`: it only bundles 2 parameters for the `private GrpcService` constructor.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
adoroszlai: I don't think we can move it after `serverBuilder.build()` call.

ratis-examples/src/main/bin/common.sh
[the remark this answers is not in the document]
adoroszlai: > The port setting should be optional, i.e. when there is only a single port, it will be used for both client and admin.

It is optional.  I changed it here to prove it works, but I'm OK reverting it.

ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcFactory.java
```diff
@@ -61,17 +64,41 @@ public static Parameters newRaftParameters(GrpcTlsConfig conf) {
...
+        GrpcConfigKeys.Server.tlsConf(parameters),
+        GrpcConfigKeys.TLS.conf(parameters));
   }
 
   public GrpcFactory(GrpcTlsConfig tlsConfig) {
+    this(null, null, null, tlsConfig);
+  }
+
+  private GrpcFactory(GrpcTlsConfig adminTlsConfig, GrpcTlsConfig clientTlsConfig,
+      GrpcTlsConfig serverTlsConfig, GrpcTlsConfig tlsConfig) {
```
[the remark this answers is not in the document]
adoroszlai: Done.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
adoroszlai: My intention was to rely on reference equality here.  But since items are added to `servers` only if they are different (using the `separateAdminServer` etc. flags), we don't really need a `Set`, so I changed it to a `List`.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
adoroszlai: Good idea, done.  Thanks.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
adoroszlai: For brevity I'd rather print it only if it is different.

adoroszlai:
Thanks @szetszwo and @amaliujia for the review.  I have updated the patch with your suggestions (except where noted).

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
szetszwo: I am fine for removing ServerConfig if it is not useful.

adoroszlai:
Checking `TestMetaServer` failure.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcService.java
szetszwo: Good point.  Please ignore my previous comment and keep it this way.

szetszwo (approved):
+1 the change looks good.

amaliujia:
LGTM thanks!

adoroszlai:
Thanks again @szetszwo and @amaliujia for the review.

PR #402, opened 2021-01-25
RATIS-1295. Use a config to enable/disable pre vote

runzhiwang:

Use a config to enable/disable pre vote

https://issues.apache.org/jira/browse/RATIS-1295

## How was this patch tested?

no need to add new ut

### DIFF
--- a/ratis-server-api/src/main/java/org/apache/ratis/server/RaftServerConfigKeys.java
+++ b/ratis-server-api/src/main/java/org/apache/ratis/server/RaftServerConfigKeys.java
@@ -556,14 +556,14 @@ public interface RaftServerConfigKeys {
       setTimeDuration(properties::setTimeDuration, LEADER_STEP_DOWN_WAIT_TIME_KEY, noLeaderTimeout);
     }
 
-    String ENABLE_PRE_VOTE_KEY = PREFIX + ".enable.pre-vote";
-    boolean ENABLE_PRE_VOTE_DEFAULT = true;
-    static boolean isEnablePreVote(RaftProperties properties) {
-      return getBoolean(properties::getBoolean, ENABLE_PRE_VOTE_KEY, ENABLE_PRE_VOTE_DEFAULT, getDefaultLog());
+    String PRE_VOTE_KEY = PREFIX + ".pre-vote";
+    boolean PRE_VOTE_DEFAULT = true;
+    static boolean preVote(RaftProperties properties) {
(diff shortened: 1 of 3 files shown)


amaliujia:
Question: shouldn't pre-vote is always a good optimization to accelerate leader election? Is there a situation that it is not suitable thus needs to be disabled?

runzhiwang:
@amaliujia Please see here.
image

amaliujia:
Thanks for explanation! Indeed for larger system it is better to benchmark such changes before adopting it.

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServerConfigKeys.java
```diff
@@ -555,6 +555,16 @@ static TimeDuration leaderStepDownWaitTime(RaftProperties properties) {
     static void setLeaderStepDownWaitTime(RaftProperties properties, TimeDuration noLeaderTimeout) {
       setTimeDuration(properties::setTimeDuration, LEADER_STEP_DOWN_WAIT_TIME_KEY, noLeaderTimeout);
     }
+
+    String PRE_VOTE_KEY = PREFIX + ".pre-vote";
+    boolean PRE_VOTE_DEFAULT = true;
```
amaliujia, line 560: Nit: might be better to make this default value as `false` so downstream applications (e.g. Ozone) won't feel surprising before benchmarking this optimization.

amaliujia:
Overall LGTM.

We can discuss what could be the default value.

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServerConfigKeys.java
runzhiwang: If we use false, when we change to true ? I think default true maybe better.

runzhiwang: @szetszwo What do you think ?

szetszwo: I do not have a strong opinion.  It seems okay to set true as the default.

szetszwo (approved):
+1 the change looks good.

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServerConfigKeys.java
amaliujia: Sure we can start from `true` then

runzhiwang:
@szetszwo @amaliujia Thanks for review. I have merged the patch.

PR #406, opened 2021-01-27
RATIS-1285. Fix download link point to mirror links instead of download area directly

runzhiwang:
image


runzhiwang:
@szetszwo @umamaheswararao Could you help review this ? Thanks a lot.

runzhiwang:
@szetszwo Hi, Can I merge this ? This is only generated code.

runzhiwang:
Because code of asf-site-source has been merged, this patch only about generated code, so I marge it.

PR #411, opened 2021-01-29
Fixes problems with large snapshots

alpapad:

Hi, 
I have been testing ratis for a solutions which uses rather largish snapshots for the state machine. Current implementation is not able to transfer files larger than 8Mb (I think), throwing all kinds of exceptions.  I guess this is something not really used by current users of ratis.

Here are the places I modified/found to be problematic (code is more precise than words):
FileChunkReader.readFileChunk uses ByteString.readFrom(in, chunkLength); which will fully load the input stream and not just a junk of size chunkLength

Grpc.proto service installSnapshot(stream ratis.common.InstallSnapshotRequestProto) should return a stream

Modifications to SnapshotManager so chunks for the same snapshot will be placed in the same directory 

Not sure if I need to create a jira issue (never used apache f. jira) I can  repro and create jiras with stacktraces etc.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/storage/FileChunkReader.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/FileChunkReader.java
@@ -25,6 +25,7 @@ import java.io.Closeable;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Path;
 import java.util.Optional;
 
@@ -65,7 +66,7 @@ public class FileChunkReader implements Closeable {
   public FileChunkProto readFileChunk(int chunkMaxSize) throws IOException {
     final long remaining = info.getFileSize() - offset;
     final int chunkLength = remaining < chunkMaxSize ? (int) remaining : chunkMaxSize;
-    final ByteString data = ByteString.readFrom(in, chunkLength);
+    final ByteString data = readChunk(in, chunkLength);
 
     final FileChunkProto proto = FileChunkProto.newBuilder()
         .setFilename(relativePath.toString())
@@ -80,6 +81,32 @@ public class FileChunkReader implements Closeable {
     return proto;
   }
 
+  /**
+   * Blocks until a chunk of the given size can be made from the stream, or EOF is reached. Calls
+   * read() repeatedly in case the given stream implementation doesn't completely fill the given
+   * buffer in one read() call.
+   *
+   * @return A chunk of the desired size, or else a chunk as large as was available when end of
+   *     stream was reached. Returns null if the given stream had no more data in it.
+   */
+  private static ByteString readChunk(InputStream in, final int chunkSize) throws IOException {
+    final byte[] buf = new byte[chunkSize];
+    int bytesRead = 0;
+    while (bytesRead < chunkSize) {
+      final int count = in.read(buf, bytesRead, chunkSize - bytesRead);
+      if (count == -1) {
+        break;
+      }
+      bytesRead += count;
+    }
+
+    if (bytesRead == 0) {
+      return null;
+    }
+
+    // Always make a copy since InputStream could steal a reference to buf.
+    return ByteString.copyFrom(buf, 0, bytesRead);
+  }
   @Override
   public void close() throws IOException {
     in.close();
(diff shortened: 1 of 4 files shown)


ratis-server/src/main/java/org/apache/ratis/server/storage/FileChunkReader.java
```diff
@@ -65,7 +66,7 @@ public FileChunkReader(FileInfo info, RaftStorageDirectory directory) throws IOE
   public FileChunkProto readFileChunk(int chunkMaxSize) throws IOException {
     final long remaining = info.getFileSize() - offset;
     final int chunkLength = remaining < chunkMaxSize ? (int) remaining : chunkMaxSize;
-    final ByteString data = ByteString.readFrom(in, chunkLength);
```
alpapad, line 68: ByteString.readFrom(in, chunkLength) will read the whole file not just chunkLength

ratis-server/src/main/java/org/apache/ratis/server/storage/FileChunkReader.java
```diff
@@ -80,6 +81,32 @@ public FileChunkProto readFileChunk(int chunkMaxSize) throws IOException {
     return proto;
   }
 
+  /**
```
alpapad, line 84: Code copied from ByteString class

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
```diff
@@ -61,22 +60,23 @@ public void installSnapshot(StateMachine stateMachine,
     final long lastIncludedIndex = snapshotChunkRequest.getTermIndex().getIndex();
     final RaftStorageDirectory dir = storage.getStorageDir();
 
-    // create a unique temporary directory
-    final File tmpDir =  new File(dir.getTmpDir(), UUID.randomUUID().toString());
```
alpapad, line 65: when receiving a snapshot which does not fit in a single chunk, then multiple directories are creates, each having a chunk of the snapshot

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
```diff
@@ -130,10 +130,9 @@ public void installSnapshot(StateMachine stateMachine,
     }
 
     if (snapshotChunkRequest.getDone()) {
-      LOG.info("Install snapshot is done, renaming tnp dir:{} to:{}",
-          tmpDir, dir.getStateMachineDir());
-      dir.getStateMachineDir().delete();
-      tmpDir.renameTo(dir.getStateMachineDir());
+      LOG.info("Install snapshot is done, moving files from dir:{} to:{}", tmpDir, dir.getStateMachineDir());
```
alpapad, line 133: Handles the case where multiple files are copied

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
```diff
@@ -61,22 +60,23 @@ public void installSnapshot(StateMachine stateMachine,
...
     final RaftStorageDirectory dir = storage.getStorageDir();
 
-    // create a unique temporary directory
-    final File tmpDir =  new File(dir.getTmpDir(), UUID.randomUUID().toString());
+    // create a unique temporary directory based on the request id
+    final File tmpDir =  new File(dir.getTmpDir(), snapshotChunkRequest.getRequestId());
     FileUtils.createDirectories(tmpDir);
     tmpDir.deleteOnExit();
 
-    LOG.info("Installing snapshot:{}, to tmp dir:{}", request, tmpDir);
```
alpapad, line 69: This will log the whole chunk.. Lots of bytes..

amaliujia:
@alpapad would you mind creating a JIRA for this PR?

amaliujia:
I guess this will worth an email to user@ratis to ask the problem you are facing.

I don't think Ratis cannot handle larger than 8MB files transferred well. There is an example in Ratis repo named FileStore. You can use that example to save files by an application based on Ratis. I believe I saw people used GBs files on FileStore.

alpapad:
@amaliujia  The problem exists when a statemachine snapshot needs to be transfered (for example because of a node failure). The snapshot should be chunked to a specific chunk-size but instead the whole file is read and the server does not expect it to be more than X Mb (I think 8, but I will need to check).

I think the FileStore statemachine does not store much and file contents are not even stored in the ratis wal.

I will create a jira for this.

amaliujia:
For such optimizations, usually I think it worths a have a benchmark to identify the bottleneck and/or provide a baseline to convince people that such optimization is useful, and how much it is useful.

PR #414, opened 2021-02-02
RATIS-1304. Do not print error if ratis log is not yet created

elek:
JIRA: https://issues.apache.org/jira/browse/RATIS-1304

I found that Ozone prints out a lot of errors related to missing ratis-meta.conf, even if everything works well.

The related code path seems to be prepared to this case, but still prints out errors.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
@@ -144,9 +144,9 @@ public class RaftStorageImpl implements RaftStorage {
       LogEntryProto confProto = LogEntryProto.newBuilder().mergeFrom(fio).build();
       return LogProtoUtils.toRaftConfiguration(confProto);
     } catch (FileNotFoundException e) {
-
       return null;
     } catch (Exception e) {
+      LOG.error("Failed reading configuration from file:" + confFile, e);
       return null;
     }
   }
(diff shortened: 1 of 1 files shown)


ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
```diff
@@ -143,8 +143,10 @@ public RaftConfiguration readRaftConfiguration() {
     try (FileInputStream fio = new FileInputStream(confFile)) {
       LogEntryProto confProto = LogEntryProto.newBuilder().mergeFrom(fio).build();
       return LogProtoUtils.toRaftConfiguration(confProto);
+    } catch (FileNotFoundException e) {
+
+      return null;
     } catch (Exception e) {
-      LOG.error("Failed reading configuration from file:" + confFile, e);
       return null;
```
adoroszlai, line 150: I think we should keep the log for other exceptions, otherwise the two `catch` blocks can be merged.

amaliujia: Or actually move the LOG.error to `FileNotFoundException` as that is a error that has happened thus printing that into log is useful?

I guess to avoid printing every exception is the goal of this PR?

runzhiwang:
@elek Thanks the patch. @adoroszlai @amaliujia Thanks for review. I have merged the patch.

PR #415, opened 2021-02-04
RATIS-1306. Eliminate duplicated GitHub Actions workflow

adoroszlai:

Merge `post-commit` and `pr` workflows.  The only step which is unique to `post-commit` workflow is executed conditionally, depending on the event that triggered the run.

https://issues.apache.org/jira/browse/RATIS-1306

## How was this patch tested?

push in fork (failed only due to some temporary network issue):
https://github.com/adoroszlai/incubator-ratis/runs/1830531946


runzhiwang:
@adoroszlai Thanks the patch. I have merged it.

adoroszlai:
Thanks @runzhiwang for reviewing and committing it.

PR #419, opened 2021-02-05
RATIS-1311. Upgrade Java for Sonar check

adoroszlai:

Sonar no longer accepts reports created using Java 8.  This change upgrades to Java 11.  Check is split from unit to avoid affecting that one, and also to let them execute concurrently.

https://issues.apache.org/jira/browse/RATIS-1311

## How was this patch tested?

Tested in own fork with temporarily removed repo name comparison: https://github.com/adoroszlai/incubator-ratis/actions/runs/540063075

The check was skipped in subsequent run with repo name comparison restored: https://github.com/adoroszlai/incubator-ratis/runs/1838364223

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -95,7 +95,7 @@ jobs:
           env:
             SONAR_TOKEN: ${{ secrets.SONARCLOUD_TOKEN }}
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        - run: ./dev-support/checks/unit.sh
+        - run: ./dev-support/checks/unit.sh -Dtest=TestVerificationTool
         - name: Summary of failures
           run: cat target/${{ github.job }}/summary.txt
           if: always()
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -90,11 +90,6 @@ jobs:
             restore-keys: |
               maven-repo-${{ hashFiles('**/pom.xml') }}
               maven-repo-
-        - run: ./dev-support/checks/sonar.sh
-          if: github.repository == 'apache/incubator-ratis' && github.event_name != 'pull_request'
-          env:
-            SONAR_TOKEN: ${{ secrets.SONARCLOUD_TOKEN }}
-            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         - run: ./dev-support/checks/unit.sh
         - name: Summary of failures
           run: cat target/${{ github.job }}/summary.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/post-commit.yml
```diff
@@ -151,3 +146,28 @@ jobs:
         - name: Delete temporary build artifacts
           run: rm -rf ~/.m2/repository/org/apache/ratis
           if: always()
+  sonar:
+    name: sonar
+    runs-on: ubuntu-18.04
+    if: github.repository == 'apache/incubator-ratis' && github.event_name != 'pull_request'
```
amaliujia, line 152: Question: after Ratis graduates as TLP, will this fixed repo name need a change?

amaliujia (approved):
LGTM. Left a minor comment.

.github/workflows/post-commit.yml
adoroszlai: Thanks @amaliujia for the review.  Yes, this will need to be updated if the repo is renamed to `ratis`.  Please note however, that this hard-coded name is already present on `master`:

https://github.com/apache/incubator-ratis/blob/aa6c5ff4bad6425a7e1dd10a1f3e4f2f93f4050c/.github/workflows/post-commit.yml#L93-L94

softgitron: I have actually also investigated this issue on my personal [fork](https://github.com/softgitron/incubator-ratis/commit/7a6541f2237b434b3767dae409493be296cd3eea) and ended up with fairly similar modifications. I think main functionality on this commit is good and should be implemented as is. What comes to this specific code line, I would drop check for repository for following reasons:
- This check does not add value. Sonar analysis will still fail couple of seconds later in the sonar.sh file due to missing environment variables so it is only matter of which error message you will get.
- This makes usage of the sonar checks annoying on the forks, because line must be always patched away. (I'm not expecting that forking will be especially common for an open source project, but still improvement)

Just to remind, after Ratis graduates as TLP sonar.sh should be also updated with new `-Dsonar.organization` and `-Dsonar.projectKey` It would be nice if those fields could update themselves, but that would possibly make the script slightly too complex. So maybe it is better to leave those parameters as is. Again those parameters are slight inconvenience for forgers, but I think it would be wrong to move those also to Github secrets for example for easier configurations since they are not really secrets.

adoroszlai: Thanks @softgitron for the review.

> Sonar analysis will still fail couple of seconds later in the sonar.sh file due to missing environment variables so it is only matter of which error message you will get.

With this check the workflow simply skips the `sonar` job in forks.  If we omit the check, the workflow will fail with an error.  I don't think these two are the same, as in the second case the error is expected and needs to be worked around mentally.  So I'd keep this check.

> This makes usage of the sonar checks annoying on the forks, because line must be always patched away. (I'm not expecting that forking will be especially common for an open source project, but still improvement)

Forks are common for development, but I think only very few would register their forks in Sonar.  What is the benefit or use case for doing that?

> Just to remind, after Ratis graduates as TLP sonar.sh should be also updated with new `-Dsonar.organization` and `-Dsonar.projectKey`

I'm not very familiar with registering projects in Sonarcloud, but I don't think they are automatically renamed there.  Project key seems to be persistent.  Apache Ozone for example still goes by [`hadoop-ozone`](https://sonarcloud.io/dashboard?id=hadoop-ozone).

softgitron: Sorry, my apologies. I misunderstood and miss-evaluated some points.

> Sonar analysis will still fail couple of seconds later in the sonar.sh file due to missing environment variables so it is only matter of which error message you will get.

The worked previously as I described, but you actually made condition more usable in your patch set. Instead of just preventing environment variables to be set, condition now actually prevents whole step from running. Condition makes perfect sense now.

> This makes usage of the sonar checks annoying on the forks, because line must be always patched away. (I'm not expecting that forking will be especially common for an open source project, but still improvement)

You may notice, that I'm not especially experienced with the open source development. I always thought that development is done mainly using branches, but I just noticed that even this pull request refers to a fork. This clause refers to a hard forks where project totally divides into two different projects.

There are no significant advantages for creating separate Sonar projects for forks. Only on some special occasions this may come in handy. For example Sonar Cloud shows full coverity statistics only on the master branch. It was crucial for #423 to be able to push directly to the master to verify that everything is working as intended. Another slight advantage from having separate Sonar would be that developer could use full Sonar analysis, before even making a pull request. However, I think, this is not especially usefull thing to do.

> Just to remind, after Ratis graduates as TLP sonar.sh should be also updated with new -Dsonar.organization and -Dsonar.projectKey

Actually there is no need to update project organization after TLP status, but I expect that projectKey should be updated by removing word incubator from the key.

All in all, I think this pull request is good and can be merged as is. Maybe these comments will be useful for some future request.

softgitron:
Highly needed CI improvement. LGTM.

bshashikant (approved):

adoroszlai:
Thanks @bshashikant for reviewing and committing it.

PR #421, opened 2021-02-08
RATIS-1308. Findbugs check is failing silently

adoroszlai:

Currently Findbugs is [failing](https://github.com/apache/incubator-ratis/runs/1776425308#step:3:2360) with:

```
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  04:02 min
[INFO] Finished at: 2021-02-02T03:02:00Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.codehaus.mojo:findbugs-maven-plugin:3.0.0:findbugs (findbugs) on project ratis: Unable to parse configuration of mojo org.codehaus.mojo:findbugs-maven-plugin:3.0.0:findbugs for parameter pluginArtifacts: Cannot assign configuration entry 'pluginArtifacts' with value '${plugin.artifacts}' of type java.util.Collections.UnmodifiableRandomAccessList to property of type java.util.ArrayList -> [Help 1]
```

but the CI check is passing (green).

1. The root cause is that Findbugs 3.0.0 does not work with Maven 3.4+, so it should be upgraded to 3.0.5.  But because Findbugs is no longer under development, switch to Spotbugs (its successor) instead.
2. Fix or suppress warnings that have been introduced while Findbugs was failing silently.  `SuppressFBWarnings` annotation is used instead of XML-based exclusion (except for protobuf generated source) for the following reasons:
   * it can be applied closer to the problem spot, in the same source file
   * easier to use
   * harder to break accidentally (by renaming, etc.)

https://issues.apache.org/jira/browse/RATIS-1308

## How was this patch tested?

```
$ ./dev-support/checks/findbugs.sh
...
[INFO] BUILD SUCCESS

$ cat target/findbugs/failures
0
```

Findbugs is now successful in CI:
https://github.com/adoroszlai/incubator-ratis/runs/1855897307#step:4:798

### DIFF
--- a/dev-support/checks/findbugs.sh
+++ b/dev-support/checks/findbugs.sh
@@ -20,19 +20,19 @@ MAVEN_OPTIONS='-B -fae'
 
 if ! type unionBugs >/dev/null 2>&1 || ! type convertXmlToText >/dev/null 2>&1; then
   #shellcheck disable=SC2086
-  mvn ${MAVEN_OPTIONS} compile findbugs:check
+  mvn ${MAVEN_OPTIONS} compile spotbugs:check
   exit $?
 fi
 
 #shellcheck disable=SC2086
-mvn ${MAVEN_OPTIONS} compile findbugs:findbugs
+mvn ${MAVEN_OPTIONS} compile spotbugs:spotbugs
 rc=$?
 
 REPORT_DIR=${OUTPUT_DIR:-"$DIR/../../target/findbugs"}
 mkdir -p "$REPORT_DIR"
 REPORT_FILE="$REPORT_DIR/summary.txt"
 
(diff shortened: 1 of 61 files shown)


dineshchitlangia (approved):
+1 @adoroszlai

adoroszlai:
@szetszwo @runzhiwang @lokeshj1703 can you please review?

runzhiwang (approved):
LGTM

szetszwo (approved):
@adoroszlai , thanks a lot for working on this.  Some of the warnings should be fixed in stead of suppressed.  Let's fix them separately.

+1

adoroszlai:
Thanks @runzhiwang and @szetszwo for the review.

> Some of the warnings should be fixed in stead of suppressed. Let's fix them separately.

Agree.

PR #422, opened 2021-02-09
RATIS-1313. Use matrix build for unit tests to reduce CI time

adoroszlai:

Execute tests in 3 splits to speed up CI feedback a bit.  Split are defined via Maven profiles.  This reduces time required for _unit_ check to about 10 minutes.

https://issues.apache.org/jira/browse/RATIS-1313

## How was this patch tested?

Compared list of tests from Maven output to a previous, single run.  Verified that no tests are missed, and each of them is executed only once.

https://github.com/adoroszlai/incubator-ratis/actions/runs/551983793

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -14,7 +14,7 @@
 # limitations under the License.
 github:
   description: "Open source Java implementation for Raft consensus protocol."
-  homepage: http://ratis.incubator.apache.org/
+  homepage: http://ratis.apache.org/
   labels:
     - raft
     - consensus-protocol
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -80,6 +80,12 @@ jobs:
   unit:
     name: unit
     runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        profile:
+          - grpc
+          - server
+          - misc
     steps:
         - uses: actions/checkout@master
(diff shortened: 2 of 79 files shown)


adoroszlai:
@szetszwo @runzhiwang can someone please review?

cnauroth:
+1 (non-binding)

That's a really nice improvement.  The changes look good.  I spot checked CI output for any missing or duplicated test runs, but it all looked good to me.  Thanks, @adoroszlai !

szetszwo (approved):
+1 the change looks good.

adoroszlai:
Thanks @cnauroth, @runzhiwang, @szetszwo for the review.

szetszwo:
Hey @adoroszlai ,
image

Is it the case that if the grpc tests have failed, the server and the misc tests will be cancelled?  It seems better to always run all the tests since the overall running time is short.  This is the reason that we moved the tests to the ratis-test module; see https://issues.apache.org/jira/browse/RATIS-399

adoroszlai:
> Is it the case that if the grpc tests have failed, the server and the misc tests will be cancelled?

Correct.  I will post a follow-up change to always run all tests for `push` and `schedule` builds, but keep this fail-fast behavior for PRs.

szetszwo:
That's great.  Thanks for the fast response!

PR #440, opened 2021-03-11
RATIS-1335. Update project mailing lists to use TLP domain.

cnauroth:

The site still documents the project mailing lists in the .incubator sub-domain. Now that the project has graduated to TLP, we can update documentation of the mailing lists.

I also noticed that we did not have documentation for the -commits and -issues lists, so I added those.

Also, many of the `mailto:` links were not working correctly because of an extraneous space after the scheme.  I corrected all of those.

https://issues.apache.org/jira/browse/RATIS-1335

## How was this patch tested?

Built site locally and verified changes.
image

### DIFF
--- a/content/community.md
+++ b/content/community.md
@@ -21,10 +21,10 @@ title: Community
 
 If you'd like to contribute to Apache Ratis, please subscribe to the Ratis developer mailing list.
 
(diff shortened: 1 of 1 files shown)


szetszwo (approved):
Thanks for working on this.  It is great that you also added the other mailing lists.

+1 the change looks good.

PR #444, opened 2021-03-12
RATIS-1336. Update download page to link to artifacts at TLP path.

cnauroth:

The site still contains download links pointing to the distribution server at incubator sub-paths. Update these links to point to the new TLP location.

https://issues.apache.org/jira/browse/RATIS-1336

## How was this patch tested?

Built site locally and verified changes.
Note in the screenshot at the bottom that it no longer shows the "/incubator" sub-path for the download link.image


szetszwo (approved):
+1 the change looks good.

PR #460, opened 2021-04-14
RATIS-1356. NotifyInstallSnapshot during SetConfiguration has leader …

bharatviswa504:
…info missing.

Use when raftconf does not have any peer info, use leader info from InstallSnapshotRequestProto.

https://issues.apache.org/jira/browse/RATIS-1356

## How was this patch tested?

Updated test to not pass peer info during raft server star in raftgroup.
Also tested SCM-HA with fix now installSnapshot which happens during setConf is getting correct leaderInfo, previously used to get "".


bharatviswa504:
Fixed CI. Test failing in CI, is passing locally.

bshashikant:
The test failures are not related and will be addressed in a new PR. Merging this as this is required for SCM HA in https://github.com/apache/ozone/pull/2155.

PR #470, opened 2021-04-24
RATIS-1368 Fix server impl NPEs

softgitron:

Fix three NPE issues reported by Sonar Qube in the server impl.
- Add default to streamEndOfRequestAsync function
- Throw error on notifyStateMachine... funtion
- Remove possibility for new cache entries to be null
- Add tests for changes

https://issues.apache.org/jira/browse/RATIS-1368

## How was this patch tested?

Changes have been tested using existing test cases and two new test cases.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -1579,10 +1579,12 @@ class RaftServerImpl implements RaftServer.Division,
         // installSnapShotRequestProto.
         RoleInfoProto roleInfoProto;
         RaftPeer raftPeer = getRaftConf().getPeer(state.getLeaderId());
-        if (raftPeer == null && leaderPeerInfo != null) {
-          roleInfoProto = getRoleInfoProto(ProtoUtils.toRaftPeer(leaderPeerInfo));
-        } else if (raftPeer == null) {
-          throw new IOException("Leader peer info is unknown.");
+        if (raftPeer == null) {
+          if (leaderPeerInfo != null) {
+            roleInfoProto = getRoleInfoProto(ProtoUtils.toRaftPeer(leaderPeerInfo));
+          } else {
+            throw new IOException("Leader peer info is unknown.");
+          }
         } else {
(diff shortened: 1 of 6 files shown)


ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1560,21 +1565,28 @@ private InstallSnapshotReplyProto notifyStateMachineToInstallSnapshot(
...
         // For the cases where RaftConf is empty on newly started peer with
         // empty peer list, we retrieve leader info from
         // installSnapShotRequestProto.
-        RoleInfoProto roleInfoProto =
-            getRaftConf().getPeer(state.getLeaderId()) == null ?
-                getRoleInfoProto(ProtoUtils.toRaftPeer(leaderPeerInfo.get())) :
-                getRoleInfoProto();
+        RoleInfoProto roleInfoProto;
+        RaftPeer raftPeer = getRaftConf().getPeer(state.getLeaderId());
+        if (raftPeer == null && leaderPeerInfo != null) {
```
runzhiwang, line 1582: please use the following code:
      ```
 if (raftPeer == null) {
	    if (leaderPeerInfo != null) {
          roleInfoProto = getRoleInfoProto(ProtoUtils.toRaftPeer(leaderPeerInfo));
		} else {
          throw new IOException("Leader peer info is unknown.");
        } 
	  } else {
        roleInfoProto = getRoleInfoProto();
      }
```

File: ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java [hunk 2]
```diff
@@ -1633,7 +1645,7 @@ void submitUpdateCommitEvent() {
     // update the retry cache
     final CacheEntry cacheEntry = retryCache.getOrCreateEntry(invocationId);
     if (getInfo().isLeader()) {
-      Preconditions.assertTrue(cacheEntry != null && !cacheEntry.isCompletedNormally(),
+      Preconditions.assertTrue(!cacheEntry.isCompletedNormally(),
```
runzhiwang, line 1648: why remove cacheEntry != null ?

PR #475, opened 2021-04-28
RATIS-1372. Support multi-client in FileStore

runzhiwang:

Support multi-client in FileStore

https://issues.apache.org/jira/browse/RATIS-1372

## How was this patch tested?

test it manually


szetszwo (approved):
+1 the change looks good.

PR #476, opened 2021-04-29
RATIS-1373. Use async model for StreamInfo#applyToRemotes

runzhiwang:

We use StreamInfo#applyToRemotes to transfer data to to other server, and eventually we call OrderedStreamAsync#sendRequest. So StreamInfo#applyToRemotes finished after sending out request, but sometimes OrderedStreamAsync#sendRequest costs hundred millseconds. So we need to make StreamInfo#applyToRemotes async model, and also keep the order of request belongs to the same stream.

https://issues.apache.org/jira/browse/RATIS-1373

## How was this patch tested?

no need new ut


szetszwo (approved):
+1 patch looks good.

PR #478, opened 2021-05-12
RATIS-1376. Improve the used memory

runzhiwang:
1. Decrease outstanding-requests
2. Do not use request in lambda, because the request contains data, before lambda finish, request can not free memory.
image

https://issues.apache.org/jira/browse/RATIS-1376

## How was this patch tested?

no need new ut.


szetszwo (approved):
+1 the change looks good.

PR #486, opened 2021-06-28
RATIS-1386. Some MetricsRegistry instances never unregistered

adoroszlai:

Make sure `Metric*Interceptor` unregisters its metrics from the global registry to avoid memory leak.

https://issues.apache.org/jira/browse/RATIS-1386

## How was this patch tested?

Built Ratis and Ozone with it.  Created objects in Ozone via S3 Gateway.  Verified that no `MetricRegistryInfo` and `RatisMetricRegistryImpl` instances are retained in S3 Gateway memory for each object.

https://github.com/adoroszlai/incubator-ratis/actions/runs/979350754


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #492, opened 2021-08-17
RATIS-1394. Wrong date for 2.1.0 release

adoroszlai:

Use date from download content page header, instead of hard-coded string.

https://issues.apache.org/jira/browse/RATIS-1394

## How was this patch tested?

Verified release date for both 2.0.0 and 2.1.0 in the Download page.

```
hugo serve
open http://localhost:1313/downloads.html
```


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #493, opened 2021-08-17
RATIS-1395. Cannot use commands from download page as is

adoroszlai:

 * Fix dash in `gpg` commands
 * Fix argument of `gpg --verify` commands (pass signature file, not the tarball)

https://issues.apache.org/jira/browse/RATIS-1395

## How was this patch tested?

```
hugo serve
open http://localhost:1313/downloads.html
```

Copied and ran command:

```
$ gpg --import KEYS
...
gpg: Total number processed: 6
...
```

Also copied, adapted and ran:

```
$ gpg --verify apache-ratis-2.1.0-src.tar.gz.asc
gpg: assuming signed data in 'apache-ratis-2.1.0-src.tar.gz'
gpg: Signature made Sat Jun 26 08:17:39 2021 CEST
gpg:                using RSA key 1CEF33FA61800117BDB2E0E0D51EA8F00EE79B28
gpg: Good signature from "Marton Elek (CODE SIGNING KEY) <elek@apache.org>" [unknown]
...
```

### DIFF
--- a/layouts/custompage/downloads.html
+++ b/layouts/custompage/downloads.html
@@ -76,7 +76,7 @@ site</a>.</li>
 KEYS</a>
 file.</li>
 <li>gpg --import KEYS</li>
-<li>gpg --verify apache-ratis-X.Y.Z-src.tar.gz</li>
+<li>gpg --verify apache-ratis-X.Y.Z-src.tar.gz.asc</li>
 </ol>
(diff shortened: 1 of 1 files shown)


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #495, opened 2021-08-28
Second param is needed for gpg --verify

sebbASF:
https://www.apache.org/info/verification.html#specify_both

&ndash; is not the same as --


adoroszlai:
Thanks @sebbASF.  This is fixed in #493 (and being applied to the website in #499).

adoroszlai (approved):
Thanks @sebbASF for the additional fix.

PR #518, opened 2021-10-12
RATIS-1415. Avoid follower gap limitation for watch requests

lokeshj1703:

RATIS-1411 introduces an optimization which uses follower gap limitation to control the number of requests accepted by leader. In case the follower gap is larger the pending requests are not cleared until gap narrows.

This also impacts the watch requests though. Since watch requests already take into account majority and all commit, the follower gap limitation is not required for them.

https://issues.apache.org/jira/browse/RATIS-1415

## How was this patch tested?

Existing UT


lokeshj1703:
Thanks @szetszwo! I have addressed the review comments.

PR #547, opened 2021-11-30
RATIS-1452. Fix log output in sendRequest.

captainzmc:
see: https://issues.apache.org/jira/browse/RATIS-1452


captainzmc:
This is caused by RATIS-1445.  @szetszwo  Could you help review this?

szetszwo (approved):
+1 the change looks good.

PR #558, opened 2021-12-08
RATIS-1463 Fix verbose logging at INFO level

kerneltime:

Logging the exception does not provide useful information. There is also a redundant trace logging right after.

https://issues.apache.org/jira/browse/RATIS-1463

## How was this patch tested?

- [ ] Basic builds and tests.


szetszwo (approved):
+1 the change looks good.

PR #573, opened 2021-12-31
RATIS-1481. make state upgradate in notifyStateMachineToInstallSnapshot serialized

Xushaohong:
The  state upgrade  ```state.updateInstalledSnapshotIndex(reply);
            state.reloadStateMachine(reply.getIndex());
            installedSnapshotIndex.set(reply.getIndex()); ```

in ```notifyStateMachineToInstallSnapshot``` should be synchronized in main thread.
Currently it is executed in whenComplete stage after ```stateMachine.followerEvent().notifyInstallSnapshotFromLeader```, which could lead to two inappropriate logic. 

1. The following check in the main thread ``` if (latestInstalledSnapshotIndex > 0)``` will be executed before state upgrade, which cause the raft server could not respond SNAPSHOT_INSTALLED in one request.
2. Once the actual upgrade of snapshot index and commit index takes place through the above state upgrade, the leader receives the inconsistency reply with the new index and then won't send installsnapshot requests anymore as the check of shouldNotifyToInstallSnapshot() will be null.

https://issues.apache.org/jira/browse/RATIS-1481

## How was this patch tested?

manual test on ozone

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/SnapshotInstallationHandler.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/SnapshotInstallationHandler.java
@@ -119,13 +119,13 @@ class SnapshotInstallationHandler {
     if (reply != null) {
       if (request.hasLastRaftConfigurationLogEntryProto()) {
         // Set the configuration included in the snapshot
-        final LogEntryProto newConfLogEntryProto = request.getLastRaftConfigurationLogEntryProto();
-        if (!state.getRaftConf().equals(LogProtoUtils.toRaftConfiguration(newConfLogEntryProto))) {
-          LOG.info("{}: set new configuration {} from snapshot", getMemberId(), newConfLogEntryProto);
-          state.setRaftConf(newConfLogEntryProto);
-          state.writeRaftConfiguration(newConfLogEntryProto);
-          server.getStateMachine().event().notifyConfigurationChanged(newConfLogEntryProto.getTerm(),
-              newConfLogEntryProto.getIndex(), newConfLogEntryProto.getConfigurationEntry());
+        final LogEntryProto proto = request.getLastRaftConfigurationLogEntryProto();
+        if (!state.getRaftConf().equals(LogProtoUtils.toRaftConfiguration(proto))) {
+          LOG.info("{}: set new configuration {} from snapshot", getMemberId(), proto);
+          state.setRaftConf(proto);
+          state.writeRaftConfiguration(proto);
+          server.getStateMachine().event().notifyConfigurationChanged(
+              proto.getTerm(), proto.getIndex(), proto.getConfigurationEntry());
         }
       }
       return reply;
@@ -236,8 +236,8 @@ class SnapshotInstallationHandler {
 
         // For the cases where RaftConf is empty on newly started peer with empty peer list,
         // we retrieve leader info from installSnapShotRequestProto.
-        final RoleInfoProto roleInfoProto = leaderProto == null || server.getRaftConf().getPeer(state.
-            getLeaderId()) != null? server.getRoleInfoProto(): getRoleInfoProto(ProtoUtils.toRaftPeer(leaderProto));
+        final RoleInfoProto proto = leaderProto == null || server.getRaftConf().getPeer(state.getLeaderId()) != null?
+            server.getRoleInfoProto(): getRoleInfoProto(ProtoUtils.toRaftPeer(leaderProto));
         // This is the first installSnapshot notify request for this term and
         // index. Notify the state machine to install the snapshot.
         LOG.info("{}: notifyInstallSnapshot: nextIndex is {} but the leader's first available index is {}.",
@@ -253,8 +253,7 @@ class SnapshotInstallationHandler {
         // is updated when state.reloadStateMachine. We shall keep this index upgraded synchronously with main thread,
         // otherwise leader could get this follower's latest nextIndex from appendEntries instead of after
         // acknowledging the SNAPSHOT_INSTALLED.
-        server.getStateMachine().followerEvent()
-            .notifyInstallSnapshotFromLeader(roleInfoProto, firstAvailableLogTermIndex)
+        server.getStateMachine().followerEvent().notifyInstallSnapshotFromLeader(proto, firstAvailableLogTermIndex)
             .whenComplete((reply, exception) -> {
               if (exception != null) {
                 LOG.error("{}: Failed to notify StateMachine to InstallSnapshot. Exception: {}",
@@ -301,9 +300,9 @@ class SnapshotInstallationHandler {
 
       // If a snapshot has been installed, return SNAPSHOT_INSTALLED with the installed snapshot index and reset
       // installedSnapshotIndex to (0,0).
-      final TermIndex latestInstalledSnapshotTermIndex = this.installedSnapshotTermIndex.get();
+      final TermIndex latestInstalledSnapshotTermIndex = this.installedSnapshotTermIndex
+          .getAndSet(TermIndex.valueOf(0,0));
       if (latestInstalledSnapshotTermIndex.getIndex() > 0) {
-        this.installedSnapshotTermIndex.set(TermIndex.valueOf(0,0));
         server.getStateMachine().pause();
         state.updateInstalledSnapshotIndex(latestInstalledSnapshotTermIndex);
         state.reloadStateMachine(latestInstalledSnapshotTermIndex.getIndex());
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/SnapshotInstallationHandler.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/SnapshotInstallationHandler.java
@@ -33,6 +33,7 @@ import org.apache.ratis.protocol.RaftPeer;
 import org.apache.ratis.protocol.RaftPeerId;
 import org.apache.ratis.server.RaftServerConfigKeys;
 import org.apache.ratis.server.protocol.TermIndex;
+import org.apache.ratis.server.raftlog.LogProtoUtils;
 import org.apache.ratis.server.util.ServerStringUtils;
 import org.apache.ratis.util.CodeInjectionForTesting;
 import org.apache.ratis.util.LifeCycle;
@@ -44,10 +45,9 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 class SnapshotInstallationHandler {
   static final Logger LOG = LoggerFactory.getLogger(SnapshotInstallationHandler.class);
@@ -57,7 +57,7 @@ class SnapshotInstallationHandler {
 
   private final boolean installSnapshotEnabled;
   private final AtomicLong inProgressInstallSnapshotIndex = new AtomicLong();
-  private final AtomicLong installedSnapshotIndex = new AtomicLong();
+  private final AtomicReference<TermIndex> installedSnapshotTermIndex = new AtomicReference<>(TermIndex.valueOf(0,0));
   private final AtomicBoolean isSnapshotNull = new AtomicBoolean();
 
   SnapshotInstallationHandler(RaftServerImpl server, RaftProperties properties) {
@@ -119,13 +119,14 @@ class SnapshotInstallationHandler {
     if (reply != null) {
       if (request.hasLastRaftConfigurationLogEntryProto()) {
         // Set the configuration included in the snapshot
-        final LogEntryProto proto = request.getLastRaftConfigurationLogEntryProto();
-        LOG.info("{}: set new configuration {} from snapshot", getMemberId(), proto);
-
-        state.setRaftConf(proto);
-        state.writeRaftConfiguration(proto);
-        server.getStateMachine().event().notifyConfigurationChanged(
-            proto.getTerm(), proto.getIndex(), proto.getConfigurationEntry());
+        final LogEntryProto newConfLogEntryProto = request.getLastRaftConfigurationLogEntryProto();
+        if (!state.getRaftConf().equals(LogProtoUtils.toRaftConfiguration(newConfLogEntryProto))) {
+          LOG.info("{}: set new configuration {} from snapshot", getMemberId(), newConfLogEntryProto);
+          state.setRaftConf(newConfLogEntryProto);
+          state.writeRaftConfiguration(newConfLogEntryProto);
+          server.getStateMachine().event().notifyConfigurationChanged(newConfLogEntryProto.getTerm(),
+              newConfLogEntryProto.getIndex(), newConfLogEntryProto.getConfigurationEntry());
+        }
       }
       return reply;
     }
@@ -203,13 +204,13 @@ class SnapshotInstallationHandler {
       }
       server.changeToFollowerAndPersistMetadata(leaderTerm, "installSnapshot");
       state.setLeader(leaderId, "installSnapshot");
-      long snapshotIndex = state.getSnapshotIndex();
-
       server.updateLastRpcTime(FollowerState.UpdateType.INSTALL_SNAPSHOT_NOTIFICATION);
+
       if (inProgressInstallSnapshotIndex.compareAndSet(0, firstAvailableLogIndex)) {
         LOG.info("{}: Received notification to install snapshot at index {}", getMemberId(), firstAvailableLogIndex);
         // Check if snapshot index is already at par or ahead of the first
         // available log index of the Leader.
+        final long snapshotIndex = state.getLog().getSnapshotIndex();
         if (snapshotIndex + 1 >= firstAvailableLogIndex && firstAvailableLogIndex > 0) {
           // State Machine has already installed the snapshot. Return the
           // latest snapshot index to the Leader.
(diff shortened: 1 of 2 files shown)


Xushaohong:
The time for the user-defined state machine to install snapshot right now only leaves 1 sec. This might be too tight. Shall we make it configurable? @szetszwo

szetszwo:
> ... Shall we make it configurable?

Yes, it is a good idea to make it configurable.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1620,44 +1620,34 @@ private InstallSnapshotReplyProto notifyStateMachineToInstallSnapshot(
...
-          //nothing to do
-        } catch (Exception t) {
-          // there are two cases:
-          //1 `get()` may throw ExecutionException if `whenComplete` throw an exception
-          //2 when generating completeFuture, `statemachine#notifyInstallSnapshotFromLeader`
+          CompletableFuture<TermIndex> future = stateMachine.followerEvent()
+              .notifyInstallSnapshotFromLeader(roleInfoProto, firstAvailableLogTermIndex);
+          // wait for at most 5 seconds for statemachine to install snapshot
+          TermIndex reply = future.get(RaftServerConfigKeys.Notification.notifyInstallSnapshotTimeout(
+              getRaftServer().getProperties()).toLong(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
```
hanishakoneru, line 1627: We are essentially going to be blocking the main thread for 5 seconds (or more if configured that way). I don't think we should be blocking the main thread.
Also, inProgressInstallSnapshotRequest is not being set to 0 when the future completes.

hanishakoneru:
@Xushaohong, thanks for working on this.
Please correct me if I am wrong, but the state change calls mentioned above are still not being synchronized. 

> The following check in the main thread if (latestInstalledSnapshotIndex > 0) will be executed before state upgrade, which cause the raft server could not respond SNAPSHOT_INSTALLED in one request.

This is intentional so that we don't block the main thread while the SM is installing the snapshot. Instead we send a IN_PROGRESS notification back to the leader.

Xushaohong:
> @Xushaohong, thanks for working on this. Please correct me if I am wrong, but the state change calls mentioned above are still not being synchronized.
> 
> > The following check in the main thread if (latestInstalledSnapshotIndex > 0) will be executed before state upgrade, which cause the raft server could not respond SNAPSHOT_INSTALLED in one request.
> 
> This is intentional so that we don't block the main thread while the SM is installing the snapshot. Instead we send a IN_PROGRESS notification back to the leader.

@hanishakoneru 
I suppose the raw idea is to make the notifyStateMachineToInstallSnapshot process asynchronously, which means the notifications could be continuously arrived at the raft server to trigger the install of snapshot and the call should be idempotent before really installed. 

The problem is due to ```whenComplete``` stage. This part would be executed serially after the future done by SM with the same thread. This is where I mean synchronized. Since the install of the snapshot is asynchronous, the main thread will execute the following part first. ```      long latestInstalledSnapshotIndex = this.installedSnapshotIndex.getAndSet(0);
      if (latestInstalledSnapshotIndex > 0) ``` . This should be fine at the beginning, coz it will pass this check and return IN_PROGRESS. 
But once the SM has installed the snapshot, the main thread needs another call to go through the ```if (latestInstalledSnapshotIndex > 0)``` part as the previous call of notifyStateMachineToInstallSnapshot is done. There is a gap time between two calls. During this gap,  ```appendEntriesAsync``` might have seen the updated state and returned the ```state.getNextIndex()```. This causes the leader not to send the next notification anymore and the notification of install will not be finished.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
Xushaohong: The main thread blocking should be fine as it essentially can do nothing but wait for the snapshot done.

hanishakoneru:
> But once the SM has installed the snapshot, the main thread needs another call to go through the if (latestInstalledSnapshotIndex > 0) part as the previous call of notifyStateMachineToInstallSnapshot is done. There is a gap time between two calls. During this gap, appendEntriesAsync might have seen the updated state and returned the state.getNextIndex(). This causes the leader to misunderstand the follower's state, not to send the next notification anymore and the notification of install will not be finished.

Leader sends a snapshot (request or notification) to follower when either LogAppender#shouldInstallSnapshot() or GrpcLogAppender#shouldNotifyToInstallSnapshot() are true. And once these are true, without a state change acknowledgement from the follower, Leader will not send an appendEntries request.

Also, if the follower does not send a response back till it finishes installing the snapshot, the Leader will not be able to update the lastRpcResponseTime for this follower. This could lead to the leader stepping down thinking that it has lost the connection with this follower.

Xushaohong:
> Leader sends a snapshot (request or notification) to follower when either LogAppender#shouldInstallSnapshot() or GrpcLogAppender#shouldNotifyToInstallSnapshot() are true. And once these are true, without a state change acknowledgement from the follower, Leader will not send an appendEntries request.

Actually, Leader will send an appendEntries request. It  keeps send empty appendEntries requests to follower to keep in contact with follower as shouldNotifyToInstallSnapshot() is true as the below.
```
      // heartbeat
      return leaderState.newAppendEntriesRequestProto(follower, Collections.emptyList(), previous, callId);
    }
```
Hence The follower will send back INCONSISTENCY to the leader and the leader will updateNextIndex. 
```
        case INCONSISTENCY:
          grpcServerMetrics.onRequestInconsistency(getFollowerId().toString());
          updateNextIndex(reply.getNextIndex());
```



> 
> Also, if the follower does not send a response back till it finishes installing the snapshot, the Leader will not be able to update the lastRpcResponseTime for this follower. This could lead to the leader stepping down thinking that it has lost the connection with this follower.

This is also explained above. append log (heartbeat) will keep the follower in contact!

hanishakoneru:
@Xushaohong, please bear with me as I try to understand what is happening here and root causing the issue. 

> Actually, Leader will send an appendEntries request. It keeps send empty appendEntries requests to follower to keep in contact with follower as shouldNotifyToInstallSnapshot() is true as the below.
Hence The follower will send back INCONSISTENCY to the leader and the leader will updateNextIndex.

What is the follower sends SNAPSHOT_INSTALLATION_IN_PROGRESS status instead of INCONSISTENCY status to the appendEntries request? IIUC, the issue is occurring because of this inconsistency reply to appendEntries request from Leader.

Xushaohong:
> What is the follower sends SNAPSHOT_INSTALLATION_IN_PROGRESS status instead of INCONSISTENCY status to the appendEntries request? 

Sry, I do not get this since I have not modified the follower to send back this status.

> IIUC, the issue is occurring because of this inconsistency reply to appendEntries request from Leader.

Yep, this is nearly the root cause. 

Precondition:
There are two communicating threads between the leader and follower. One is to notify install snapshot, and the other is appendEntries thread(sending heartbeat and entries).
The statemachine installing snapshot(lets say thread A) and return progress of installing snapshot(thread B) are asynchronous. The update of index(commit index and next index) is executed serially by the thread A. 
In practical, the statemachine installing snapshot takes increasing time to finish. The main thread (lets say thread B) will response `IN_PROGRESS` during the install, but once the install is done, follower needs new request to return  `SNAPSHOT_INSTALLED`, since last request should have already replied IN_PROGRESS.


If we keep the non-blocking style to let follower response to the leader with its instant status, there are two problems.

Once snapshot is done, the updated index could be reached first by appendentries thread(lets say thread C), and C will return the updated nextIndex to the leader. The leader hence will misunderstand the status of the follower and not send the notification. The follower needs the new notification to send back SNAPSHOT_INSTALLED. The consequence is that thread B will get stuck, and thread C will wait for B done and always return INCONSISTENCY to the leader.
	
My solution: let update of index be executed by the main thread B, so the follower can ensure to send out the SNAPSHOT_INSTALLED after update of index.

Currently, appendentries thread C return the `state.getNextIndex()` when snapshot installation in progress.  The nextIndex is calculated by Math.max(logNextIndex, snapshotNextIndex). Here snapshotNextIndex is got through server.getStateMachine().getLatestSnapshot(). We should return the updated index by thread B instead of statemachine’s.
My solution: replace `snapshotNextIndex = getSnapshotIndex() + 1;` with `snapshotNextIndex = log.getSnapshotIndex() + 1;`

Xushaohong:
install_snapshot drawio

hanishakoneru:
Thanks @Xushaohong for the detailed explanation. Appreciate it.

I understand the problem now and your change makes sense to me. The overall logic LGTM. I will take closer look at it tomorrow. Can you please check the CI failures if they are related. Thanks.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1561,7 +1561,8 @@ private InstallSnapshotReplyProto installSnapshotImpl(InstallSnapshotRequestProt
     }
 
     if (reply != null) {
-      if (request.hasLastRaftConfigurationLogEntryProto()) {
+      if (request.hasLastRaftConfigurationLogEntryProto() && reply.getResult() ==
+              InstallSnapshotResult.SNAPSHOT_INSTALLED) {
```
hanishakoneru, line 1565: SNAPSHOT_INSTALLED result is returned when SM is notified to install snapshot. For the case when snapshot is sent as part of installSnapshot request by Ratis,  SUCCESS is returned as the result.

hanishakoneru (changes requested):
Patch LGTM overall. 

AppendEntries should not return INCONSISTENCY result when a snapshot is being installed. It would be good to add another result type to AppendEntries request. We can open another Jira to address that.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
Xushaohong: Oops, I missed that. Now I added the difference check to reduce the frequency of resetting conf. Coz this is redundant to overwrite the conf for each request.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1561,18 +1561,16 @@ private InstallSnapshotReplyProto installSnapshotImpl(InstallSnapshotRequestProt
...
-            request.getLastRaftConfigurationLogEntryProto();
-        LOG.info("{}: set new configuration {} from snapshot", getMemberId(),
-            newConfLogEntryProto);
-
-        state.setRaftConf(newConfLogEntryProto);
-        state.writeRaftConfiguration(newConfLogEntryProto);
-        stateMachine.event().notifyConfigurationChanged(newConfLogEntryProto.getTerm(), newConfLogEntryProto.getIndex(),
-            newConfLogEntryProto.getConfigurationEntry());
+        LogEntryProto newConfLogEntryProto = request.getLastRaftConfigurationLogEntryProto();
+        if (!state.getRaftConf().equals(LogProtoUtils.toRaftConfiguration(newConfLogEntryProto))) {
```
hanishakoneru, line 1567: If we do not call _state.writeRaftConfiguration(newConfLogEntryProto)_ here, a bootstrapping follower might not write it's configuration to disk at all till the StateMachine applies a log containing a configuration entry. 

_ServerState#writeRaftConfiguration()_ is called only from 2 places - installSnapshot and applyLog. If the conf is persisted on disk some other way also, then we are good. If not, we should check that condition also here before deciding to not update the conf entry on disk.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1651,7 +1651,7 @@ private InstallSnapshotReplyProto notifyStateMachineToInstallSnapshot(
       }
       changeToFollowerAndPersistMetadata(leaderTerm, "installSnapshot");
       state.setLeader(leaderId, "installSnapshot");
-      long snapshotIndex = state.getSnapshotIndex();
+      long snapshotIndex = state.getLog().getSnapshotIndex();
```
hanishakoneru, line 1652: Why is this change required? ServerState snapshotIndex should be the same as RaftLog snapshotIndex. 
Please let me know if I am missing something here.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
Xushaohong: Here the conf entry on disk is only loaded when the server is restarted, as if the server is the first time to start, it has no persisted conf.

    // read configuration from the storage
    Optional.ofNullable(storage.readRaftConfiguration()).ifPresent(this::setRaftConf);
Thus there are two cases:

The init conf is as same as the conf which is about to update. This won't affect at all, as it will still write conf when actually applied a log.
The init conf is different from the conf which is about to update. Here the equivalence check will ensure it will do the write.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
Xushaohong: There is the possibility that after statemachine just installed the snapshot and before the main thread return the  SNAPSHOT_INSTALLED, the snapshotIndex here will get from the statemachine instead of memory and  return ALREADY_INSTALLED, which will cause the follower not to update the index and reload statemachine

hanishakoneru (approved):
Thanks @Xushaohong for working on this and fixing it.

JacksonYao287 (commented):
thanks @Xushaohong for the work!
in the background thread , which is started by `stateMachine.followerEvent().notifyInstallSnapshotFromLeader.whenComplete()`, `inProgressInstallSnapshotRequest` will be set back to `0` if snapshot is installed successfully.

there may be a case that `inProgressInstallSnapshotRequest` is just set back to `0` in the background thread, a `notifyInstallSnapshot` request from leader comes and  `stateMachine.followerEvent().notifyInstallSnapshotFromLeader.whenComplete()` will be called again. is this a problem? 
i think we should add some check to see if the`firstAvailableLogTermIndex` included in the request is larger than the lastTermIndex in the current follower`s log
Please let me know if I am missing something here.

Xushaohong:
> thanks @Xushaohong for the work! in the background thread , which is started by `stateMachine.followerEvent().notifyInstallSnapshotFromLeader.whenComplete()`, `inProgressInstallSnapshotRequest` will be set back to `0` if snapshot is installed successfully.
> 
> there may be a case that `inProgressInstallSnapshotRequest` is just set back to `0` in the background thread, a `notifyInstallSnapshot` request from leader comes and `stateMachine.followerEvent().notifyInstallSnapshotFromLeader.whenComplete()` will be called again. is this a problem? i think we should add some check to see if the`firstAvailableLogTermIndex` included in the request is larger than the lastTermIndex in the current follower`s log Please let me know if I am missing something here.

Only when an exception happens will the ```inProgressInstallSnapshotRequest``` set be 0 immediately, otherwise it will set ```installedSnapshotTermIndex``` value as what successful snapshot index is and wait for another notify install snapshot request comeing, and when the notify main thread finds that this.installedSnapshotTermIndex.get() is not 0, then it unsets ```inProgressInstallSnapshotRequest``` to 0 to ensure the safety.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1687,44 +1686,38 @@ private InstallSnapshotReplyProto notifyStateMachineToInstallSnapshot(
...
+        // state.reloadStateMachine. We shall keep this index upgrade synchronized with main thread, or else leader
+        // could get this follower's latest nextIndex from appendEntries, it will not send any notify install snapshot
+        // request, not receive SNAPSHOT_INSTALLED answer and then get stuck in the loop of sending appendEntries and
+        // receiving inconsistency entries.
+        stateMachine.followerEvent().notifyInstallSnapshotFromLeader(roleInfoProto, firstAvailableLogTermIndex)
+            .whenComplete((reply, exception) -> {
+              if (exception != null) {
+                LOG.error("{}: Failed to notify StateMachine to InstallSnapshot. Exception: {}",
+                    getMemberId(), exception.getMessage());
+                inProgressInstallSnapshotRequest.compareAndSet(firstAvailableLogIndex, 0);
```
JacksonYao287, line 1706: seems it is better to use `inProgressInstallSnapshotRequest.set(0)` if an exception is thrown, no need to compare!

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1650,7 +1649,7 @@ private InstallSnapshotReplyProto notifyStateMachineToInstallSnapshot(
       }
       changeToFollowerAndPersistMetadata(leaderTerm, "installSnapshot");
       state.setLeader(leaderId, "installSnapshot");
-      long snapshotIndex = state.getSnapshotIndex();
+      long snapshotIndex = state.getLog().getSnapshotIndex();
```
JacksonYao287, line 1652: maybe we can put this line before `if (snapshotIndex + 1 >= firstAvailableLogIndex && firstAvailableLogIndex > 0)`

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
JacksonYao287: > There is the possibility that after statemachine just installed the snapshot and before the main thread return the SNAPSHOT_INSTALLED, the snapshotIndex here will get from the statemachine instead of memory and return ALREADY_INSTALLED, which will cause the follower not to update the index and reload statemachine

before the main thread return the SNAPSHOT_INSTALLED, `inProgressInstallSnapshotRequest` will not be set back to 0 if no exception is thrown. so the following code will not be executed until `inProgressInstallSnapshotRequest` is set back to 0. 
```
        if (snapshotIndex + 1 >= firstAvailableLogIndex && firstAvailableLogIndex > 0) {
          // State Machine has already installed the snapshot. Return the
          // latest snapshot index to the Leader.

          inProgressInstallSnapshotRequest.compareAndSet(firstAvailableLogIndex, 0);
          LOG.info("{}: InstallSnapshot notification result: {}, current snapshot index: {}", getMemberId(),
              InstallSnapshotResult.ALREADY_INSTALLED, snapshotIndex);
          return ServerProtoUtils.toInstallSnapshotReplyProto(leaderId, getMemberId(), currentTerm,
              InstallSnapshotResult.ALREADY_INSTALLED, snapshotIndex);
        }
```
if `inProgressInstallSnapshotRequest` is set back to 0, it means the snapshot has been installed successfully,  so now ServerState snapshotIndex should be the same as RaftLog snapshotIndex because the statemachine has been reloaded. 
so seems the changes here is unnecessary , please correct me if i miss something here!

JacksonYao287 (commented):
thanks @Xushaohong for the explanation. the changes overall looks good to me. i left several comments , please take a look.
when the follower is installing snapshot , leader will send a lot of notifyInstallSnapshot requests repeatedly, which is unnecessary。so can you please create a jira to solve this problem? or handle this issue in this patch?

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
Xushaohong: Since it is protected by the descriptor ```synchronized```, it should be fine here

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
Xushaohong: It should not have the case when an exception happens, the inProgressInstallSnapshotRequest is not the value of ```firstAvailableLogIndex```. On the other hand, this is somehow protection that if the corner case happens, there  
 must be something wrong, and this will leave the follower keeping the state of IN_PROGRESS and alert the maintainer.

Xushaohong:
> thanks @Xushaohong for the explanation. the changes overall looks good to me. i left several comments , please take a look. when the follower is installing snapshot , leader will send a lot of notifyInstallSnapshot requests repeatedly, which is unnecessary。so can you please create a jira to solve this problem? or handle this issue in this patch?

Yep, i have a patch in my local env to fix it and this PR is the prerequisite. I would create a new PR ASAP.

JacksonYao287 (approved):
LGTM. thanks @Xushaohong for this work! 
@szetszwo please take a look

szetszwo:
@Xushaohong , sorry that there are some conflicts.  Could you update it?

szetszwo (commented):
@Xushaohong , thanks for the update.  Some minor comments inlined.

ratis-server/src/main/java/org/apache/ratis/server/impl/SnapshotInstallationHandler.java
```diff
@@ -289,24 +284,34 @@ private InstallSnapshotReplyProto notifyStateMachineToInstallSnapshot(
...
         return ServerProtoUtils.toInstallSnapshotReplyProto(leaderId, getMemberId(),
             currentTerm, InstallSnapshotResult.SNAPSHOT_UNAVAILABLE, -1);
       }
 
       // If a snapshot has been installed, return SNAPSHOT_INSTALLED with the installed snapshot index and reset
-      // installedSnapshotIndex to 0.
-      long latestInstalledSnapshotIndex = this.installedSnapshotIndex.getAndSet(0);
-      if (latestInstalledSnapshotIndex > 0) {
+      // installedSnapshotIndex to (0,0).
+      final TermIndex latestInstalledSnapshotTermIndex = this.installedSnapshotTermIndex.get();
```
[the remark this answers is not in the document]
Xushaohong: Here I have a question, why the operation is not atomic here?
1. The process is protected by the ```synchronized`` lock.
2. With this lock protection, here the processing logic is idempotent. The second request would not affect the previous one's result.

szetszwo (approved):
+1 the change looks good.

PR #620, opened 2022-03-09
RATIS-1549. If the stream client is abnormally disconnected, reconnec…

guohao-rosicky:

If the stream client is abnormally disconnected, reconnect the stream client.

Such as:
1. Socket channle is close
2. The network connection is down

https://issues.apache.org/jira/browse/RATIS-1549

### DIFF
--- a/ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
+++ b/ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
@@ -29,23 +29,34 @@ import org.apache.ratis.protocol.ClientInvocationId;
 import org.apache.ratis.protocol.DataStreamReply;
 import org.apache.ratis.protocol.DataStreamRequest;
 import org.apache.ratis.protocol.RaftPeer;
+import org.apache.ratis.protocol.exceptions.AlreadyClosedException;
 import org.apache.ratis.security.TlsConf;
 import org.apache.ratis.thirdparty.io.netty.bootstrap.Bootstrap;
 import org.apache.ratis.thirdparty.io.netty.buffer.ByteBuf;
-import org.apache.ratis.thirdparty.io.netty.channel.*;
-import org.apache.ratis.thirdparty.io.netty.channel.nio.NioEventLoopGroup;
+import org.apache.ratis.thirdparty.io.netty.channel.Channel;
+import org.apache.ratis.thirdparty.io.netty.channel.ChannelFuture;
+import org.apache.ratis.thirdparty.io.netty.channel.ChannelFutureListener;
+import org.apache.ratis.thirdparty.io.netty.channel.ChannelHandlerContext;
+import org.apache.ratis.thirdparty.io.netty.channel.ChannelInboundHandler;
+import org.apache.ratis.thirdparty.io.netty.channel.ChannelInboundHandlerAdapter;
+import org.apache.ratis.thirdparty.io.netty.channel.ChannelInitializer;
+import org.apache.ratis.thirdparty.io.netty.channel.ChannelOption;
+import org.apache.ratis.thirdparty.io.netty.channel.ChannelPipeline;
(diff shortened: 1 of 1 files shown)


guohao-rosicky:
What do you think of this feature? @szetszwo

guohao-rosicky:
1. Ctx.writeAndFlush() actually puts the request into the socket buffer. It just returns a Future object, and we need to add a listener to get its status
2. If channel inactive, we will fail to send with the current channel, so we need to trigger reconnection
3. When we connect, we add a listener to get the state of the connection

szetszwo:
@guohao-rosicky , please add some tests.  We cannot easily tell if the change is working.

guohao-rosicky:
Okay, I'll make a UT @szetszwo

guohao-rosicky:
@captainzmc  can you  talk about the test results of this patch?

captainzmc:
Thanks @guohao-rosicky  update this. I did a comparative test and found that this patch can indeed optimize the problem of client hung.  While testing three clients to write 400 files concurrently, ratis master hung every time, but RATIS-1549 runs normally.
image

Hi @szetszwo, In UT it is difficult to simulate a broken connection during write. Can we continue to discuss this patch based on the above test results?

szetszwo:
@captainzmc , @guohao-rosicky , thanks for testing it manually.  I will review this change first.  Let me think about how to add a unit test.

szetszwo:
@guohao-rosicky , I suggest to refactor the connection related code to a class; see https://issues.apache.org/jira/secure/attachment/13041023/620_review.patch

guohao-rosicky:
@szetszwo  I have applied your patch

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
```diff
@@ -125,30 +139,109 @@ int size() {
...
+      getWorkerGroup().schedule(this::reconnect, RECONNECT.getDuration(), RECONNECT.getUnit());
+    }
+
+    private ChannelFuture reconnect() {
+      final MemoizedSupplier<ChannelFuture> supplier = MemoizedSupplier.valueOf(this::connect);
+      final ChannelFuture previous = ref.getAndUpdate(prev -> prev == null? null: supplier.get());
+      if (previous != null) {
+        previous.channel().close();
+      }
+      return supplier.get();
```
szetszwo, line 209: We should check supplier.isInitialized().
```
      return supplier.isInitialized()? supplier.get(): null;
```

szetszwo (commented):
@guohao-rosicky , thanks for the update.  Just two minor comments inlined.

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
```diff
@@ -261,15 +362,27 @@ protected void decode(ChannelHandlerContext context, ByteBuf buf, List<Object> o
...
+      f.completeExceptionally(new AlreadyClosedException(this + ": Failed to send " + request));
+      return f;
+    }
     LOG.debug("{}: write {}", this, request);
-    getChannel().writeAndFlush(request);
+    channel.writeAndFlush(request).addListener(future -> {
+      if (!future.isSuccess()) {
+        final IOException e = new IOException(this + ": Failed to send " + request, future.cause());
+        LOG.error("Channel write failed", e);
+        if (!f.isDone()) {
```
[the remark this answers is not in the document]
guohao-rosicky: I see. Thank you

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
guohao-rosicky: Has been modified. @szetszwo Please take a look.

szetszwo (approved):
+1 the change looks good.

PR #636, opened 2022-04-24
RATIS-1571. Add ratis-shell snapshot command unit test

codings-dan:
see https://issues.apache.org/jira/browse/RATIS-1571


codings-dan:
@szetszwo Can you help to review this patch, thanks!

szetszwo (approved):
+1 the change looks good.

PR #645, opened 2022-05-12
RATIS-1580. Add ratis-shell peer and election command unit test

codings-dan:
see https://issues.apache.org/jira/browse/RATIS-1580


szetszwo (approved):
+1 the change looks good.

codings-dan:
Thanks @szetszwo  for reviewing the code!

PR #652, opened 2022-06-01
RATIS-1588. Add a builder of RaftStorageImpl

codings-dan:
see https://issues.apache.org/jira/projects/RATIS/issues/RATIS-1588

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
@@ -42,11 +42,6 @@ public class RaftStorageImpl implements RaftStorage {
   private final CorruptionPolicy logCorruptionPolicy;
   private volatile RaftStorageMetadataFileImpl metaFile;
 
-  public RaftStorageImpl(File dir, CorruptionPolicy logCorruptionPolicy,
-      long storageFeeSpaceMin) throws IOException {
-    this(dir, logCorruptionPolicy, null, storageFeeSpaceMin);
-  }
-
   RaftStorageImpl(File dir, CorruptionPolicy logCorruptionPolicy, StartupOption option,
       long storageFeeSpaceMin) throws IOException {
     this.storageDir = new RaftStorageDirectoryImpl(dir, storageFeeSpaceMin);
--- a/ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
@@ -36,46 +36,6 @@ import java.util.Optional;
 /** The storage of a {@link org.apache.ratis.server.RaftServer}. */
 public class RaftStorageImpl implements RaftStorage {
 
-  public enum StartupOption {
-    /** Format the storage. */
-    FORMAT
-  }
-
-  public static Builder newBuilder() {
-    return new RaftStorageImpl.Builder();
-  }
-
-  public static class Builder {
-    private File dir;
-    private CorruptionPolicy logCorruptionPolicy;
-    private StartupOption option;
-    private long storageFeeSpaceMin;
-
-    public Builder setDir(File dir) {
-      this.dir = dir;
-      return this;
-    }
-
-    public Builder setLogCorruptionPolicy(CorruptionPolicy logCorruptionPolicy) {
-      this.logCorruptionPolicy = logCorruptionPolicy;
-      return this;
-    }
-
-    public Builder setOption(StartupOption option) {
-      this.option = option;
-      return this;
-    }
-
-    public Builder setStorageFeeSpaceMin(long storageFeeSpaceMin) {
-      this.storageFeeSpaceMin = storageFeeSpaceMin;
-      return this;
-    }
-
-    public RaftStorageImpl build() throws IOException {
-      return new RaftStorageImpl(dir, logCorruptionPolicy, option, storageFeeSpaceMin);
-    }
-  }
-
   // TODO support multiple storage directories
   private final RaftStorageDirectoryImpl storageDir;
   private final StorageState state;
--- a/ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
@@ -19,6 +19,7 @@ package org.apache.ratis.server.storage;
 
 import org.apache.ratis.proto.RaftProtos.LogEntryProto;
 import org.apache.ratis.server.RaftConfiguration;
+import org.apache.ratis.server.RaftServerConfigKeys;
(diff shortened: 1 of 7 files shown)


szetszwo:
@codings-dan , sorry that it broke Alluxio.  RaftStorageImpl is not a public API so that we do not have to maintain compatibility.  We may consider adding a RaftStorage.Builder as a public API.

BTW, this is backward compatibility -- a new version of the software works for the existing applications.  Forward compatibility means the current design of the software can work on future applications.

codings-dan:
> We may consider adding a RaftStorage.Builder as a public API.

Good idea, I will add a RaftStorage.Builder.

> a new version of the software works for the existing applications. Forward compatibility means the current design of the software can work on future applications

Thanks for telling me about these, I now understand the difference!

codings-dan:
@szetszwo I have added a RaftStorage.Builder, PTAL again, thx!

ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
```diff
@@ -41,6 +41,41 @@ public enum StartupOption {
     FORMAT
   }
 
+  public static Builder newBuilder() {
+    return new RaftStorageImpl.Builder();
+  }
+
+  public static class Builder {
```
szetszwo, line 48: The Builder should be moved to RaftStorage so that it will be in raft-server-api.

ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
```diff
@@ -41,6 +41,41 @@ public enum StartupOption {
...
+      this.dir = dir;
+      return this;
+    }
+
+    public Builder setLogCorruptionPolicy(CorruptionPolicy logCorruptionPolicy) {
+      this.logCorruptionPolicy = logCorruptionPolicy;
+      return this;
+    }
+
+    public Builder setOption(StartupOption option) {
```
szetszwo, line 64: We should move StartupOption from RaftStorageImpl to RaftStorage so that it becomes a public API.

Also, let's add a new StartupOption.RECOVER instead of using null.

ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
```diff
@@ -41,6 +41,41 @@ public enum StartupOption {
...
+      this.option = option;
+      return this;
+    }
+
+    public Builder setStorageFeeSpaceMin(long storageFeeSpaceMin) {
```
szetszwo, line 69: Let's call it setFreeSpaceMin.   BTW, there is a TYPO: "Fee" should "Free".

Also, let's use SizeInBytes for the parameter.

szetszwo (commented):
@codings-dan , thanks for the update!  Some comments inlined.

ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
codings-dan: done

ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
```diff
@@ -41,6 +41,41 @@ public enum StartupOption {
...
+      this.storageFeeSpaceMin = storageFeeSpaceMin;
+      return this;
+    }
+
+    public RaftStorageImpl build() throws IOException {
+      return new RaftStorageImpl(dir, logCorruptionPolicy, option, storageFeeSpaceMin);
```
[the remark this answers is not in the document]
codings-dan: done

ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
```diff
@@ -41,6 +41,41 @@ public enum StartupOption {
...
+    return new RaftStorageImpl.Builder();
+  }
+
+  public static class Builder {
+    private File dir;
+    private CorruptionPolicy logCorruptionPolicy;
+    private StartupOption option;
+    private long storageFeeSpaceMin;
+
+    public Builder setDir(File dir) {
```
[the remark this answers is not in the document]
codings-dan: done

ratis-server/src/main/java/org/apache/ratis/server/storage/RaftStorageImpl.java
codings-dan: done

codings-dan:
@szetszwo PTAL again, thank you!

szetszwo (approved):
+1 the change looks good.

szetszwo:
@codings-dan , The test failure seems related.  Could you take a look?

BTW, let's remove the public RaftStorageImpl constructor below since it is no longer used.
```
  public RaftStorageImpl(File dir, CorruptionPolicy logCorruptionPolicy,
      long storageFeeSpaceMin) throws IOException {
```

Also, let's use SizeInBytes for storageFreeSpaceMin? Or, we may have two setStorageFreeSpaceMin methods.
```
    public Builder setStorageFreeSpaceMin(SizeInBytes storageFreeSpaceMin) {
      this.storageFreeSpaceMin = storageFreeSpaceMin.getSize();
      return this;
    }

    public Builder setStorageFreeSpaceMin(long storageFreeSpaceMin) {
      this.storageFreeSpaceMin = storageFreeSpaceMin;
      return this;
    }
```

codings-dan:
> The test failure seems related. Could you take a look

This is caused by improper handling of exceptions in reflection，I have fixed it. 

> let's remove the public RaftStorageImpl constructor below since it is no longer used.

I have remove the extra code.

> let's use SizeInBytes for storageFreeSpaceMin

Done.

codings-dan:
@szetszwo Thanks for help review the code, PTAL, thx!

szetszwo:
@codings-dan , thanks for the update.  TestRaftStorage failed.  Please take a look.
https://github.com/apache/ratis/runs/6920545804?check_suite_focus=true#step:5:562

codings-dan:
> TestRaftStorage failed.
Similar to the previous ut error, it was caused by improper handling of exceptions in reflection, I have fixed it

codings-dan:
The ci test error seems have noting to do with this code change.

https://github.com/apache/ratis/runs/6930404226?check_suite_focus=true#step:5:611

szetszwo:
Sure, let's restart the failed jobs.

codings-dan:
https://github.com/apache/ratis/runs/6931402074?check_suite_focus=true#step:5:611

This unit test can run locally, see as below
image

szetszwo:
#635 should be able to reduce the timeout problem.  @codings-dan , could you review it?

codings-dan:
> #635 should be able to reduce the timeout problem. @codings-dan , could you review it?

Of course, happy to help review the code!

PR #671, opened 2022-07-07
RATIS-1601. Use the shaded dropwizard metrics and remove the dependency

codings-dan:
see https://issues.apache.org/jira/browse/RATIS-1601


codings-dan:
@szetszwo Could you help review this patch, thank you in advance!

szetszwo (approved):
+1 the change looks good.

PR #673, opened 2022-07-07
RATIS-1612. Support starting a server as a Listener.

qian0817:

Support starting a server as a Listener.

https://issues.apache.org/jira/browse/RATIS-1612

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/protocol/RaftGroup.java
+++ b/ratis-common/src/main/java/org/apache/ratis/protocol/RaftGroup.java
@@ -17,9 +17,11 @@
  */
 package org.apache.ratis.protocol;
 
+import org.apache.ratis.proto.RaftProtos;
 import org.apache.ratis.util.Preconditions;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * Description of a raft group, which has a unique {@link RaftGroupId} and a collection of {@link RaftPeer}.
@@ -43,28 +45,17 @@ public final class RaftGroup {
     return new RaftGroup(groupId, peers);
   }
 
-  public static RaftGroup valueOf(RaftGroupId groupId, Collection<RaftPeer> peers,
-                                  Collection<RaftPeer> listeners) {
-    return new RaftGroup(groupId, peers, listeners);
-  }
-
   /** The group id */
   private final RaftGroupId groupId;
   /** The group of raft peers */
   private final Map<RaftPeerId, RaftPeer> peers;
-  private final Map<RaftPeerId, RaftPeer> listeners;
 
   private RaftGroup() {
     this.groupId = RaftGroupId.emptyGroupId();
     this.peers = Collections.emptyMap();
-    this.listeners = Collections.emptyMap();
   }
 
   private RaftGroup(RaftGroupId groupId, Collection<RaftPeer> peers) {
-    this(groupId, peers, Collections.emptyList());
-  }
-
-  private RaftGroup(RaftGroupId groupId, Collection<RaftPeer> peers,Collection<RaftPeer> listeners){
     this.groupId = Objects.requireNonNull(groupId, "groupId == null");
     Preconditions.assertTrue(!groupId.equals(EMPTY_GROUP.getGroupId()),
         () -> "Group Id " + groupId + " is reserved for the empty group.");
@@ -76,13 +67,6 @@ public final class RaftGroup {
       peers.forEach(p -> map.put(p.getId(), p));
       this.peers = Collections.unmodifiableMap(map);
     }
-    if (listeners == null || listeners.isEmpty()) {
-      this.listeners = Collections.emptyMap();
-    } else {
-      final Map<RaftPeerId, RaftPeer> map = new HashMap<>();
-      listeners.forEach(p -> map.put(p.getId(), p));
-      this.listeners = Collections.unmodifiableMap(map);
-    }
   }
 
   public RaftGroupId getGroupId() {
@@ -93,8 +77,11 @@ public final class RaftGroup {
     return peers.values();
   }
 
-  public Collection<RaftPeer> getListeners() {
-    return listeners.values();
+  public Collection<RaftPeer> getPeers(RaftProtos.RaftPeerRole role) {
+    return peers.values()
+        .stream()
+        .filter(peer -> peer.getStartupRole() == role)
+        .collect(Collectors.toList());
   }
 
   /** @return the peer with the given id if it is in this group; otherwise, return null. */
--- a/ratis-common/src/main/java/org/apache/ratis/protocol/RaftGroup.java
+++ b/ratis-common/src/main/java/org/apache/ratis/protocol/RaftGroup.java
@@ -17,11 +17,9 @@
  */
 package org.apache.ratis.protocol;
 
-import org.apache.ratis.proto.RaftProtos;
 import org.apache.ratis.util.Preconditions;
 
 import java.util.*;
-import java.util.stream.Collectors;
 
 /**
  * Description of a raft group, which has a unique {@link RaftGroupId} and a collection of {@link RaftPeer}.
@@ -41,7 +39,7 @@ public final class RaftGroup {
   }
 
   /** @return a group with the given id and peers. */
-  public static RaftGroup valueOf(RaftGroupId groupId, Collection<RaftPeer> peers) {
+  public static RaftGroup valueOf(RaftGroupId groupId, Iterable<RaftPeer> peers) {
     return new RaftGroup(groupId, peers);
   }
 
@@ -55,12 +53,12 @@ public final class RaftGroup {
     this.peers = Collections.emptyMap();
   }
 
-  private RaftGroup(RaftGroupId groupId, Collection<RaftPeer> peers) {
+  private RaftGroup(RaftGroupId groupId, Iterable<RaftPeer> peers) {
     this.groupId = Objects.requireNonNull(groupId, "groupId == null");
     Preconditions.assertTrue(!groupId.equals(EMPTY_GROUP.getGroupId()),
         () -> "Group Id " + groupId + " is reserved for the empty group.");
 
-    if (peers == null || peers.isEmpty()) {
+    if (peers == null || !peers.iterator().hasNext()) {
       this.peers = Collections.emptyMap();
     } else {
       final Map<RaftPeerId, RaftPeer> map = new HashMap<>();
--- a/ratis-common/src/main/java/org/apache/ratis/protocol/RaftGroup.java
+++ b/ratis-common/src/main/java/org/apache/ratis/protocol/RaftGroup.java
@@ -43,17 +43,28 @@ public final class RaftGroup {
     return new RaftGroup(groupId, peers);
   }
 
+  public static RaftGroup valueOf(RaftGroupId groupId, Collection<RaftPeer> peers,
+                                  Collection<RaftPeer> listeners) {
+    return new RaftGroup(groupId, peers, listeners);
+  }
+
   /** The group id */
   private final RaftGroupId groupId;
   /** The group of raft peers */
   private final Map<RaftPeerId, RaftPeer> peers;
+  private final Map<RaftPeerId, RaftPeer> listeners;
 
   private RaftGroup() {
     this.groupId = RaftGroupId.emptyGroupId();
     this.peers = Collections.emptyMap();
+    this.listeners = Collections.emptyMap();
   }
 
   private RaftGroup(RaftGroupId groupId, Collection<RaftPeer> peers) {
+    this(groupId, peers, Collections.emptyList());
+  }
+
+  private RaftGroup(RaftGroupId groupId, Collection<RaftPeer> peers,Collection<RaftPeer> listeners){
     this.groupId = Objects.requireNonNull(groupId, "groupId == null");
     Preconditions.assertTrue(!groupId.equals(EMPTY_GROUP.getGroupId()),
         () -> "Group Id " + groupId + " is reserved for the empty group.");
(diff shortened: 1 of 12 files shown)


qian0817:
@codings-dan I noticed that the Jira issue was assigned to you,can you take a look?

qian0817:
> > @codings-dan I noticed that the Jira issue was assigned to you,can you take a look?
> 
> Sure , happy to help review this pull request. BTW, could you left a comment in the https://issues.apache.org/jira/browse/RATIS-1612 , and I will re-assign this jira to you, Thanks.

done

codings-dan (changes requested):
@qian0817 Thanks for working on this, just as @szetszwo said, since update the raftgroup state is not easy and may introduce many problems, we should try to avoid changing the RaftGroup. In addition, you may also need to change the code related to RaftServer.builder

qian0817:
@codings-dan @szetszwo Can you take a look at the new implementation?

ratis-common/src/main/java/org/apache/ratis/protocol/RaftGroup.java
```diff
@@ -75,6 +77,13 @@ public Collection<RaftPeer> getPeers() {
     return peers.values();
   }
 
+  public Collection<RaftPeer> getPeers(RaftProtos.RaftPeerRole role) {
+    return peers.values()
+        .stream()
+        .filter(peer -> peer.getStartupRole() == role)
```
codings-dan, line 83: I think we can't use `startupRole` as the difference between raftpeer, because the peer will change its identity

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
```diff
@@ -75,6 +77,7 @@ public static class Builder {
     private String clientAddress;
     private String dataStreamAddress;
     private int priority;
+    private RaftProtos.RaftPeerRole startupRole;
```
codings-dan, line 80: set `startupRole` defalut value `RaftProtos.RaftPeerRole.FOLLOWER`

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
```diff
@@ -133,10 +136,24 @@ public Builder setPriority(int priority) {
...
+        throw new IllegalArgumentException(
+            "At startup the role can only be set to FOLLOWER or LISTENER, the current value is " +
+                startupRole);
+      }
+      this.startupRole = startupRole;
+      return this;
+    }
+
     public RaftPeer build() {
+      if (startupRole == null) {
```
codings-dan, line 151: Please remove the if statement

ratis-common/src/main/java/org/apache/ratis/util/ProtoUtils.java
```diff
@@ -118,6 +119,7 @@ static RaftPeer toRaftPeer(RaftPeerProto p) {
         .setClientAddress(p.getClientAddress())
         .setAdminAddress(p.getAdminAddress())
         .setPriority(p.getPriority())
+        .setStartupRole(p.hasStartupRole() ? p.getStartupRole() : RaftProtos.RaftPeerRole.FOLLOWER)
```
codings-dan, line 122: Since the default value is set above, redundant judgment logic can be deleted here

ratis-proto/src/main/proto/Raft.proto
```diff
@@ -28,6 +28,7 @@ message RaftPeerProto {
   string dataStreamAddress = 4; // address of the data stream server
   string clientAddress = 5; // address of the client RPC server
   string adminAddress = 6; // address of the admin RPC server
+  optional RaftPeerRole startupRole=7; // peer start up role
```
codings-dan, line 31: typo. `optional RaftPeerRole startupRole = 7`

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
```diff
@@ -133,10 +136,24 @@ public Builder setPriority(int priority) {
       return this;
     }
 
+    public Builder setStartupRole(RaftProtos.RaftPeerRole startupRole) {
+      if (startupRole != RaftProtos.RaftPeerRole.FOLLOWER
```
codings-dan, line 140: I think `!=` can be replaced by `!...equals` here

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -951,9 +951,11 @@ private void yieldLeaderToHigherPriorityPeer() {
     for (LogAppender logAppender : senders.getSenders()) {
       final FollowerInfo followerInfo = logAppender.getFollower();
       final RaftPeerId followerID = followerInfo.getPeer().getId();
-      final RaftPeer follower = conf.getPeer(followerID);
+      final RaftPeer follower = conf.getPeer(followerID, RaftPeerRole.FOLLOWER);
```
codings-dan, line 954: There is no need to increase this variable, the default is to take FOLLOWER

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -951,9 +951,11 @@ private void yieldLeaderToHigherPriorityPeer() {
...
       if (follower == null) {
-        LOG.error("{} the follower {} is not in the conf {}", this, server.getId(), conf);
+        if (conf.getPeer(followerID, RaftPeerRole.LISTENER) == null) {
```
codings-dan, line 956: Why do you need to add this judgment here? In the election process, it seems that listeners do not need to participate

codings-dan (changes requested):
@qian0817 Thanks for working on this. I left some comments for your reference, thank you!

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
qian0817: `senders.getSenders()` will return all peers including the listener.

ratis-common/src/main/java/org/apache/ratis/util/ProtoUtils.java
qian0817: This is to ensure forward compatibility. Using the previous raft log will cause startupRole item to not exist.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
qian0817: Enume classes will have the same result when using `==` and `equals`.

ratis-proto/src/main/proto/Raft.proto
qian0817: fix.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
qian0817: fix.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftGroup.java
qian0817: fix.

ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeer.java
codings-dan: ok, ignore my personal habits

ratis-common/src/main/java/org/apache/ratis/util/ProtoUtils.java
codings-dan: Good Catch! I forget the forward compatibility

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
codings-dan: But when designing this method parameter, we thought that the variable will only be specified when the Listener is obtained. Also, this kind of call is a lot in ratis, see below, so I think we should remove this change

image

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
codings-dan: Thanks for letting me know, I forgot about this

codings-dan (approved):
Thanks for working on this! Overall LGTM, just left some comments for reference. @szetszwo Could you help review the pull request, thank you in advance!

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -77,7 +80,14 @@ default RaftPeer getPeer() {
...
     default RaftGroup getGroup() {
-      return RaftGroup.valueOf(getMemberId().getGroupId(), getRaftConf().getAllPeers());
+      Collection<RaftPeer> allFollowerPeers =
+          getRaftConf().getAllPeers(RaftProtos.RaftPeerRole.FOLLOWER);
+      Collection<RaftPeer> allListenerPeers =
+          getRaftConf().getAllPeers(RaftProtos.RaftPeerRole.LISTENER);
+      Iterable<RaftPeer> peers = Iterables.concat(allFollowerPeers, allListenerPeers);
+      return RaftGroup.valueOf(
+          getMemberId().getGroupId(),
+          Lists.newArrayList(peers));
```
szetszwo, line 90: Change the `peers` parameter of `RaftGroup.valueOf(..)` to `Iterable<RaftPeer>`.  Then, we don't have to create a new array list.
```
    default RaftGroup getGroup() {
      final Collection<RaftPeer> followers = getRaftConf().getAllPeers(RaftPeerRole.FOLLOWER);
      final Collection<RaftPeer> listeners = getRaftConf().getAllPeers(RaftPeerRole.LISTENER);
      final Iterable<RaftPeer> peers = Iterables.concat(followers, listeners);
      return RaftGroup.valueOf(getMemberId().getGroupId(), peers);
    }
```

szetszwo (commented):
@qian0817 , thanks a lot for the update.  Some comments inlined; see also https://issues.apache.org/jira/secure/attachment/13046691/673_review.patch

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -951,9 +951,11 @@ private void yieldLeaderToHigherPriorityPeer() {
...
+          LOG.error("{} the follower {} is not in the conf {}", this, server.getId(), conf);
```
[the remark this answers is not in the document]
qian0817: fix

ratis-server/src/main/java/org/apache/ratis/server/impl/PeerConfiguration.java
```diff
@@ -97,7 +97,7 @@ int size() {
 
   @Override
   public String toString() {
-    return peers.values().toString();
+    return "peers:" + peers.values() + "|listeners:" + listeners.toString();
```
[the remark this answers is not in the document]
qian0817: fix

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -494,10 +494,10 @@ public void close() {
    */
   private synchronized boolean changeToFollower(long newTerm, boolean force, Object reason) {
     final RaftPeerRole old = role.getCurrentRole();
+    final boolean metadataUpdated = state.updateCurrentTerm(newTerm);
     if (old == RaftPeerRole.LISTENER) {
-      throw new IllegalStateException("Unexpected role " + old);
+      return metadataUpdated;
     }
-    final boolean metadataUpdated = state.updateCurrentTerm(newTerm);
```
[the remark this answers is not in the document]
qian0817: fix

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
qian0817: fix

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -492,14 +492,18 @@ public void close() {
...
     final RaftPeerRole old = role.getCurrentRole();
-    if (old == RaftPeerRole.LISTENER) {
+    final boolean metadataUpdated = state.updateCurrentTerm(newTerm);
+    if (old == RaftPeerRole.LISTENER && !allowListener) {
       throw new IllegalStateException("Unexpected role " + old);
     }
-    final boolean metadataUpdated = state.updateCurrentTerm(newTerm);
 
-    if (old != RaftPeerRole.FOLLOWER || force) {
+    if ((old != RaftPeerRole.FOLLOWER && old != RaftPeerRole.LISTENER) || force) {
```
szetszwo, line 506: This is better to be
```
if ((old != RaftPeerRole.FOLLOWER || force) && old != RaftPeerRole.LISTENER) {
```
Otherwise, `force == true` will change a Listener to a Follower.  Fortunately, when `force == true`, we always have `allowListener == false` in the current code.  Therefore, it won't cause any problems.

szetszwo (approved):
+1 the change looks good.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
qian0817: fix.

PR #675, opened 2022-07-07
RATIS-1617. Concurrent issue on LogSegmentList

Xushaohong:

Fix Concurrent issue on LogSegmentList which could lead to IndexOutOfBoundsException.
The consecutive operations should be protected by the lock.

https://issues.apache.org/jira/browse/RATIS-1617

## How was this patch tested?

Simple UT. not included in PR


szetszwo (approved):
+1 the change looks good.

PR #683, opened 2022-07-18
RATIS-1624: fix issue when set configuration to add a listener.

qian0817:

fix issue when set configuration to add a listener.

see https://issues.apache.org/jira/browse/RATIS-1624


codings-dan (commented):
@qian0817 Please take a look at the ci test error, https://github.com/apache/ratis/runs/7385492944?check_suite_focus=true#step:5:651

qian0817:
I tried to run `./dev-support/checks/unit.sh -Pserver-tests` on my machine but did not see the test fail. This should be a flaky test. @codings-dan

PR #686, opened 2022-07-20
RATIS-1629. Fix flaky TestRaftReconfigurationWithSimulatedRpc#testKillLeaderDuringReconf

ChenSammi:
https://issues.apache.org/jira/browse/RATIS-1629

This testKillLeaderDuringReconf fail randomly at the check 
RaftReconfigurationBaseTest.runTestKillLeaderDuringReconf -> RaftServerTestUtil#waitAndCheckNewConf#Assert.assertFalse(conf.containsInConf(server.getId()));

The test tries to move a cluster from
 [s0(L), s1, s2]  to 
 [s0(killed), s1(removed), s2, s3(new)]. 

There is chance that s3 cannot join the group because of timing. s3 is rejected and shutdown itself when it receives an shutdown command from the new Leader which isn't in the configuration transitional state. 

If s3 cannot join the group, the "Assert.assertFalse(conf.containsInConf(server.getId()))" will fail because the set configuration request will mot get majority responses.


drriguz:
It might be better to retry the test instead of skipping it, I've make another pull request for fixing this, please review it. See https://github.com/apache/ratis/pull/687

szetszwo (approved):
+1 the change looks good.

PR #687, opened 2022-07-20
RATIS-1629. Fix flaky TestRaftReconfigurationWithSimulatedRpc#testKillLeaderDuringReconf

drriguz:
Another approach of fix using retry.
See: https://github.com/apache/ratis/pull/686


drriguz:
This fix seems no longer necessary and I'll close it.

PR #690, opened 2022-07-21
RATIS-1630. Use maven site plugin instead of Hugo to generate ratis documents for better compatibility

drriguz:

This patch replaced Hugo with maven site plugin to generate documents in ratis-doc module.  Previously, using Hugo reqiures the user to install Hugo, and also it's not compatible to execute bash script in Windows platform. 

https://issues.apache.org/jira/browse/RATIS-1630

## How was this patch tested?

No code change. This patch is document only, and here's a screenshot of the rendered document:

image


szetszwo (approved):
+1 the change looks good.

Tried to generate the doc.  It worked well.

PR #691, opened 2022-07-22
RATIS-1633. Improve the grpc config output log information.

leo65535:

When run the rpc server, always output redundant and wrong log, like admin.port、client.port.
```
2022-07-22 09:24:21 INFO  RaftServer:46 - raft.rpc.type = GRPC (default)
2022-07-22 09:24:21 INFO  GrpcFactory:48 - PERFORMANCE WARNING: useCacheForAllThreads is true that may cause Netty to create a lot garbage objects and, as a result, trigger GC.
	It is recommended to disable useCacheForAllThreads by setting -Dorg.apache.ratis.thirdparty.io.netty.allocator.useCacheForAllThreads=false in command line.
2022-07-22 09:24:21 INFO  GrpcConfigKeys:46 - raft.grpc.admin.port = -1 (default)
2022-07-22 09:24:21 INFO  GrpcConfigKeys:46 - raft.grpc.server.port = 10024 (custom)
2022-07-22 09:24:21 INFO  GrpcConfigKeys:46 - raft.grpc.client.port = -1 (default)
2022-07-22 09:24:21 INFO  GrpcConfigKeys:46 - raft.grpc.server.port = 10024 (custom)
2022-07-22 09:24:21 INFO  GrpcConfigKeys:46 - raft.grpc.server.port = 10024 (custom)
2022-07-22 09:24:21 INFO  GrpcService:46 - raft.grpc.message.size.max = 64MB (=67108864) (default)
2022-07-22 09:24:21 INFO  RaftServerConfigKeys:46 - raft.server.log.appender.buffer.byte-limit = 4MB (=4194304) (default)
2022-07-22 09:24:21 INFO  GrpcService:46 - raft.grpc.flow.control.window = 1MB (=1048576) (default)
```

After patch
```
2022-07-21 17:36:03 INFO  RaftServer:46 - raft.rpc.type = GRPC (default)
2022-07-21 17:36:03 INFO  GrpcFactory:48 - PERFORMANCE WARNING: useCacheForAllThreads is true that may cause Netty to create a lot garbage objects and, as a result, trigger GC.
	It is recommended to disable useCacheForAllThreads by setting -Dorg.apache.ratis.thirdparty.io.netty.allocator.useCacheForAllThreads=false in command line.
2022-07-21 17:36:03 INFO  GrpcConfigKeys:46 - raft.grpc.admin.port = 10024 (custom)
2022-07-21 17:36:03 INFO  GrpcConfigKeys:46 - raft.grpc.client.port = 10024 (custom)
2022-07-21 17:36:03 INFO  GrpcConfigKeys:46 - raft.grpc.server.port = 10024 (custom)
2022-07-21 17:36:03 INFO  GrpcService:46 - raft.grpc.message.size.max = 64MB (=67108864) (default)
2022-07-21 17:36:03 INFO  RaftServerConfigKeys:46 - raft.server.log.appender.buffer.byte-limit = 4MB (=4194304) (default)
2022-07-21 17:36:03 INFO  GrpcService:46 - raft.grpc.flow.control.window = 1MB (=1048576) (default)
```

https://issues.apache.org/jira/browse/RATIS-1633

## How was this patch tested?

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/conf/ConfUtils.java
+++ b/ratis-common/src/main/java/org/apache/ratis/conf/ConfUtils.java
@@ -154,6 +154,14 @@ public interface ConfUtils {
   }
 
   @SafeVarargs
+  static int getInt(
+      BiFunction<String, Integer, Integer> integerGetter,
+      String key, int defaultValue, int fallbackValue,
+      Consumer<String> logger, BiConsumer<String, Integer>... assertions) {
+    return get(integerGetter, key, defaultValue, fallbackValue, logger, assertions);
+  }
+
+  @SafeVarargs
   static long getLong(
       BiFunction<String, Long, Long> longGetter,
       String key, long defaultValue, Consumer<String> logger, BiConsumer<String, Long>... assertions) {
@@ -212,6 +220,19 @@ public interface ConfUtils {
     final T value = getter.apply(key, defaultValue);
     logGet(key, value, defaultValue, logger);
     Arrays.asList(assertions).forEach(a -> a.accept(key, value));
+    return value;
+  }
+
+  @SafeVarargs
+  static <T> T get(BiFunction<String, T, T> getter,
+      String key, T defaultValue, T fallbackValue,
+      Consumer<String> logger, BiConsumer<String, T>... assertions) {
+    T value = getter.apply(key, defaultValue);
+    value = value != defaultValue ? value : fallbackValue;
+    logGet(key, value, defaultValue, logger);
+
+    T finalValue = value;
+    Arrays.asList(assertions).forEach(a -> a.accept(key, finalValue));
     return value;
   }
 
--- a/ratis-common/src/main/java/org/apache/ratis/conf/ConfUtils.java
+++ b/ratis-common/src/main/java/org/apache/ratis/conf/ConfUtils.java
@@ -227,12 +227,9 @@ public interface ConfUtils {
   static <T> T get(BiFunction<String, T, T> getter,
       String key, T defaultValue, T fallbackValue,
       Consumer<String> logger, BiConsumer<String, T>... assertions) {
-    T value = getter.apply(key, defaultValue);
+    T value = get(getter, key, defaultValue, null, assertions);
     value = value != defaultValue ? value : fallbackValue;
     logGet(key, value, defaultValue, logger);
-
-    T finalValue = value;
-    Arrays.asList(assertions).forEach(a -> a.accept(key, finalValue));
     return value;
   }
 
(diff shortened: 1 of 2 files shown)


adoroszlai (approved):
Thanks @leo65535 for the improvement.

leo65535:
Thanks @adoroszlai

> I don't think admin.port = -1 (default) is wrong, but it may be confusing if you don't know about the fallback behavior.

We can add some descriptions to user document.

> There is a difference between a configuration that has explicit custom settings for client/admin ports, and one that has no setting and falls back to the server port value.

Emm, it keeps the same logic as before, it seems not a problem. Or, maybe I don't get your points.

image

adoroszlai:
> > There is a difference between a configuration that has explicit custom settings for client/admin ports, and one that has no setting and falls back to the server port value.
> 
> Emm, it keeps the same logic as before, it seems not a problem. Or, maybe I don't get your points.

I didn't mean the patch breaks anything.

Consider two different configurations:

```
raft.grpc.server.port=12345
```

and 

```
raft.grpc.admin.port=12345
raft.grpc.client.port=12345
raft.grpc.server.port=12345
```

Ratis works the same way with both due to the fallback logic.  Prior to the patch they were logged differently (`-1 (default)` for admin/client in the first case, `12345 (custom)` in the second case), but with the patch they both result in the same log output.

Now assume we only change `raft.grpc.server.port` to a different value in both cases.  Ratis behavior is different for these two configurations (regardless of the patch), since the first one uses the new port as fallback, while the second one has explicit custom config for admin/client ports.

My point is that this difference in log messages may be useful in some cases.  (E.g. if someone else is making the config change and you only see the logs.)

But the new log content is probably better in most cases.

leo65535:
Thanks @adoroszlai, I've learned most of your points.

You are right, they were logged differently  (-1 (default) for admin/client before. But the fallback logic has always existed,
although the default value is logged, but the fallback port is actually used. Just from my side, it seems the patch doesn't need to be modified any more.

image

adoroszlai:
> it seems the patch doesn't need to be modified any more

Agreed, I didn't ask for that.

leo65535:
> > it seems the patch doesn't need to be modified any more
> 
> Agreed, I didn't ask for that.

Thanks.

szetszwo:
@leo65535 , it is good to print out the effective value but we should print also the valued set in conf.  How about the following?
```
raft.grpc.client.port is set to -1 (default); the effective value is 10024 (=raft.grpc.server.port)
```

szetszwo:
Oops, I accidentally merged this.  Let's me file a followup JIRA instead of reverting this.

szetszwo:
Let's make the message shorter as below
```
raft.grpc.admin.port = 10024 (fallback to raft.grpc.server.port)
```
Filed https://issues.apache.org/jira/browse/RATIS-1637

PR #693, opened 2022-07-22
RATIS-1634. Close SegmentedRaftLogReader if init failed.

drriguz:

In SegmentedRaftLogInputStream#init, the SegmentedRaftLogReader instance is not correctly closed if the log is corrupted. The LogSegment#readSegmentFile function will try to delete empty segment, however it may fail on Windows because previously it's not closed. 

https://issues.apache.org/jira/browse/RATIS-1634

## How was this patch tested?

Already coverd in unit test ServerRestartTests#testRestartWithCorruptedLogHeader.


szetszwo (approved):
+1 the change looks good.

PR #707, opened 2022-08-07
RATIS-1661. Support configurable hostname in GrpcService

SzyWilliam:
GrpcService will use `NettyServerBuilder.forport` to bind the listening address, which is by default 0.0.0.0::port. This may bring potential security risks.
In this PR I propose to use a configurable hostname to construct the binding address. 
BTW, the InetSocketAddress support ipv6 address, so this PR can also enable users to assign ipv6 address to grpc listening address.

https://issues.apache.org/jira/browse/RATIS-1661

## How was this patch tested?
unit tests.


SzyWilliam:
@szetszwo Please take a look at this, thanks!

szetszwo (approved):
+1 the change looks good.

PR #720, opened 2022-08-22
RATIS-1681. Use atomic_move to enhance robustness

SzyWilliam:

Use StandardCopyOption.ATOMIC_MOVE to enhance robustness.

https://issues.apache.org/jira/browse/RATIS-1681

## How was this patch tested?
unit tests

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/util/FileUtils.java
+++ b/ratis-common/src/main/java/org/apache/ratis/util/FileUtils.java
@@ -76,9 +76,15 @@ public interface FileUtils {
   }
 
   static void move(Path src, Path dst) throws IOException {
-    LogUtils.runAndLog(LOG,
+    try {
+      LogUtils.runAndLog(LOG,
         () -> Files.move(src, dst, StandardCopyOption.ATOMIC_MOVE),
-        () -> "Files.move " + src + " to " + dst);
+        () -> "Atomic Files.move " + src + " to " + dst);
+    } catch (AtomicMoveNotSupportedException e) {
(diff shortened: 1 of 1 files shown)


ratis-common/src/main/java/org/apache/ratis/util/FileUtils.java
```diff
@@ -77,7 +77,7 @@ static void move(File src, File dst) throws IOException {
 
   static void move(Path src, Path dst) throws IOException {
     LogUtils.runAndLog(LOG,
-        () -> Files.move(src, dst),
+        () -> Files.move(src, dst, StandardCopyOption.ATOMIC_MOVE),
```
adoroszlai, line 80: According to [javadoc](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#move-java.nio.file.Path-java.nio.file.Path-java.nio.file.CopyOption...-), this would fail with exception if move cannot be performed as an atomic operation.  If I understand correctly, moving across filesystems is such a non-atomic operation, performed as "copy and delete" instead.

Should we retry without the option in case of `AtomicMoveNotSupportedException`?

SzyWilliam: Thanks for your review! Indeed we may encounter cross disk move, so I'll add the fallback retry logic.

adoroszlai: Thanks for updating the patch, LGTM.

szetszwo (approved):
+1 the change looks good.

PR #729, opened 2022-08-26
RATIS-1691. Potential deadlock in server shutdown

adoroszlai:

Move shutdown of `RaftServerProxy#executor` inside the block which is executed after lifecycle check.  This prevents subsequent callers (notably `StateMachineUpdater`) from blocking until the executor is shutdown.

https://issues.apache.org/jira/browse/RATIS-1691

## How was this patch tested?

Ran all Ratis unit tests locally.

Ran Ozone integration tests with 2.4.0-rc0 + RATIS-1687 + this change.  Most tests passed (most importantly HA tests that timed out previously).

Regular CI:
https://github.com/adoroszlai/incubator-ratis/actions/runs/2935727046


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #733, opened 2022-09-03
RATIS-1695 Improve error observability and state mgmt in RaftServer

jiacheliu3:

1. Updates the RaftServer API so external user can retrieve the internal exception in Ratis
2. Adds UncaughtExceptionHandler to most critical `Daemon` threads so uncaught exceptions can propagated to the RaftServer level

https://issues.apache.org/jira/browse/RATIS-1695

## How was this patch tested?

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
+++ b/ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
@@ -17,14 +17,36 @@
  */
 package org.apache.ratis.util;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import java.util.concurrent.atomic.AtomicReference;
+
 public class Daemon extends Thread {
   {
     setDaemon(true);
   }
+  // TODO(jiacheng): Ideally, use the logger from the thread class itself
+  static final Logger LOG = LoggerFactory.getLogger(Daemon.class);
+
+  /** If the thread meets an uncaught exception, this field will be set. */
+  private final AtomicReference<Throwable> throwable = new AtomicReference<>(null);
+  protected Stated statedServer;
 
   /** Construct a daemon thread. */
+  // TODO(jiacheng): Consolidate all constructors
   public Daemon() {
     super();
+    setUncaughtExceptionHandler((thread, t) -> {
+      onError(t);
+    });
+  }
+
+  public Daemon(String name, Stated server) {
+    this();
+    this.setName(name);
+    this.statedServer = server;
   }
 
   /** Construct a daemon thread with the given runnable. */
@@ -36,5 +58,34 @@ public class Daemon extends Thread {
   public Daemon(Runnable runnable, String name) {
     super(runnable);
     this.setName(name);
+  }
+
+  public Daemon(Runnable runnable, String name, Stated server) {
+    this(runnable, name);
+    this.statedServer = server;
+  }
+
+  /**
+   * Handles the uncaught error on thread crashing.
+   *
+   * @param t the crashing error
+   */
+  public void onError(Throwable t) {
+    throwable.set(t);
+    if (statedServer != null) {
+      LOG.error("Daemon thread {} exiting due to an uncaught exception, marking RaftServer {} state to ERROR",
+              getName(), statedServer, t);
+      statedServer.setError(t);
+      // TODO(jiacheng): Transition the server state to ERROR
+    }
+
+    // TODO(jiacheng): should i set the lifecycle to close and exit? or it is possible to recover?
+    //  Do a RaftServer state transition in a heartbeat thread
+    // TODO(jiacheng): what if this thread is created in a threadpool?
+  }
+
+  @Nullable
+  public Throwable getError() {
+    return throwable.get();
   }
 }
--- a/ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
+++ b/ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
@@ -34,6 +34,7 @@ import org.apache.ratis.thirdparty.com.google.common.collect.Iterables;
 import org.apache.ratis.util.IOUtils;
 import org.apache.ratis.util.LifeCycle;
 import org.apache.ratis.util.ReflectionUtils;
+import org.apache.ratis.util.Stated;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -49,11 +50,11 @@ import java.util.Optional;
 public interface RaftServer extends Closeable, RpcType.Get,
     RaftServerProtocol, RaftServerAsynchronousProtocol,
     RaftClientProtocol, RaftClientAsynchronousProtocol,
-    AdminProtocol, AdminAsynchronousProtocol {
+    AdminProtocol, AdminAsynchronousProtocol, Stated {
   Logger LOG = LoggerFactory.getLogger(RaftServer.class);
 
   /** A division of a {@link RaftServer} for a particular {@link RaftGroup}. */
-  interface Division extends Closeable {
+  interface Division extends Closeable, Stated {
     Logger LOG = LoggerFactory.getLogger(Division.class);
 
     /** @return the {@link DivisionProperties} for this division. */
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/FollowerState.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/FollowerState.java
@@ -65,6 +65,7 @@ class FollowerState extends Daemon {
   private final AtomicInteger outstandingOp = new AtomicInteger();
 
   FollowerState(RaftServerImpl server, Object reason) {
+    // implicitly calling super() here
     this.name = server.getMemberId() + "-" + JavaUtils.getClassSimpleName(getClass());
     this.setName(this.name);
     this.server = server;
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderElection.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderElection.java
@@ -138,6 +138,7 @@ class LeaderElection implements Runnable {
 
     Executor(Object name, int size) {
       Preconditions.assertTrue(size > 0);
+      // TODO(jiacheng): intentionally ignoring the server ref here? Why?
       executor = Executors.newFixedThreadPool(size, r -> new Daemon(r, name + "-" + count.incrementAndGet()));
       service = new ExecutorCompletionService<>(executor);
     }
@@ -190,7 +191,7 @@ class LeaderElection implements Runnable {
   LeaderElection(RaftServerImpl server, boolean skipPreVote) {
     this.name = server.getMemberId() + "-" + JavaUtils.getClassSimpleName(getClass()) + COUNT.incrementAndGet();
     this.lifeCycle = new LifeCycle(this);
-    this.daemon = new Daemon(this);
+    this.daemon = new Daemon(this, name, server);
     this.server = server;
     this.skipPreVote = skipPreVote ||
         !RaftServerConfigKeys.LeaderElection.preVote(
@@ -281,6 +282,7 @@ class LeaderElection implements Runnable {
       r = new ResultAndTerm(Result.PASSED, electionTerm);
     } else {
       final TermIndex lastEntry = server.getState().getLastEntry();
+      // TODO(jiacheng): The daemon can actually have the server ref here! Do we want that?
       final Executor voteExecutor = new Executor(this, others.size());
       try {
         final int submitted = submitRequests(phase, electionTerm, lastEntry, others, voteExecutor);
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -69,6 +69,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.NoSuchFileException;
 import java.util.*;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
@@ -97,7 +98,7 @@ import org.apache.ratis.util.function.CheckedSupplier;
 
 class RaftServerImpl implements RaftServer.Division,
     RaftServerProtocol, RaftServerAsynchronousProtocol,
-    RaftClientProtocol, RaftClientAsynchronousProtocol{
+    RaftClientProtocol, RaftClientAsynchronousProtocol, Stated {
   private static final String CLASS_NAME = JavaUtils.getClassSimpleName(RaftServerImpl.class);
   static final String REQUEST_VOTE = CLASS_NAME + ".requestVote";
   static final String APPEND_ENTRIES = CLASS_NAME + ".appendEntries";
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
@@ -374,6 +374,7 @@ class RaftServerProxy implements RaftServer {
 
   @Override
   public LifeCycle.State getLifeCycleState() {
+    // TODO(jiacheng): How do i expose the exception from here?
     return lifeCycle.getCurrentState();
   }
 
--- a/ratis-server/src/main/java/org/apache/ratis/server/leader/LogAppenderDaemon.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/leader/LogAppenderDaemon.java
@@ -17,6 +17,7 @@
  */
 package org.apache.ratis.server.leader;
 
+import org.apache.ratis.server.RaftServer;
 import org.apache.ratis.util.Daemon;
 import org.apache.ratis.util.JavaUtils;
 import org.apache.ratis.util.LifeCycle;
@@ -25,6 +26,7 @@ import java.io.InterruptedIOException;
 import java.util.function.UnaryOperator;
 
 import org.apache.ratis.util.LifeCycle.State;
+import org.apache.ratis.util.Stated;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -44,11 +46,11 @@ class LogAppenderDaemon {
 
   private final LogAppender logAppender;
 
-  LogAppenderDaemon(LogAppender logAppender) {
+  LogAppenderDaemon(LogAppender logAppender, Stated server) {
     this.logAppender = logAppender;
     this.name = logAppender + "-" + JavaUtils.getClassSimpleName(getClass());
     this.lifeCycle = new LifeCycle(name);
-    this.daemon = new Daemon(this::run, name);
+    this.daemon = new Daemon(this::run, name, server);
   }
 
   public boolean isWorking() {
--- a/ratis-common/src/main/java/org/apache/ratis/util/Stated.java
+++ b/ratis-common/src/main/java/org/apache/ratis/util/Stated.java
@@ -1,0 +1,10 @@
+package org.apache.ratis.util;
+
+import javax.annotation.Nullable;
+
+public interface Stated {
+  void setError(Throwable t);
+
+  @Nullable
+  Throwable getError();
+}
--- a/ratis-server-api/src/main/java/org/apache/ratis/server/DivisionInfo.java
+++ b/ratis-server-api/src/main/java/org/apache/ratis/server/DivisionInfo.java
@@ -22,11 +22,12 @@ import org.apache.ratis.proto.RaftProtos.RaftPeerRole;
 import org.apache.ratis.proto.RaftProtos.RoleInfoProto;
 import org.apache.ratis.protocol.RaftPeerId;
 import org.apache.ratis.util.LifeCycle;
+import org.apache.ratis.util.Stated;
 
 /**
  * Information of a raft server division.
  */
-public interface DivisionInfo {
+public interface DivisionInfo extends Stated {
   /** @return the current role of this server division. */
   RaftPeerRole getCurrentRole();
 
@@ -58,6 +59,7 @@ public interface DivisionInfo {
    */
   RaftPeerId getLeaderId();
 
+  // TODO(jiacheng): How is this lifecycle mapped to RaftServer lifecycle? See RaftServerProxy
   /** @return the life cycle state of this server division. */
   LifeCycle.State getLifeCycleState();
 
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
@@ -621,7 +621,7 @@ class LeaderStateImpl implements LeaderState {
    */
   private class EventProcessor extends Daemon {
     public EventProcessor(String name) {
-      setName(name);
+      super(name, server);
     }
     @Override
     public void run() {
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/StateMachineUpdater.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/StateMachineUpdater.java
@@ -110,7 +110,7 @@ class StateMachineUpdater implements Runnable {
     };
     this.purgeUptoSnapshotIndex = RaftServerConfigKeys.Log.purgeUptoSnapshotIndex(properties);
 
-    updater = new Daemon(this);
+    updater = new Daemon(this, name, server);
     this.awaitForSignal = new AwaitForSignal(name);
   }
 
--- a/ratis-server/src/main/java/org/apache/ratis/server/leader/LogAppenderBase.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/leader/LogAppenderBase.java
@@ -69,7 +69,7 @@ public abstract class LogAppenderBase implements LogAppender {
     final SizeInBytes bufferByteLimit = RaftServerConfigKeys.Log.Appender.bufferByteLimit(properties);
     final int bufferElementLimit = RaftServerConfigKeys.Log.Appender.bufferElementLimit(properties);
     this.buffer = new DataQueue<>(this, bufferByteLimit, bufferElementLimit, EntryWithData::getSerializedSize);
-    this.daemon = new LogAppenderDaemon(this);
+    this.daemon = new LogAppenderDaemon(this, server);
     this.eventAwaitForSignal = new AwaitForSignal(name);
   }
 


ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -97,14 +98,31 @@
...
 
+  private Throwable throwable;
+
+  @Override
+  public void setError(Throwable t) {
+    throwable = t;
+    LOG.error("Server transitioning to EXCEPTION state due to", t);
+    // TODO(jiacheng): will the server keep serving or just die itself?
+    //  What is the cleanup I should consider?
+    lifeCycle.transition(LifeCycle.State.EXCEPTION);
```
jiacheliu3, line 117: The implication is, many methods in `RaftServerImpl` rely on the state to be `RUNNING` so operations will just fail. I wanted to double check with you guys if that's okay. We might wanna distinguish errors that should crash the server vs ones that should not (we want the RaftServer to hide the error from the outside observer and just retry/recover internally).

The good news is, if we don't want something to crash the server, just don't call this `ErrorRecorded.setError()` here. 

Specifically to this PR, I presume all uncaught exceptions should crash the server because otherwise they should be explicitly caught and eaten. WDYT?

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -97,14 +98,31 @@
...
   static final String LOG_SYNC = APPEND_ENTRIES + ".logComplete";
   static final String START_LEADER_ELECTION = CLASS_NAME + ".startLeaderElection";
 
+  private Throwable throwable;
+
+  @Override
+  public void setError(Throwable t) {
+    throwable = t;
+    LOG.error("Server transitioning to EXCEPTION state due to", t);
+    // TODO(jiacheng): will the server keep serving or just die itself?
```
jiacheliu3, line 115: Sorry I'm new to Ratis code in general so not sure what's the recommended design. Will this be a good place for the server to exit, or it should check this state and exit somewhere else?

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
```diff
@@ -407,6 +426,7 @@ public void close() {
       try {
         getServerRpc().close();
       } catch(IOException ignored) {
+        // TODO(jiacheng): transition state to EXCEPTION here?
```
jiacheliu3, line 429: In the existing code I don't think we transition to EXCEPTION very often. In many places like this, we might wanna transition the state and expose the exception to the outside. And the Ratis user can just observe the `RaftServer` lifecycle and choose to restart the RaftServer or do whatever else.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderElection.java
```diff
@@ -138,6 +138,7 @@ static class Executor {
 
     Executor(Object name, int size) {
       Preconditions.assertTrue(size > 0);
+      // TODO(jiacheng): intentionally ignoring the server ref here? Why?
       executor = Executors.newFixedThreadPool(size, r -> new Daemon(r, name + "-" + count.incrementAndGet()));
```
jiacheliu3, line 142: Intentionally not passing the `RaftServerImpl` reference to the Daemon thread so a dead thread doesn't set the RaftServerImpl state to EXCEPTION. My intuition is that a failed prevote doesn't matter that much because it will be retried. Could someone confirm?

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
```diff
@@ -37,4 +57,32 @@ public Daemon(Runnable runnable, String name) {
     super(runnable);
     this.setName(name);
   }
+
+  public Daemon(Runnable runnable, String name, ErrorRecorded server) {
```
jiacheliu3, line 61: I ignored `Daemon` threads created in `JvmPauseMonitor` and `TimeoutScheduler` because my intuition is even if they crash, the RaftServer can still live.

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
```diff
@@ -17,24 +17,77 @@
...
+  @Nullable
+  public Throwable getError() {
+    return throwable.get();
   }
 
-  /** Construct a daemon thread with the given runnable. */
-  public Daemon(Runnable runnable, String name) {
-    super(runnable);
-    this.setName(name);
+  public static class Builder {
```
jiacheliu3, line 69: Use the builder patter to avoid having too many overloaded constructors

jiacheliu3:
@codings-dan PTAL if the design meets expectation and pls feel free to add any reviewers to the discussion. If you are all okay with the idea and design, I will move on to add unit tests.

On unit tests, I'm honestly not sure what you would expect me to add. I mean, this PR adds UnexpectedExceptionHandlers then I guess I should add tests to trigger unhandled exceptions to each Daemon thread? Or just having one test on the Daemon thread class itself would be acceptable? I'm new to Ratis repo so I'd like to conform to how you normally test. Any advice is appreciated!

codings-dan:
@jiacheliu3 Thanks for working on this, I will review the change recently @szetszwo Could you help take a look at this pull request?

szetszwo:
@codings-dan , sure, will review this.

@jiacheliu3 , thanks a lot for working this!  There were some failures in https://github.com/apache/ratis/actions/runs/2984287869 .  Could you take a look?

jiacheliu3:
Is this one flaky? This UT passes on my local build.
```
Error:  Tests run: 16, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 42.736 s <<< FAILURE! - in org.apache.ratis.server.simulation.TestLeaderElectionWithSimulatedRpc
Error:  testRemoveListener(org.apache.ratis.server.simulation.TestLeaderElectionWithSimulatedRpc)  Time elapsed: 1.711 s  <<< ERROR!
java.lang.IllegalStateException: Found duplicated element s0|rpc:localhost:38565|admin:localhost:42183|client:localhost:37945|dataStream:localhost:42749|priority:0|startupRole:FOLLOWER in [s0|rpc:localhost:38565|admin:localhost:42183|client:localhost:37945|dataStream:localhost:42749|priority:0|startupRole:FOLLOWER, s2|rpc:localhost:40489|admin:localhost:40079|client:localhost:44937|dataStream:localhost:40043|priority:0|startupRole:FOLLOWER, s0|rpc:localhost:38565|admin:localhost:42183|client:localhost:37945|dataStream:localhost:42749|priority:0|startupRole:FOLLOWER]
	at org.apache.ratis.util.Preconditions.assertTrue(Preconditions.java:73)
	at org.apache.ratis.util.Preconditions.assertUnique(Preconditions.java:128)
	at org.apache.ratis.util.Preconditions.assertUnique(Preconditions.java:122)
	at org.apache.ratis.protocol.SetConfigurationRequest$Arguments.<init>(SetConfigurationRequest.java:59)
	at org.apache.ratis.protocol.SetConfigurationRequest$Arguments.<init>(SetConfigurationRequest.java:36)
	at org.apache.ratis.protocol.SetConfigurationRequest$Arguments$Builder.build(SetConfigurationRequest.java:144)
	at org.apache.ratis.client.api.AdminApi.setConfiguration(AdminApi.java:55)
	at org.apache.ratis.client.api.AdminApi.setConfiguration(AdminApi.java:40)
	at org.apache.ratis.server.impl.LeaderElectionTests.testRemoveListener(LeaderElectionTests.java:353)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:299)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:293)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

codings-dan:
> Is this one flaky? This UT passes on my local build.
> 
> ```
> Error:  Tests run: 16, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 42.736 s <<< FAILURE! - in org.apache.ratis.server.simulation.TestLeaderElectionWithSimulatedRpc
> Error:  testRemoveListener(org.apache.ratis.server.simulation.TestLeaderElectionWithSimulatedRpc)  Time elapsed: 1.711 s  <<< ERROR!
> java.lang.IllegalStateException: Found duplicated element s0|rpc:localhost:38565|admin:localhost:42183|client:localhost:37945|dataStream:localhost:42749|priority:0|startupRole:FOLLOWER in [s0|rpc:localhost:38565|admin:localhost:42183|client:localhost:37945|dataStream:localhost:42749|priority:0|startupRole:FOLLOWER, s2|rpc:localhost:40489|admin:localhost:40079|client:localhost:44937|dataStream:localhost:40043|priority:0|startupRole:FOLLOWER, s0|rpc:localhost:38565|admin:localhost:42183|client:localhost:37945|dataStream:localhost:42749|priority:0|startupRole:FOLLOWER]
> 	at org.apache.ratis.util.Preconditions.assertTrue(Preconditions.java:73)
> 	at org.apache.ratis.util.Preconditions.assertUnique(Preconditions.java:128)
> 	at org.apache.ratis.util.Preconditions.assertUnique(Preconditions.java:122)
> 	at org.apache.ratis.protocol.SetConfigurationRequest$Arguments.<init>(SetConfigurationRequest.java:59)
> 	at org.apache.ratis.protocol.SetConfigurationRequest$Arguments.<init>(SetConfigurationRequest.java:36)
> 	at org.apache.ratis.protocol.SetConfigurationRequest$Arguments$Builder.build(SetConfigurationRequest.java:144)
> 	at org.apache.ratis.client.api.AdminApi.setConfiguration(AdminApi.java:55)
> 	at org.apache.ratis.client.api.AdminApi.setConfiguration(AdminApi.java:40)
> 	at org.apache.ratis.server.impl.LeaderElectionTests.testRemoveListener(LeaderElectionTests.java:353)
> 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
> 	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
> 	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
> 	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
> 	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
> 	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
> 	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:299)
> 	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:293)
> 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
> 	at java.base/java.lang.Thread.run(Thread.java:829)
> ```

let me re-run the ci test

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
szetszwo: When it is in CLOSING state, it must transit to CLOSED but won't transit to EXCEPTION.  CLOSING must handle the exceptions, if there are any.

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
```diff
@@ -17,24 +17,76 @@
...
+    private final String name;
+    private Runnable runnable;
+    private ErrorRecorded statedServer;
```
szetszwo, line 72: Let's pass an `UncaughtExceptionHandler` instead of adding ErrorRecorded.

ratis-common/src/main/java/org/apache/ratis/util/TimeoutScheduler.java
```diff
@@ -84,7 +84,10 @@ ScheduledFuture<?> schedule(Runnable task, Supplier<String> name, TimeDuration t
 
     private static ScheduledThreadPoolExecutor newExecutor() {
       LOG.debug("new ScheduledThreadPoolExecutor");
-      final ScheduledThreadPoolExecutor e = new ScheduledThreadPoolExecutor(1, (ThreadFactory) Daemon::new);
+      AtomicInteger count = new AtomicInteger(0);
```
szetszwo, line 87: It should be private static final:
```java
    private static final AtomicInteger COUNT = new AtomicInteger();
```

ratis-common/src/main/java/org/apache/ratis/util/ErrorRecorded.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package org.apache.ratis.util;
+
+import javax.annotation.Nullable;
+
+/**
+ * An internal error will be recorded so it can be later retrieved for report or recovery.
+ */
+public interface ErrorRecorded {
```
szetszwo, line 25: Pass a `UncaughtExceptionHandler` instead of adding this interface.

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
```diff
@@ -17,24 +17,76 @@
...
+
+    public Builder(String name) {
```
szetszwo, line 74: Instead of adding a constructor, please add
```java
  /** @return a {@link Builder}. */
  static Builder newBuilder() {
    return new Builder();
  }
```

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
szetszwo: We should let the application decide.  Therefore, we should pass an `UncaughtExceptionHandler` from the application.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
szetszwo: Pass an `UncaughtExceptionHandler`.

szetszwo (changes requested):
@jiacheliu3 , thanks again for working on this.  It is better to pass an `UncaughtExceptionHandler` so that the application can decide what to do.  Please see the comments inlined.

jiacheliu3:
@szetszwo Thanks for the comments! What do you think about exposing the exception on the `RaftServer`? I come from Alluxio and we use Ratis RaftServer to manage the file system metadata. 
https://github.com/Alluxio/alluxio/blob/aacf14507174029fac4a997b6037207fae92ea91/core/server/common/src/main/java/alluxio/master/journal/raft/RaftJournalSystem.java#L429
It'll be very helpful if the lifecycle state on the RaftServer is more up-to-date and more ideally, we can see what error caused RaftServer to fail. It is fair if you say we don't expose the exception outside RaftServer (because the external party normally can't do better recovery than the RaftServer is capable of), but it could be more extensible to future use cases if we do expose the exception out of RaftServer. WDYT?

szetszwo:
@jiacheliu3 , What do you think about exposing the exception on the `RaftServer`?

It is a good idea.  We can do it by passing an `UncaughtExceptionHandler`; see https://issues.apache.org/jira/secure/attachment/13048963/733_review.patch

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
jiacheliu3: I intended to force all threads to have a valid name (otherwise java Thread.init() will throw an exception anyways). Do you see use cases where the caller creates a Daemon and `setName()` later? WDYT?

ratis-common/src/main/java/org/apache/ratis/util/ErrorRecorded.java
jiacheliu3: will do

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
jiacheliu3: This might be a larger question than this particular fix but, would an `UncaughtExceptionHandler` be too specific? I imagine the application prefers something like a pre-crash hook that's applied to all places where the RaftServer is going to crash or be set to `EXCEPTION` lifecycle state.

I'm good with just an `UncaughtExceptionHandler` for now of course because that's what my application needs.

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -166,9 +170,10 @@ class Builder {
     private static final Method NEW_RAFT_SERVER_METHOD = initNewRaftServerMethod();
 
     private static Method initNewRaftServerMethod() {
+      // TODO(jiacheng): backward compatibility?
```
jiacheliu3, line 173: Consider how to keep the most backward compatibility as possible

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -168,21 +172,28 @@ class Builder {
...
+          StateMachine.Registry.class, RaftProperties.class, Parameters.class, Thread.UncaughtExceptionHandler.class};
       try {
         final Class<?> clazz = ReflectionUtils.getClassByName(className);
-        return clazz.getMethod("newRaftServer", argClasses);
+        try {
+          return clazz.getMethod("newRaftServer", argClasses);
+        } catch (NoSuchMethodException e) {
+          final Class<?>[] argClassesNoExHandler = {RaftPeerId.class, RaftGroup.class, RaftStorage.StartupOption.class,
+              StateMachine.Registry.class, RaftProperties.class, Parameters.class};
+          return ReflectionUtils.getClassByName(className).getMethod("newRaftServer", argClassesNoExHandler);
```
jiacheliu3, line 183: This is an ugly attempt to keep backward compatibility. I totally have no idea how many people have in-house implementations of `RaftServer` and things may break because of my change. PTAL if this is okay and i'll really appreciate it if you could point out a better way :)

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -249,5 +262,11 @@ public Builder setParameters(Parameters parameters) {
       this.parameters = parameters;
       return this;
     }
+
+    // TODO(jiacheng): add UTs
```
jiacheliu3, line 266: If this structure looks okay to reviewers, I will add some UTs to cover the logics I added

ratis-server/src/main/java/org/apache/ratis/server/impl/FollowerState.java
```diff
@@ -65,8 +64,9 @@ int update(AtomicInteger outstanding) {
   private final AtomicInteger outstandingOp = new AtomicInteger();
 
   FollowerState(RaftServerImpl server, Object reason) {
-    this.name = server.getMemberId() + "-" + JavaUtils.getClassSimpleName(getClass());
-    this.setName(this.name);
+    super(newBuilder()
```
jiacheliu3, line 67: this is to get rid of the extra `Daemon()` no-arg constructor

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
jiacheliu3: FYI this logic is gone tgt with the `ErrorRecorded` interface. I no longer have this default behavior in the `RaftServerProxy` because it now takes an `uncaughtExceptionHandler` from the application. Is that okay? @szetszwo

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
jiacheliu3: @szetszwo PTAL if this is okay, thanks a lot!

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -116,6 +116,8 @@ default RaftGroup getGroup() {
     /** @return the internal {@link RaftClient} of this division. */
     RaftClient getRaftClient();
 
+    Thread.UncaughtExceptionHandler getUncaughtExceptionHandler();
```
szetszwo, line 119: Remove it since this is an internal API but not a public API.

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -156,6 +158,8 @@ default RpcType getRpcType() {
 
   LifeCycle.State getLifeCycleState();
 
+  Thread.UncaughtExceptionHandler getUncaughtExceptionHandler();
+
```
szetszwo, line 162: Same as before, remove this method declaration.

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
```diff
@@ -17,24 +17,59 @@
...
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Objects;
+
 public class Daemon extends Thread {
+  static final Logger LOG = LoggerFactory.getLogger(Daemon.class);
+  public static final Thread.UncaughtExceptionHandler LOG_EXCEPTION =
+      (t, e) -> LOG.error(t.getName() + " threw an uncaught exception", e);
```
szetszwo, line 28: Move them to MiniRaftCluster since they are only used there.

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
```diff
@@ -17,24 +17,59 @@
...
 
-  /** Construct a daemon thread with the given runnable. */
-  public Daemon(Runnable runnable, String name) {
-    super(runnable);
-    this.setName(name);
+  public static class Builder {
+    private String name;
+    private Runnable runnable;
+    // By default, uncaught exceptions are just logged without further actions
+    private UncaughtExceptionHandler uncaughtExceptionHandler = LOG_EXCEPTION;
```
szetszwo, line 53: Use null for default.

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
```diff
@@ -17,24 +17,59 @@
...
-  /** Construct a daemon thread. */
-  public Daemon() {
-    super();
+  /** Construct a daemon thread with flexible arguments. */
+  protected Daemon(Builder builder) {
+    super(builder.runnable);
+    setName(builder.name);
+    if (builder.uncaughtExceptionHandler != null) {
+      setUncaughtExceptionHandler(builder.uncaughtExceptionHandler);
+    }
```
szetszwo, line 40: Use Optional:
```java
    Optional.ofNullable(builder.uncaughtExceptionHandler ).ifPresent(this::setUncaughtExceptionHandler);
```

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
```diff
@@ -17,24 +17,59 @@
...
   }
 
-  /** Construct a daemon thread with the given runnable. */
-  public Daemon(Runnable runnable) {
-    this(runnable, runnable.toString());
+  /** @return a {@link Builder}. */
+  // TODO(jiacheng): For exceptions that should crash the server, apply a correct UncaughtExceptionHandler
```
szetszwo, line 44: What does this TODO mean?

ratis-common/src/main/java/org/apache/ratis/util/JvmPauseMonitor.java
```diff
@@ -137,7 +137,8 @@ private void handle(TimeDuration extraSleep) {
 
   /** Start this monitor. */
   public void start() {
-    final MemoizedSupplier<Thread> supplier = JavaUtils.memoize(() -> new Daemon(this::run));
+    final MemoizedSupplier<Thread> supplier = JavaUtils.memoize(() ->
+        Daemon.newBuilder().setName("JvmPauseMonitor").setRunnable(this::run).build());
```
szetszwo, line 141: Add an id to the name.

szetszwo (commented):
@jiacheliu3 , thanks for the update!  This change becomes big and involves many files.  Could you split it into two pull requests?
1. Add a Daemon.Builder.
2. Support UncaughtExceptionHandler.

Some comments inlined but I have not checked all the changes.

jiacheliu3:
> @jiacheliu3 , thanks for the update! This change becomes big and involves many files. Could you split it into two pull requests?
> 
> 1. Add a Daemon.Builder.
> 2. Support UncaughtExceptionHandler.
> 
> Some comments inlined but I have not checked all the changes.

Sure will do. My biggest concern at this moment is the backward compatibility, could you shed some light on this thread? 
https://github.com/apache/ratis/pull/733#discussion_r967057138
Thx!

szetszwo:
UncaughtExceptionHandler is optional so that it is backward compatible.  Could you give more details of your concern?  For example, what will break by this change?

jiacheliu3:
> @jiacheliu3 , thanks for the update! This change becomes big and involves many files. Could you split it into two pull requests?
> 
> 1. Add a Daemon.Builder.
> 2. Support UncaughtExceptionHandler.
> 
> Some comments inlined but I have not checked all the changes.

A separate PR https://github.com/apache/ratis/pull/747 is created to segregate the `Daemon.Builder` part of changes. All relevant comments will be addressed in that PR and I'll click `Resolve comment` on this one after that. Sorry about the inconvenience in review, if any.

Then this PR will be rebased on #747 and only takes care of adding the `UncaughtExceptionHandler`. I will reuse this PR so all comment history is kept intact.

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
jiacheliu3: Moved to #747 and this TODO has been resolved.

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
jiacheliu3: I feel the uncaught exceptions should default to be printed so at least users can be aware of issues and fix them. Otherwise it's really hard to catch silent crashes. I figure logging is a mild action and doesn't change the existing behavior. WDYT?

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
jiacheliu3: As mentioned above, I figure logging as the default action doesn't break existing behavior and leaves a way for users to identify issues so they can be fixed later.

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
jiacheliu3: Ok i misunderstood the code here. I initially thought this is to invoke user extensions (i.e. a user may extend `ServerImplUtils.newRaftServer()` to init a `RaftServer` child class), so the `newRaftServer()` signature should be kept the same to have backward compatibility. 

Then after a while I realized `ServerImplUtils` is a `final` class so no extension is expected. I guess this reflection is used just to avoid importing `ratis-server` module (avoid circular dependency). 

Therefore no backward compatibility is needed here.

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -249,5 +262,11 @@ public Builder setParameters(Parameters parameters) {
       this.parameters = parameters;
       return this;
     }
+
+    // TODO(jiacheng): add UTs
+    public Builder setUncaughtExceptionHandler(Thread.UncaughtExceptionHandler exceptionHandler) {
```
jiacheliu3, line 267: Adding an optional field to `RaftServer.Builder` already gives backward compatibility. Existing usages don't need to change anything if they upgrade Ratis, it's just they will be using the default `UncaughtExceptionHandler`

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -95,9 +95,9 @@
 import static org.apache.ratis.util.LifeCycle.State.RUNNING;
 import static org.apache.ratis.util.LifeCycle.State.STARTING;
 
-class RaftServerImpl implements RaftServer.Division,
+public class RaftServerImpl implements RaftServer.Division,
```
jiacheliu3, line 98: have to add the `public` due to the cast in `LogAppenderDaemon`

ratis-server/src/main/java/org/apache/ratis/server/leader/LogAppenderDaemon.java
```diff
@@ -44,11 +46,17 @@ class LogAppenderDaemon {
 
   private final LogAppender logAppender;
 
-  LogAppenderDaemon(LogAppender logAppender) {
+  LogAppenderDaemon(LogAppender logAppender, RaftServer.Division server) {
```
jiacheliu3, line 49: I have to use `RaftServer.Division` here and cast to `RaftServerImpl` internally because the `server` ref is passed in from `GrpcFactory/ServerFactory` all the way down and `ServerFactory` cannot import `RaftServerImpl`(circular dependency).

jiacheliu3:
Not sure how much UT would be necessary. If necessary, I'll try writing something similar to `RaftStateMachineExceptionTests` using a `MiniRaftCluster` which creates a `RaftServer` with my user defined `UncaughtExceptionHandler`, and when I throw an exception internally, check if the handler is able to catch that.

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
```diff
@@ -49,6 +60,11 @@ public Builder setRunnable(Runnable runnable) {
       return this;
     }
 
+    public Builder setUncaughtExceptionHandler(UncaughtExceptionHandler exceptionHandler) {
```
jiacheliu3, line 63: Not sure how much UT would be necessary. If necessary, I'll try writing something similar to RaftStateMachineExceptionTests using a MiniRaftCluster which creates a RaftServer with my user defined UncaughtExceptionHandler, and when I throw an exception internally, check if the handler is able to catch that.

szetszwo (commented):
@jiacheliu3 , thanks for the update!  After checked the Java API in more details, it seems better to pass a `ThreadGroup` instead a `UncaughtExceptionHandler` (sorry that it was a bad idea); see https://issues.apache.org/jira/secure/attachment/13049347/733_review.patch

ratis-common/src/main/java/org/apache/ratis/util/Daemon.java
jiacheliu3: Sounds good. This has been removed.

jiacheliu3:
> @jiacheliu3 , thanks for the update! After checked the Java API in more details, it seems better to pass a `ThreadGroup` instead a `UncaughtExceptionHandler` (sorry that it was a bad idea); see https://issues.apache.org/jira/secure/attachment/13049347/733_review.patch

@szetszwo Thanks for the suggestion that makes sense to me. One subtle implication I'm not sure about is, if the `ThreadGroup` is left `null` when creating a thread, that thread uses the current thread's `ThreadGroup`: [code](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#Thread-java.lang.ThreadGroup-java.lang.Runnable-)

I'm not 100% sure what that suggests. Specifically I'm not sure if threads from `TimeoutScheduler.newExecutor()` [code](https://github.com/apache/ratis/blob/8fbe434efad64436e5505f5a03a4ef048a425a0a/ratis-common/src/main/java/org/apache/ratis/util/TimeoutScheduler.java#L89) will implicitly also inherit this `ThreadGroup` and uncaught exceptions there will propagate to the application in the same way. **Is that true** and is that desirable? In other words I imagine there are critical threads whose uncaught exceptions should crash the `RaftServer`, whereas there are unimportant threads whose crashes should be somehow tolerated and just generate a warning at most.

In contrast if we specify an `UncaughtExceptionHandler` in the old way, it is possible to distinguish the critical threads from the others. Not saying `setUncaughtExceptionHandler()` doesn't have downsides, just saying it's easier to manually configure.

ratis-server/src/main/java/org/apache/ratis/server/leader/LogAppenderDaemon.java
codings-dan: Is it possible to pass in a `ThreadGroup` here instead of passing in a `RaftServer.Division`

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
```diff
@@ -379,6 +383,10 @@ public LifeCycle.State getLifeCycleState() {
     return lifeCycle.getCurrentState();
   }
 
+  public ThreadGroup getThreadGroup() {
+    return this.threadGroup;
```
codings-dan, line 387: remove `this`

codings-dan (commented):
@jiacheliu3 Thanks a lot for working on this , overall lgtm, left some comments for ref

ratis-server/src/main/java/org/apache/ratis/server/leader/LogAppenderDaemon.java
jiacheliu3: That is possible but I will need to pass in the `ThreadGroup` from `RaftServerImpl` [here](https://github.com/apache/ratis/blob/8fbe434efad64436e5505f5a03a4ef048a425a0a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java#L245) and essentially modify the signature on [this interface](https://github.com/apache/ratis/blob/8fbe434efad64436e5505f5a03a4ef048a425a0a/ratis-server-api/src/main/java/org/apache/ratis/server/ServerFactory.java#L37). 

So which one do you prefer @codings-dan :
Option 1: Update the interface `ServerFactory.newLogAppender()` to pass in `ThreadGroup`. If this interface is not public-facing then I'll just do that.

Option 2: Do the cast here as the code currently shows.

codings-dan: Can we add a `getThreadGroup` method declaration to the `RaftServer` interface, and then this constructor can be directly changed to
```
Daemon.Builder builder = Daemon.newBuilder().setName(name).setRunnable(this::run)
        .setThreadGroup(logAppender.getServer().getThreadGroup());
```

Do you think this is okay?

jiacheliu3: @codings-dan Sounds good to me. I think a getter on the `RaftServer` interface would be totally fine and non-distruptive, with a `default` implementation too.

codings-dan: Yean, maybe a default implementation needs to be added as well.

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -116,6 +116,11 @@ default RaftGroup getGroup() {
     /** @return the internal {@link RaftClient} of this division. */
     RaftClient getRaftClient();
 
+    /** @return the {@link ThreadGroup} the threads of this Division belong to. */
+    default ThreadGroup getThreadGroup() {
+      return Thread.currentThread().getThreadGroup();
+    }
```
szetszwo, line 122: Remove the default -- it will never be used.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
szetszwo: Please remove `public`.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -197,6 +198,7 @@ public long[] getFollowerNextIndices() {
     this.lifeCycle = new LifeCycle(id);
     this.stateMachine = stateMachine;
     this.role = new RoleInfo(id);
+    this.threadGroup = proxy.getThreadGroup();
```
szetszwo, line 201: Create a child `ThreadGroup` for each impl:
```java
 this.threadGroup = new ThreadGroup(proxy.getThreadGroup(), getMemberId().toString());
```

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -274,6 +276,10 @@ TimeDuration getSleepDeviationThreshold() {
     return sleepDeviationThreshold;
   }
 
+  public ThreadGroup getThreadGroup() {
```
szetszwo, line 279: Add `@Override`.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
```diff
@@ -379,6 +383,10 @@ public LifeCycle.State getLifeCycleState() {
     return lifeCycle.getCurrentState();
   }
 
+  public ThreadGroup getThreadGroup() {
```
szetszwo, line 386: Remove `public`.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
```diff
@@ -196,9 +197,11 @@ String toString(RaftGroupId groupId, CompletableFuture<RaftServerImpl> f) {
   private final ExecutorService executor;
 
   private final JvmPauseMonitor pauseMonitor;
+  @Nullable
```
szetszwo, line 200: Let's create a default `ThreadGroup` and make it non-nullable.

ratis-server/src/main/java/org/apache/ratis/server/impl/ServerImplUtils.java
```diff
@@ -47,26 +47,28 @@ private ServerImplUtils() {
   /** Create a {@link RaftServerProxy}. */
   public static RaftServerProxy newRaftServer(
       RaftPeerId id, RaftGroup group, RaftStorage.StartupOption option, StateMachine.Registry stateMachineRegistry,
-      RaftProperties properties, Parameters parameters) throws IOException {
+      RaftProperties properties, Parameters parameters, ThreadGroup threadGroup)
```
szetszwo, line 50: Please reorder the parameters:
```java
ThreadGroup threadGroup, RaftProperties properties, Parameters parameters)
```

szetszwo (commented):
@jiacheliu3 , thanks for the update!  Some comments inlined.

ratis-server/src/main/java/org/apache/ratis/server/impl/ServerImplUtils.java
jiacheliu3: Rearranged together with all the places that have the same argument list

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
jiacheliu3: Actually I figure a `null` here is fine. [doc](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#Thread-java.lang.ThreadGroup-java.lang.Runnable-) specifies that is the `ThreadGroup` arg is null, the current thread's group is inherited. That behavior can be better than we guess a default here?

jiacheliu3: Interestingly... OpenJDK does not really respect a `null` ThreadGroup here and throws NPE on me. I will change this to non-null.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerProxy.java
```diff
@@ -221,6 +222,7 @@ String toString(RaftGroupId groupId, CompletableFuture<RaftServerImpl> f) {
     final TimeDuration leaderStepDownWaitTime = RaftServerConfigKeys.LeaderElection.leaderStepDownWaitTime(properties);
     this.pauseMonitor = new JvmPauseMonitor(id,
         extraSleep -> handleJvmPause(extraSleep, rpcSlownessTimeout, leaderStepDownWaitTime));
+    this.threadGroup = threadGroup == null ? new ThreadGroup("raft-server-proxy") : threadGroup;
```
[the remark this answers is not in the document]
jiacheliu3: Here I'm just applying a `ThreadGroup` that has no `unhandledException()`. In the future if it's proved to be needed I can submit another PR for a better default `ThreadGroup`.

szetszwo, line 225: Please use `id` for the name.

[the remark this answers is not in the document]
szetszwo: No default `uncaughtException(..)` probably is okay since it will use the system default.

ratis-server/src/test/java/org/apache/ratis/server/impl/MiniRaftCluster.java
```diff
@@ -390,7 +390,8 @@ private RaftServerProxy newRaftServer(RaftPeerId id, RaftGroup group, boolean fo
       RaftServerConfigKeys.setStorageDir(prop, Collections.singletonList(dir));
       return ServerImplUtils.newRaftServer(id, group,
           format? RaftStorage.StartupOption.FORMAT: RaftStorage.StartupOption.RECOVER,
-          getStateMachineRegistry(prop), prop, setPropertiesAndInitParameters(id, group, prop));
+          getStateMachineRegistry(prop),
+              new ThreadGroup("test-raft-server"), prop, setPropertiesAndInitParameters(id, group, prop));
```
szetszwo, line 394: Pass `null` so that it will use the default `ThreadGroup`.

szetszwo (commented):
@jiacheliu3 , thanks for the update!  Just some minor comments inlined.

jiacheliu3:
This test passes on my local PC, I believe this one is flaky.
```
Error:  Failures: 
Error:    TestLeaderElectionWithSimulatedRpc>LeaderElectionTests.testEnforceLeader:284->LeaderElectionTests.enforceLeader:298 expected:<s[1]> but was:<s[3]>
```

codings-dan:
Let me re-run the failed-job.

szetszwo (approved):
+1 the change looks good.

PR #746, opened 2022-09-14
RATIS-1708. RaftLog file channel closed before async flush ends

SzyWilliam:

When the RaftLog's FileChannel is closed, some async futures may still in pending state. Unexpected error would occur when these futures start executing on a closed file channel, see[ issue-1708](https://issues.apache.org/jira/browse/RATIS-1708). We should wait for these futures.

https://issues.apache.org/jira/browse/RATIS-1708

## How was this patch tested?

unit tests.


szetszwo (commented):
@SzyWilliam , thanks for working on this!  

1. We should move `SegmentedRaftLogWorker.flushFuture` to `BufferedWriteChannel`
2. Then, call `flushFuture.get().join()` in `BufferedWriteChannel .close()`.

Please see https://issues.apache.org/jira/secure/attachment/13049277/746_review.patch

szetszwo (approved):
+1 the change looks good.

PR #752, opened 2022-09-29
RATIS-1674. In GrpcLogAppender, disable retry and add minWait.

SzyWilliam:
1. Add minWait for GrpcLogAppender. When cluster contains a slow follower, 0 waiting time ends up high cpu load and lots of Inconsistent AppendEntries. 
2. Disable retry for Server RPC channel. There's leak in gRPC retry mechanism that could lead to OutOfDirectMemoryError when appendEntries timeout and retried on a slow follower. see https://github.com/grpc/grpc-java/issues/9563.

https://issues.apache.org/jira/projects/RATIS/issues/RATIS-1674

## How was this patch tested?
unit tests
manual tests


szetszwo (approved):
+1 the change looks good.

PR #766, opened 2022-10-19
RATIS-1726. Add execute permission to ratis shell script

maobaolong:

Add execute permission to ratis shell script, so that `bin/ratis` can be executed.

https://issues.apache.org/jira/browse/RATIS-1726

## How was this patch tested?

`bin/ratis` can display the help content for ratis script


szetszwo (approved):
+1 the change looks good.

PR #772, opened 2022-10-29
RATIS-1340: Ratis scripts' discovery of Bash source path is incompatible with CDPATH.

cnauroth:
…ble with CDPATH.

Several Bash scripts in Ratis use this common idiom to discover the current source path and then build relative paths from it:

```
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
This mostly works well, but it isn't compatible with interactive shells using the `CDPATH` environment variable to provide a custom search path for resolving the directory referenced by `cd`.  (See [bash man page](https://linux.die.net/man/1/bash) discussion of `CDPATH` and how the resolved directory is "written to the standard output.")

The standard solution is to redirect stdout of `cd` to `/dev/null`:

```
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" > /dev/null && pwd )"
```

Some Ratis scripts already do this, but not all.

https://issues.apache.org/jira/browse/RATIS-1340

## How was this patch tested?

Manual testing invoking the scripts.

### DIFF
--- a/ratis-examples/src/main/bin/client.sh
+++ b/ratis-examples/src/main/bin/client.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" > /dev/null && pwd )"
 
 USAGE="client.sh <example> <command>"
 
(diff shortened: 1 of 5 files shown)


szetszwo (approved):
+1 the change looks good.

adoroszlai (approved):

adoroszlai:
Thanks @cnauroth for the patch, @szetszwo for the review.

cnauroth:
Thank you, @adoroszlai and @szetszwo !

PR #786, opened 2022-11-17
RATIS-1748. Cache getMetricName to reduce garbage

lhotari:

- reduce unnecessary object allocations to reduce memory and garbage collection pressure

- https://issues.apache.org/jira/browse/RATIS-1748

## How was this patch tested?

Existing tests cover this


szetszwo (approved):
+1 the change looks good.

PR #797, opened 2022-12-08
RATIS-1758. Add linearizable read in Counter example

SzyWilliam:

Add linearizable read usage in Counter example.
Also fixed some problems in ratis-examples/README

https://issues.apache.org/jira/browse/RATIS-1758

## How was this patch tested?

manual test.
WX20221208-170241

### DIFF
--- a/ratis-examples/src/main/java/org/apache/ratis/examples/counter/server/CounterServer.java
+++ b/ratis-examples/src/main/java/org/apache/ratis/examples/counter/server/CounterServer.java
@@ -24,12 +24,14 @@ import org.apache.ratis.protocol.RaftPeer;
 import org.apache.ratis.server.RaftServer;
 import org.apache.ratis.server.RaftServerConfigKeys;
 import org.apache.ratis.util.NetUtils;
+import org.apache.ratis.util.TimeDuration;
 
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Optional;
 import java.util.Scanner;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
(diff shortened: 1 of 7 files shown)


ratis-examples/src/main/java/org/apache/ratis/examples/counter/server/CounterServer.java
```diff
@@ -52,6 +52,7 @@ public CounterServer(RaftPeer peer, File storageDir) throws IOException {
 
     //set the storage directory (different for each peer) in the RaftProperty object
     RaftServerConfigKeys.setStorageDir(properties, Collections.singletonList(storageDir));
+    RaftServerConfigKeys.Read.setOption(properties, RaftServerConfigKeys.Read.Option.LINEARIZABLE);
```
codings-dan, line 55: Please add a comment to explain what this parameter does

szetszwo:
@SzyWilliam , I got some idea to use the Counter example to test Linearizable Read:
- Add a configurable SIMULATED_SLOWNESS to the `CounterStateMachine` -- add a sleep time such that it will sleep in each `applyTransaction(..)` call.
- Setup a 3-node (n0, n1, n2) cluster as following
  - Set n0 to have a higher priority, so it will always be the leader.
  - Set n1 to be slow -- set SIMULATED_SLOWNESS to 1 second or higher.
- After the cluster started
  - run a client to update the leader 10 times.
  - after that, read from the n1 to see if it will wait (~10s) and get the latest value.

ratis-examples/src/main/java/org/apache/ratis/examples/counter/server/CounterServer.java
SzyWilliam: Done

SzyWilliam:
@szetszwo Thanks for your advices! I've made changes according to these steps and verified linearizable read, see the following screenshot.
WX20221209-171535
The original purpose of this PR is to simply demonstrate linearizable read usages. Hope that this patch will not make it too complicated for users to grasp.

ratis-examples/src/main/java/org/apache/ratis/examples/counter/server/CounterServer.java
```diff
@@ -80,35 +90,46 @@ public void close() throws IOException {
...
+        throw new IllegalArgumentException("Invalid argument number: expected to be 1 or 2 but actual is "
+            + args.length);
       }
       final int peerIndex = Integer.parseInt(args[0]);
       if (peerIndex < 0 || peerIndex > 2) {
         throw new IllegalArgumentException("The server index must be 0, 1 or 2: peerIndex=" + peerIndex);
       }
+      TimeDuration simulatedSlowness = TimeDuration.ZERO;
+      if (args.length == 2) {
+        final int slowness = Integer.parseInt(args[1]);
```
szetszwo, line 104: Use  `TimeDuration.valueOf(String timeString, TimeUnit defaultUnit)`.

ratis-examples/src/main/resources/conf.properties
```diff
@@ -15,4 +15,5 @@
 # limitations under the License.
 
 raft.server.address.list=127.0.0.1:10024,127.0.0.1:10124,127.0.0.1:11124
-# raft.server.root.storage.path
\ No newline at end of file
+# raft.server.root.storage.path
+# raft.server.priority.list=1,0,0
```
szetszwo, line 19: How about adding slowness list here instead of passing it in CLI?
```
# raft.server.simulated-slowness.list=0,1s,0
```

szetszwo (commented):
@SzyWilliam , the change looks great!  Adding `raft.server.priority.list` is good.  How about we also add `raft.server.simulated-slowness.list`?

szetszwo:
> ... I've made changes according to these steps and verified linearizable read, see the following screenshot. ...

Thanks a lot for trying it!

ratis-examples/src/main/java/org/apache/ratis/examples/counter/server/CounterServer.java
SzyWilliam: Done

ratis-examples/src/main/resources/conf.properties
SzyWilliam: It's a great idea! Saved me many nasty parsing code.

szetszwo (approved):
+1 the change looks good.

PR #803, opened 2023-01-05
RATIS-1765. [GrpcLogAppender] Calculate streaming md5 file-wise when installSnapshot

SzyWilliam:
Stream MD5 is introduced  by https://github.com/apache/ratis/commit/db3623a22e39a7957d9985fc9fe6784fe977f070 to reduce duplicate IO costs. However, current implementation calculates MD5 chunk-wise instead of file-wise.
We should calculate MD5 once per file, instead of many times per chunk.

https://issues.apache.org/jira/browse/RATIS-1765

## How was this patch tested?

1. unit tests.
2. new unit tests.


szetszwo (approved):
+1 the change looks good.

SzyWilliam:
@szetszwo Thanks for reviewing this PR!

PR #824, opened 2023-02-21
RATIS-1785. Use SingleThreadExecutor to manage the lifetime of single thread

OneSizeFitsQuorum:
see [jira1785](https://issues.apache.org/jira/browse/RATIS-1785)


OneSizeFitsQuorum:
@szetszwo PTAL

szetszwo (approved):
+1 the change looks good.

PR #845, opened 2023-03-07
RATIS-1770. Yield leader to higher priority peer by TransferLeadership

kaijchen:

Yield leader to higher priority peer by TransferLeadership.
To avoid blocking client requests for a long time,
it will only start TransferLeadership when the higher priority peer is up-to-date.

https://issues.apache.org/jira/browse/RATIS-1770

## How was this patch tested?

Covered by `LeaderElectionTests#testYieldLeaderToHigherPriority()`.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
@@ -26,8 +26,6 @@ import org.apache.ratis.proto.RaftProtos.LogEntryProto.LogEntryBodyCase;
 import org.apache.ratis.proto.RaftProtos.RaftPeerRole;
 import org.apache.ratis.proto.RaftProtos.ReplicationLevel;
 import org.apache.ratis.proto.RaftProtos.RoleInfoProto;
-import org.apache.ratis.proto.RaftProtos.StartLeaderElectionReplyProto;
-import org.apache.ratis.proto.RaftProtos.StartLeaderElectionRequestProto;
 import org.apache.ratis.protocol.Message;
 import org.apache.ratis.protocol.RaftClientReply;
 import org.apache.ratis.protocol.RaftClientRequest;
@@ -663,52 +661,7 @@ class LeaderStateImpl implements LeaderState {
     return pendingStepDown.submitAsync(request);
   }
 
-  private synchronized TransferLeadership.Result sendStartLeaderElection(RaftPeerId follower, TermIndex lastEntry) {
-    final TermIndex currLastEntry = getLastEntry();
-    if (ServerState.compareLog(currLastEntry, lastEntry) != 0) {
-      return new TransferLeadership.Result(TransferLeadership.Result.Type.LAST_ENTRY_CHANGED,
-          "leader lastEntry changed from " + lastEntry + " to " + currLastEntry);
-    }
-    LOG.info("{}: sendStartLeaderElection to follower {} on term {}, lastEntry={}",
-        this, follower, currentTerm, lastEntry);
-
-    final StartLeaderElectionRequestProto r = ServerProtoUtils.toStartLeaderElectionRequestProto(
-        server.getMemberId(), follower, lastEntry);
-    final CompletableFuture<StartLeaderElectionReplyProto> f = CompletableFuture.supplyAsync(() -> {
-      server.getLeaderElectionMetrics().onTransferLeadership();
-      try {
-        return server.getServerRpc().startLeaderElection(r);
-      } catch (IOException e) {
-        throw new CompletionException("Failed to sendStartLeaderElection to follower " + follower, e);
-      }
-    }, server.getServerExecutor()).whenComplete((reply, exception) -> {
-      if (reply != null) {
-        LOG.info("{}: Received startLeaderElection reply from {}: success? {}",
-            this, follower, reply.getServerReply().getSuccess());
-      } else if (exception != null) {
-        LOG.warn(this + ": Failed to startLeaderElection for " + follower, exception);
-      }
-    });
-
-    if (f.isCompletedExceptionally()) { // already failed
-      try {
-        f.join();
-      } catch (Throwable t) {
-        return new TransferLeadership.Result(t);
-      }
-    }
-    return TransferLeadership.Result.SUCCESS;
-  }
-
-  TransferLeadership.Result sendStartLeaderElection(FollowerInfo transferee, TermIndex leaderLastEntry) {
-    final TransferLeadership.Result result = isFollowerUpToDate(transferee, leaderLastEntry);
-    if (result != TransferLeadership.Result.SUCCESS) {
-      return result;
-    }
-    return sendStartLeaderElection(transferee.getId(), leaderLastEntry);
-  }
-
-  private static TransferLeadership.Result isFollowerUpToDate(FollowerInfo follower, TermIndex leaderLastEntry) {
+  static TransferLeadership.Result isFollowerUpToDate(FollowerInfo follower, TermIndex leaderLastEntry) {
     if (follower == null) {
       return TransferLeadership.Result.NULL_FOLLOWER;
     } else if (leaderLastEntry != null) {
@@ -1094,7 +1047,7 @@ class LeaderStateImpl implements LeaderState {
     final TermIndex leaderLastEntry = getLastEntry();
     final LogAppender appender = chooseUpToDateFollower(highestPriorityInfos, leaderLastEntry);
     if (appender != null) {
-      server.getTransferLeadership().start(this, appender, leaderLastEntry);
+      server.getTransferLeadership().start(appender, leaderLastEntry);
     }
   }
 
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
@@ -677,7 +677,7 @@ class LeaderStateImpl implements LeaderState {
 
   private static LogAppender chooseUpToDateFollower(List<LogAppender> followers, TermIndex leaderLastEntry) {
     for(LogAppender f : followers) {
-      if (isFollowerUpToDate(f.getFollower(), leaderLastEntry) == null) {
+      if (isFollowerUpToDate(f.getFollower(), leaderLastEntry) == TransferLeadership.Result.SUCCESS) {
         return f;
       }
     }
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
@@ -62,6 +62,7 @@ import org.apache.ratis.util.TimeDuration;
 import org.apache.ratis.util.Timestamp;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -424,6 +425,10 @@ class LeaderStateImpl implements LeaderState {
     return currentTerm;
   }
 
+  TermIndex getLastEntry() {
+    return server.getState().getLastEntry();
+  }
+
   @Override
   public boolean onFollowerTerm(FollowerInfo follower, long followerTerm) {
     if (isAttendingVote(follower) && followerTerm > getCurrentTerm()) {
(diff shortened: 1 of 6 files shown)


ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -684,16 +685,16 @@ private synchronized void sendStartLeaderElection(RaftPeerId follower, TermIndex
     });
   }
 
+  boolean isFollowerUpToDate(FollowerInfo followerInfo) {
+    final TermIndex leaderLastEntry = server.getState().getLastEntry();
```
szetszwo, line 689: It should not call getLastEntry() two times.

szetszwo:
@kaijchen , I reviewed this and the TransferLeadership code earlier and found some suggestions.  Will post some comments soon.

szetszwo (commented):
@kaijchen , just have posted some comments in https://issues.apache.org/jira/browse/RATIS-1770 .  Please take a look.

kaijchen:
@szetszwo based on your patch, I have added some fixes and improvements.

Should we change the default timeout of ratis shell in this PR also?

| TransferLeadership by | default timeout |
| --- | --- |
| client (ratis shell) | `3s` (hard-coded) |
| client (non ratis shell) | default request timeout |
| server (priority yield) | random election timeout |

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
```diff
@@ -93,49 +132,76 @@ boolean isSteppingDown() {
...
+    } else if (error.contains("not up-to-date")) {
+      LOG.info("{}: notifying LogAppender to send AppendEntries as transferee {} is not up-to-date: {}",
+          server.getMemberId(), transferee, error);
       appender.notifyLogAppender();
+      return null;
     }
+    return error;
+  }
+
+  TransferLeadershipRequest newRequest(RaftPeerId transfereeId) {
```
szetszwo, line 167: Now `newRequest` is used once.  Let's move the code to the `start(..)` method below.

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
```diff
@@ -93,49 +132,76 @@ boolean isSteppingDown() {
...
+    return start(leaderState, context);
+  }
+
+  private CompletableFuture<RaftClientReply> start(LeaderStateImpl leaderState, Context context) {
+    final TransferLeadershipRequest request = context.getRequest();
     final MemoizedSupplier<PendingRequest> supplier = JavaUtils.memoize(() -> new PendingRequest(request));
     final PendingRequest previous = pending.getAndUpdate(f -> f != null? f: supplier.get());
     if (previous != null) {
-      return createReplyFutureFromPreviousRequest(request, previous);
+      return request != null? createReplyFutureFromPreviousRequest(request, previous): null;
```
szetszwo, line 191: Sorry, this needs to be reverted since `request` won't be null after we added `Context`.

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
```diff
@@ -93,49 +132,76 @@ boolean isSteppingDown() {
...
     final FollowerInfo follower = appender.getFollower();
-    if (leaderState.sendStartLeaderElection(follower)) {
+    final String error = leaderState.sendStartLeaderElection(follower, context.getLeaderLastEntry());
+    if (error == null) {
       LOG.info("{}: sent StartLeaderElection to transferee {} immediately as it already has up-to-date log",
           server.getMemberId(), transferee);
-    } else {
-      LOG.info("{}: notifying LogAppender to send AppendEntries as transferee {} is not up-to-date",
-          server.getMemberId(), transferee);
+    } else if (error.contains("not up-to-date")) {
```
szetszwo, line 158: Let's add a `Result` class to avoid depending on error message.

szetszwo (commented):
@kaijchen , thanks for the update!  Please see the comments inlined and also https://issues.apache.org/jira/secure/attachment/13056384/845_review2.patch

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
```diff
@@ -93,49 +132,76 @@ boolean isSteppingDown() {
...
       appender.notifyLogAppender();
+      return null;
     }
+    return error;
+  }
+
+  TransferLeadershipRequest newRequest(RaftPeerId transfereeId) {
+    // TransferLeadership will block client request, so we don't want wait too long.
+    // If everything goes well, transferee should be elected within an election timeout.
+    final long timeout = server.getRandomElectionTimeout().toLong(TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
kaijchen: I just looked at `getRandomElectionTimeout()`, and timeout in range `[minRpcTimeoutMs, maxRpcTimeoutMs]` is currently used for election timeout.
What does `minRpcTimeoutMs` and `maxRpcTimeoutMs` mean? Are they only used for election, or also for rpc timeout as the name suggests?

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
kaijchen: There is another approach, I removed the overloaded `sendStartElection` (the non-synchronized one).
The caller should check `isFollowerUpToDate` before calling `sendStartElection`.

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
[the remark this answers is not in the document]
szetszwo: Yes, they are for server rpc, i.e. (appendEntries).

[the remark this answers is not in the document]
kaijchen: Does it mean, it is possible for a follower to start election as soon as an appendEntries times out?
Should we seperate these two timeouts (in the future)?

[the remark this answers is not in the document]
szetszwo: Sorry, I was not clear -- It is the timeout to for sending out heartbeat appendEntries.  (Not the timeout to wait for an appendEntries reply)

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
```diff
@@ -151,29 +150,29 @@ private String tryTransferLeadership(LeaderStateImpl leaderState, Context contex
       return "LogAppender for transferee " + transferee + " is null";
     }
     final FollowerInfo follower = appender.getFollower();
-    final String error = leaderState.sendStartLeaderElection(follower, context.getLeaderLastEntry());
-    if (error == null) {
-      LOG.info("{}: sent StartLeaderElection to transferee {} immediately as it already has up-to-date log",
+    if (follower == null) {
+      return "Follower for LogAppender" + appender + " is null";
```
szetszwo, line 154: `appender.getFollower()` never returns null.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -663,7 +663,7 @@ CompletableFuture<RaftClientReply> submitStepDownRequestAsync(TransferLeadership
     return pendingStepDown.submitAsync(request);
   }
 
-  private synchronized String sendStartLeaderElection(RaftPeerId follower, TermIndex lastEntry) {
+  synchronized String sendStartLeaderElection(RaftPeerId follower, TermIndex lastEntry) {
```
szetszwo, line 666: `synchronized` is incorrect since it does not change anything in `LeaderStateImpl`.

Actually, this method has nothing to do with `LeaderStateImpl`.  It should be moved to `RaftServerImpl` since it use many methods there.

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
```diff
@@ -134,12 +134,11 @@ boolean isSteppingDown() {
...
-      final String error = leaderState.sendStartLeaderElection(follower, leaderState.getLastEntry());
-      if (error == null) {
-        LOG.info("{}: sent StartLeaderElection to transferee {} after received AppendEntriesResponse",
-            server.getMemberId(), follower.getId());
-      }
+    if (getTransferee().filter(t -> t.equals(follower.getId())).isPresent()
+        && LeaderStateImpl.isFollowerUpToDate(follower, leaderState.getLastEntry())) {
+      LOG.info("{}: sending StartLeaderElection to transferee {} after received AppendEntriesResponse",
+          server.getMemberId(), follower.getId());
+      leaderState.sendStartLeaderElection(follower.getId(), leaderState.getLastEntry());
```
szetszwo, line 141: - It should reuse `leaderState.getLastEntry()`.

- Check return value before LOG
```java
    if (!getTransferee().filter(t -> t.equals(follower.getId())).isPresent()) {
      return;
    }
    final TermIndex lastEntry = leaderState.getLastEntry();
    if (LeaderStateImpl.isFollowerUpToDate(follower, lastEntry)) {
      final String error = sendStartLeaderElection(follower.getId(), lastEntry);
      if (error != null) {
        LOG.warn("{}: Failed to sendStartLeaderElection to transferee {}",
            server.getMemberId(), follower.getId());
      } else {
        LOG.info("{}: sent StartLeaderElection to transferee {} after received AppendEntriesResponse",
            server.getMemberId(), follower.getId());
      }
    }
```

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
```diff
@@ -151,29 +150,29 @@ private String tryTransferLeadership(LeaderStateImpl leaderState, Context contex
...
     final FollowerInfo follower = appender.getFollower();
-    final String error = leaderState.sendStartLeaderElection(follower, context.getLeaderLastEntry());
-    if (error == null) {
-      LOG.info("{}: sent StartLeaderElection to transferee {} immediately as it already has up-to-date log",
+    if (follower == null) {
+      return "Follower for LogAppender" + appender + " is null";
+    }
+    if (LeaderStateImpl.isFollowerUpToDate(follower, context.getLeaderLastEntry())) {
+      LOG.info("{}: sending StartLeaderElection to transferee {} immediately as it already has up-to-date log",
+          server.getMemberId(), transferee);
```
szetszwo: Let's move the LOG after `tryTransferLeadership(..)` returns.  Then, we can check the return value before log.

szetszwo (commented):
@kaijchen , thanks for the update.  Please see the comments inlined and also https://issues.apache.org/jira/secure/attachment/13056443/845_review3.patch

kaijchen:
Thanks @szetszwo for the review and suggestion.

Changes since your last review: https://github.com/apache/ratis/pull/845/files/060df2c530dcb73344f18e4db45586861ba39d81..23810a1308085d3db12ec8d9006b4a434cf9c5a1

Changes on top of patch-3:
https://github.com/apache/ratis/pull/845/files/8e150d503eca5fc5681cf6cd367fdfe726860fe4..23810a1308085d3db12ec8d9006b4a434cf9c5a1

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
kaijchen: OK, but tryTransferLeadership returns `null` doesn't mean we have sent StartLeaderElection immediately:

https://github.com/apache/ratis/pull/845/files/b19cccebaf2316fa9d16de21b77b4439f8514d33..d5d6b612c2015ade11e642041c1bdab77728f09a

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
kaijchen: Removed, this was added in https://github.com/apache/ratis/pull/845/commits/060df2c530dcb73344f18e4db45586861ba39d81 to match the old logic just in case.

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
kaijchen: Updated, I have swapped the two branches when checking `error`.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
kaijchen: Good catch. Thanks for the suggestion.

kaijchen:
Besides, since there are many refactors included in this PR, shall we change the title of this Jira?

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
szetszwo: @kaijchen , you are right.  How about adding a `Result` class?  Everything becomes clear and easy to understand.

kaijchen: > How about adding a Result class? Everything becomes clear and easy to understand.

OK, let me rebase new changes to https://issues.apache.org/jira/secure/attachment/13056384/845_review2.patch

kaijchen:
Rebased on  https://issues.apache.org/jira/secure/attachment/13056384/845_review2.patch with fixes and improvements.

ratis-server/src/main/java/org/apache/ratis/server/impl/TransferLeadership.java
```diff
@@ -92,50 +188,135 @@ boolean isSteppingDown() {
...
+
+  private Result sendStartLeaderElection(FollowerInfo follower, TermIndex lastEntry) {
+    final Result result = isFollowerUpToDate(follower, lastEntry);
+    if (result != Result.SUCCESS) {
+      return result;
+    }
+
+    final TermIndex currLastEntry = server.getState().getLastEntry();
+    if (ServerState.compareLog(currLastEntry, lastEntry) != 0) {
+      return new Result(Result.Type.LAST_ENTRY_CHANGED,
```
kaijchen, line 212: I'm wondering why we should abort TransferLeadership when `LAST_ENTRY_CHANGED`.
The `StartLeaderElection` request shoud be idempotent (by including sender's term).

Then, as long as follower is up-to-date, it is OK to send the request.
i.e. Only check `isFollowerUpToDate(follower, currLastEntry)`.

szetszwo (approved):
+1 the change looks good.

kaijchen:
@szetszwo Thanks a lot for your review and suggestions.

PR #849, opened 2023-03-08
RATIS-1811. Improve StreamObserver's log messages

SzyWilliam:
see https://issues.apache.org/jira/browse/RATIS-1811


szetszwo (approved):
+1 the change looks good.

(We may simply use op as below.)
```java
      GrpcUtil.warn(LOG, () -> getId() + ": " + op + " onError, lastRequest: " + getPreviousRequestString(), t);
```

SzyWilliam:
@szetszwo Thanks for the review! Changed to the simple version.

PR #851, opened 2023-03-09
RATIS-1813. Allow ratis-shell to run in any JDK 8+ version

kaijchen:

This is mainly for supporting Java 17, but any other Java 8+ version should be allowed.

https://issues.apache.org/jira/browse/RATIS-1813

## How was this patch tested?

 Before

```console
$ bin/ratis sh
Error: ratis-shell requires Java 8 or Java 11, currently Java 17.0.6 found. 
```

After (tested in Java 8 and Java 17)

```console
$ bin/ratis sh
[main] INFO org.reflections.Reflections - Reflections took 32 ms to scan 1 urls, producing 5 keys and 18 values
Usage: ratis sh [generic options]
	 [election [transfer] [stepDown] [pause] [resume]]         
	 [group [info] [list]]                                     
	 [peer [add] [remove] [setPriority]]                       
	 [snapshot [create]]                                       
```

### DIFF
--- a/ratis-shell/src/main/libexec/ratis-shell-config.sh
+++ b/ratis-shell/src/main/libexec/ratis-shell-config.sh
@@ -50,12 +50,12 @@ if [[ -z "${JAVA}" ]]; then
   fi
 fi
 
-# Check Java version == 1.8 or == 11
+# Check Java version == 1.8 or >= 8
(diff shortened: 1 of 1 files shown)


szetszwo (approved):
+1 the change looks good.

PR #857, opened 2023-03-20
RATIS-1817. Do not send StartLeaderElection when leaderLastEntry is null

kaijchen:

We should not send StartLeaderElection to the transferee when `leaderLastEntry` is null.
Because `leaderLastEntry` is expected to be non-null in `RaftServerImpl#startLeaderElection`:

```java
if (leaderLastEntry == null) {
  LOG.warn("{}: receive null leaderLastEntry which is unexpected", getMemberId());
  return ServerProtoUtils.toStartLeaderElectionReplyProto(leaderId, getMemberId(), false);
}
```

https://issues.apache.org/jira/browse/RATIS-1817

## How was this patch tested?

Existing tests in CI.


szetszwo (approved):
+1 the change looks good.

Indeed, `leaderLastEntry` should never be null, this is a protective check.

PR #859, opened 2023-03-20
RATIS-1819. Use Maven Wrapper in CI workflows

tisonkun:

Split from https://github.com/apache/ratis/pull/854. This is a follow-up for [RATIS-1795](https://issues.apache.org/jira/browse/RATIS-1795). By using Maven Wrapper in CI workflows, we build reproducibly as users/devs in local env.

RATIS-1819

## How was this patch tested?

Should pass existing CI checks.

cc @szetszwo


szetszwo (approved):
+1 the change looks good.

Tried running some of the scripts.  They worked well.

PR #860, opened 2023-03-21
RATIS-1662. Intermittent failure in testEnforceLeader

kaijchen:

Increase sleep time in `blockQueueAndSetDelay()` to fix intermittent failure in `testEnforceLeader()`

https://issues.apache.org/jira/browse/RATIS-1662

## How was this patch tested?

Repeat 400x in CI
Before: https://github.com/kaijchen/ratis/actions/runs/4477152298
After: https://github.com/kaijchen/ratis/actions/runs/4477155987


kaijchen:
There is 1 failure (1/400) in TestLeaderElectionWithGrpc though,
https://github.com/kaijchen/ratis/actions/runs/4477451715/jobs/7868996121#step:5:7719

Should we further increase the sleep time? Is there any downsides if the sleep time is too large?

PR #863, opened 2023-03-27
RATIS-1822. Disable first election on changeToFollower

SzyWilliam:

Currently, first election is disabled on `setLeader`. However, it does not cover all scenarios. For example, when a candidate receives `DISCOVERED_A_NEW_TERM` as the response to `PreVote`, it should also disable first election.

Therefore, I propose to disable first election on `changeToFollower`

https://issues.apache.org/jira/browse/RATIS-1822

## How was this patch tested?
Existing unit tests.


szetszwo (approved):
+1 the change looks good.

PR #878, opened 2023-04-27
RATIS-1838. Compare commit index with last included index when installSnapshot

SzyWilliam:

See https://issues.apache.org/jira/browse/RATIS-1838

## How was this patch tested?

Existing unit tests and manual tests


SzyWilliam:
Since the 2.5.1 release is incoming, I think we can include this simple PR in 2.5.1 for now. The ALREADY_INSTALLED solution and other more sophisticated tests can be performed later.

OneSizeFitsQuorum (approved):
LGTM~
For Raft, in extreme cases it is possible for lastIndex to be greater than lastSnapshotIndex, so this assertion is not reasonable

OneSizeFitsQuorum:
@szetszwo cc

szetszwo (approved):
+1 the change looks good.

PR #885, opened 2023-06-11
RATIS-1848. Simplify PeerMap inheritance

tisonkun:

`createProxyImpl` is implemented for Netty variant only while the gRPC variant pass the handle to constructor directly. Perhaps it's better to use the same flavor so that improving readability and avoid unexpected `UnsupportedOperationException`.

Chore work. If it's good to have and a JIRA ticket is required, I'll file one.

https://issues.apache.org/jira/browse/RATIS-1848

## How was this patch tested?

Compiled.


szetszwo (commented):
+1 the change looks good.  Could you file an Apache jira?

tisonkun:
JIRA ticket filed. cc @szetszwo @SzyWilliam

PR #887, opened 2023-06-20
RATIS-1850. When the stream server channelInactive, cleanup the request.

guohao-rosicky:

When the stream server side receives a disconnection, memory is cleared

https://issues.apache.org/jira/browse/RATIS-1850


guohao-rosicky:
@szetszwo PTAL, thanks.

PR #891, opened 2023-07-05
RATIS-1855 Fix some sonar code smell and bugs in ratis-server

OneSizeFitsQuorum:
see [jira1855](https://issues.apache.org/jira/browse/RATIS-1855)


OneSizeFitsQuorum:
For some reason, my commit in the source branch doesn't match the commit shown in this PR. This is the first time I've encountered this strange phenomenon, seems to be a github bug, maybe I need to submit a pr again

PR #899, opened 2023-08-07
RATIS-1867 To uniformly manage the timeout parameters for detecting gc

OneSizeFitsQuorum:
see [jira](https://issues.apache.org/jira/browse/RATIS-1867)


OneSizeFitsQuorum:
@szetszwo PTAL~

PR #902, opened 2023-08-12
RATIS-1870. Refactor hasMajority code during configuration changes

SzyWilliam:
See https://issues.apache.org/jira/browse/RATIS-1870.

The patch includes two changes:
1. Add considerations on OldConf when deciding whether leader is working as `singleton`.
2. Move `hasMajority` related code to RaftConfigurationImpl.


szetszwo (approved):
+1 the refactoring is very nice!  It simplifies the code.

PR #914, opened 2023-09-05
RATIS-1883. Next Index should be always larger than Match Index in Gr…

Brokenice0415:
See [https://issues.apache.org/jira/browse/RATIS-1883](https://issues.apache.org/jira/browse/RATIS-1883)

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -420,6 +420,7 @@ public class GrpcLogAppender extends LogAppenderBase {
           grpcServerMetrics.onRequestSuccess(getFollowerId().toString(), reply.getIsHearbeat());
           getLeaderState().onFollowerCommitIndex(getFollower(), reply.getFollowerCommit());
           if (getFollower().updateMatchIndex(reply.getMatchIndex())) {
+            increaseNextIndex(reply.getMatchIndex() + 1, reply.getResult());
             getLeaderState().onFollowerSuccessAppendEntries(getFollower());
           }
           break;
@@ -433,7 +434,7 @@ public class GrpcLogAppender extends LogAppenderBase {
         case INCONSISTENCY:
           grpcServerMetrics.onRequestInconsistency(getFollowerId().toString());
           LOG.warn("{}: received {} reply with nextIndex {}", this, reply.getResult(), reply.getNextIndex());
-          updateNextIndex(reply.getNextIndex());
+          updateNextIndex(Math.max(getFollower().getMatchIndex() + 1, reply.getNextIndex()));
           break;
         default:
           throw new IllegalStateException("Unexpected reply result: " + reply.getResult());
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -420,7 +420,7 @@ public class GrpcLogAppender extends LogAppenderBase {
           grpcServerMetrics.onRequestSuccess(getFollowerId().toString(), reply.getIsHearbeat());
           getLeaderState().onFollowerCommitIndex(getFollower(), reply.getFollowerCommit());
           if (getFollower().updateMatchIndex(reply.getMatchIndex())) {
-            increaseNextIndex(reply.getMatchIndex() + 1, reply.getResult());
+            getFollower().increaseNextIndex(reply.getMatchIndex() + 1);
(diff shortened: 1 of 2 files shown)


ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -420,6 +420,7 @@ private void onNextImpl(AppendEntriesReplyProto reply) {
           grpcServerMetrics.onRequestSuccess(getFollowerId().toString(), reply.getIsHearbeat());
           getLeaderState().onFollowerCommitIndex(getFollower(), reply.getFollowerCommit());
           if (getFollower().updateMatchIndex(reply.getMatchIndex())) {
+            increaseNextIndex(reply.getMatchIndex() + 1, reply.getResult());
```
SzyWilliam, line 423: `increaseNextIndex` is used to update nextIndex during snapshot installation. I guess `getFollower().increaseNextIndex()` is what you need instead.

SzyWilliam (commented):
@Brokenice0415 Thanks for your PR and welcome to Ratis Community! There are related test failures, please take a look.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
Brokenice0415: Thank you for your reminder. I have replaced it with the correct function.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -433,7 +434,7 @@ private void onNextImpl(AppendEntriesReplyProto reply) {
         case INCONSISTENCY:
           grpcServerMetrics.onRequestInconsistency(getFollowerId().toString());
           LOG.warn("{}: received {} reply with nextIndex {}", this, reply.getResult(), reply.getNextIndex());
-          updateNextIndex(reply.getNextIndex());
+          updateNextIndex(Math.max(getFollower().getMatchIndex() + 1, reply.getNextIndex()));
```
SzyWilliam, line 437: If a follower shutdowns, clear the logs and meta conf, and then restarts (just as what we do in `runTestBootstrapReconf`), the leader will still maintains the matchIndex (previously before shutdown) of this follower.  The changes here will cause the nextIndex never being correctly updated. I think we can tolerate some inconsistent AppendEntries in trade for fault tolerance and robustness. What do you think @szetszwo ?

SzyWilliam (commented):
There are still test failures, just made some investigations.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
szetszwo: > ... clears the logs and meta conf, ...

The follower cannot clear the existing log.  Otherwise, the committed entries can be removed.  As an example, see the description of RATIS-1677.   BTW, `runTestBootstrapReconf` is starting new peers but not removing existing logs.

> ... Match index in FollowerInfoImpl is only allowed to be changed monotonously.  ...

Once the log entries are matched, it cannot be changed to unmatched later on.  This is an assumption of the Raft algorithm.

szetszwo (approved):
+1 the change looks good.  However, there were some test failure.  Let me rerun the tests.

szetszwo:
> ... In testBootstrapReconf, we first start the new servers [here](https://github.com/apache/ratis/blob/88405b037ac9db4471603dfdd8ddc7c542f0229c/ratis-server/src/test/java/org/apache/ratis/server/impl/RaftReconfigurationBaseTest.java#L405), but then restart the same new servers again [here](https://github.com/apache/ratis/blob/88405b037ac9db4471603dfdd8ddc7c542f0229c/ratis-server/src/test/java/org/apache/ratis/server/impl/RaftReconfigurationBaseTest.java#L425). ...

Would it work if we don't format the new peer again?  Anyway, we should fix the test then.

Brokenice0415:
> @Brokenice0415 , Indeed, `TestRaftReconfigurationWithGrpc.testBootstrapReconf()` failed and the failure can be reproduced. Could you take a look?
> 
> BTW, thanks a lot for working on this!

I tried to skip formatting the new peer when test `TestRaftReconfigurationWithGrpc.testBootstrapReconf()` when testing locally, and the tests will pass successfully. I think the deliberate fomat is the exact reason for the test failure.

szetszwo:
image
Not sure why the other commits are included in the "File changed".

Brokenice0415:
> image Not sure why the other commits are included in the "File changed".

Opps... My mistake. I have fixed it now.

szetszwo (approved):
Thanks for the update!

+1 the change looks good.

PR #916, opened 2023-09-09
RATIS-1885. Incorrect clientWriteRequest timer usage

jojochuang:

Start clientWriteRequest timer as soon as the leader receives the client request.

https://issues.apache.org/jira/browse/RATIS-1885

## How was this patch tested?

Patched on a 2.x Ozone cluster and confirmed the metrics look sane.


szetszwo (approved):
+1 the change looks good.

PR #931, opened 2023-10-04
RATIS-1898. Netty should use EpollEventLoopGroup by default

ivandika3:

From Netty documentation (https://netty.io/wiki/native-transports.html), `EpollEventLoopGroup` generally should have better performance than `NioEventLoopGroup` and therefore should be used by default. The only reason to use `NioEventLoopGroup` is if the underlying OS does not support it (`EpollEventLoopGroup` is only supported in Linux).

From the one of the core developer of Netty (Norman Maurer): https://stackoverflow.com/questions/35568365/netty-epolleventloopgroup-vs-nioeventloopgroup-which-should-i-choose-on-centos

https://issues.apache.org/jira/browse/RATIS-1898

## How was this patch tested?

Existing tests (`NioEventLoopGroup` to `EpollEventLoopGroup` should not introduce regressions).

### DIFF
--- a/ratis-netty/src/main/java/org/apache/ratis/netty/NettyConfigKeys.java
+++ b/ratis-netty/src/main/java/org/apache/ratis/netty/NettyConfigKeys.java
@@ -48,6 +48,9 @@ public interface NettyConfigKeys {
     String PORT_KEY = PREFIX + ".port";
     int PORT_DEFAULT = 0;
 
+    String USE_EPOLL_KEY = PREFIX + ".use-epoll";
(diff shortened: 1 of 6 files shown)


ivandika3:
@szetszwo @guohao-rosicky Could you help take a look if you have time?

szetszwo (approved):
@ivandika3 , thanks a lot for working on this!

+1 the change looks good.

ivandika3:
Thank you for the review @szetszwo.

PR #936, opened 2023-10-14
RATIS-1905. Fix some sonar lint complains

SzyWilliam:
see https://issues.apache.org/jira/browse/RATIS-1905


OneSizeFitsQuorum (approved):
LGTM~
BTW, ratis also have some bug-level code smells from sonar, maybe we can fix them first!

SzyWilliam:
@OneSizeFitsQuorum Thanks for reviewing this!

> BTW, ratis also have some bug-level code smells from sonar, maybe we can fix them first!

That's right. Would you like to work on fixing them? ;)

OneSizeFitsQuorum:
> That's right. Would you like to work on fixing them? ;)

Yes, I'd love to try and fix those bug-level code smells, and it's also an opportunity for me to learn more about ratis code :)
However, to avoid duplication, I will continue to fix it after this PR is merged.

PR #941, opened 2023-10-18
RATIS-1911. Add MembershipManager example.

wojiaodoubao:

The 《CONSENSUS: BRIDGING THEORY AND PRACTICE》 has explain the JOINT consensus, which is implemented by ratis. But it didn't say how to configure the initial conf for new peers. I didn't find demonstration in ratis either.
 
Initial conf for new peers is easy to make mistakes and cause split brain or cluster unavailable. I think we can add an example to illustrate how to initiate new peers when performing membership change.

https://issues.apache.org/jira/browse/RATIS-1911

## How was this patch tested?

Manual tests. This example is an interactive console.
image

### DIFF
--- a/ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/CServer.java
+++ b/ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/CServer.java
@@ -84,7 +84,7 @@ public class CServer implements Closeable {
     return server.getPeer();
   }
 
-  private static final FileVisitor deleter = new SimpleFileVisitor<Path>() {
+  private static final FileVisitor DELETER = new SimpleFileVisitor<Path>() {
     @Override
     public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOException {
       Files.delete(dir);
@@ -101,7 +101,7 @@ public class CServer implements Closeable {
   @Override
   public void close() throws IOException {
     server.close();
-    Files.walkFileTree(storageDir.toPath(), deleter);
+    Files.walkFileTree(storageDir.toPath(), DELETER);
   }
 
   @Override
--- a/ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/CServer.java
+++ b/ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/CServer.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.ratis.examples.membership.server;
+
+import org.apache.ratis.RaftConfigKeys;
+import org.apache.ratis.conf.RaftProperties;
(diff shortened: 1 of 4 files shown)


ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/RaftCluster.java
```diff
@@ -0,0 +1,223 @@
...
+
+    Collection<CServer> oldPeers = members.values();
+    List<CServer> newPeers = new ArrayList<>();
+    List<CServer> peerToStart = new ArrayList<>();
+    List<CServer> peerToStop = new ArrayList<>();
+
+    for (Integer port : newPorts) {
+      CServer server = members.get(port);
+      if (server == null) {
+        RaftGroup group = newPeerGroup(port);
```
SzyWilliam, line 79: When adding new peers to the group, a more robust way would be
1. Start the peer with an *EMPTY* group configuration.
2. Call `setConfiguration` to update the membership info.  By the time `setConfiguration` returns, the group leader guarantees the membership change succeed (in an atomic way) and all new peers are caught up and can serve requests.

There are chances that the newly started peer will be shut down if we provide a full group configuration instead of an EMPTY group:
The new peer may start an election *before* the leader receives the `setConfiguration` call (this is especially common if you turn on the fast first election timeout). Under this situation, this peer will be shutdown by the leader since in leader's perspective, the peer does not belong to the group.

SzyWilliam (commented):
Thanks @wojiaodoubao for the patch and welcome to the Ratis!

The membership change is subtle and we could use a detailed example. There are some issues, please take a look.

ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/RaftCluster.java
wojiaodoubao: Hi @SzyWilliam, thanks for pointing out my mistake. I did miss the 'shut down' issue and seems misunderstand the empty conf. I used to believe 'empty conf' has a chance causing cluster unavailable. After your reminding I tried to reproduce the 'unavailable case' in ut, but failed. So empty conf is the right way.

Let me fix this.

wojiaodoubao: The 'empty conf' does have a chance stuck the cluster into 'infinity election'. I described it in https://issues.apache.org/jira/browse/RATIS-1912. 

But the possibility of 'infinity election' is much much less than 'newly started peer be shut down'. IMHO 'empty conf' is still the best solution. Let's use the 'empty conf' solution in the example.

ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/CServer.java
```diff
@@ -0,0 +1,122 @@
...
+    public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+      Files.delete(file);
+      return FileVisitResult.CONTINUE;
+    }
+  };
+
+  @Override
+  public void close() throws IOException {
+    server.close();
+    Files.walkFileTree(storageDir.toPath(), deleter);
```
SzyWilliam, line 104: We can use simply `FileUtils.deleteFully` here.

ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/RaftCluster.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    // Step 2: update membership.
+    System.out.println("Update membership ...... Step 2: update membership from C_old to C_new.");
+    System.out.println(peersInfo(oldPeers, "C_old"));
+    System.out.println(peersInfo(newPeers, "C_new"));
+    if (members.size() > 0) {
+      RaftClient client = createClient();
+      try {
+        RaftClientReply reply = client.admin().setConfiguration(newPeers.stream()
```
SzyWilliam, line 107: `RaftClient` is also `Closeable`, so we can use try-with-resources here.

SzyWilliam (approved):
+1 the change looks good overall to me. Just some small issues, PTAL.

szetszwo (approved):
@wojiaodoubao , thanks a lots for adding this wonderful example!

+1 the change looks good.

wojiaodoubao:
Thanks @szetszwo @SzyWilliam a lot for your kindly help :)

PR #951, opened 2023-10-24
RATIS-1919 Fix some sonar code smell and bugs in ratis-client/common/grpc

OneSizeFitsQuorum:
see [jira](https://issues.apache.org/jira/browse/RATIS-1919)

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
@@ -133,6 +133,7 @@ public final class RaftClientImpl implements RaftClient {
   private final RaftGroupId groupId;
   private final RetryPolicy retryPolicy;
 
+  @SuppressWarnings("java:S3077")
   private volatile RaftPeerId leaderId;
 
   private final TimeoutExecutor scheduler = TimeoutExecutor.getInstance();
--- a/ratis-common/src/main/java/org/apache/ratis/util/AutoCloseableLock.java
+++ b/ratis-common/src/main/java/org/apache/ratis/util/AutoCloseableLock.java
@@ -39,6 +39,7 @@ public final class AutoCloseableLock implements AutoCloseable {
     return acquire(lock, null);
   }
 
+  @SuppressWarnings("java:S2222")
   public static AutoCloseableLock acquire(final Lock lock, Runnable preUnlock) {
     lock.lock();
     return new AutoCloseableLock(lock, preUnlock);
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/BlockingImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/BlockingImpl.java
@@ -132,6 +132,7 @@ class BlockingImpl implements BlockingApi {
       try {
         sleepTime.sleep();
       } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
         throw new InterruptedIOException("retry policy=" + retryPolicy);
       }
     }
--- a/ratis-common/src/main/java/org/apache/ratis/util/AwaitToRun.java
+++ b/ratis-common/src/main/java/org/apache/ratis/util/AwaitToRun.java
@@ -42,7 +42,7 @@ public class AwaitToRun implements AutoCloseable {
 
     @Override
     public void run() {
-      for (; ; ) {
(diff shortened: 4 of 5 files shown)


ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
```diff
@@ -133,6 +133,7 @@ void set(Collection<RaftPeer> newPeers) {
   private final RaftGroupId groupId;
   private final RetryPolicy retryPolicy;
 
+  @SuppressWarnings("java:S3077")
```
szetszwo, line 136: It seem that we may use @Immutable to avoid it according to https://community.sonarsource.com/t/rule-s3077-frequently-generates-a-false-positive/27479/14
```java
+++ b/ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeerId.java
@@ -22,6 +22,7 @@ import org.apache.ratis.thirdparty.com.google.protobuf.ByteString;
 
+import javax.annotation.concurrent.Immutable;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Objects;
@@ -33,6 +34,7 @@ import java.util.function.Supplier;
  * <p>
  * This is a value-based class.
  */
+@Immutable
 public final class RaftPeerId {
```

ratis-common/src/main/java/org/apache/ratis/util/AutoCloseableLock.java
```diff
@@ -39,6 +39,7 @@ public static AutoCloseableLock acquire(final Lock lock) {
     return acquire(lock, null);
   }
 
+  @SuppressWarnings("java:S2222")
```
szetszwo, line 42: Please add a comment:
```java
@SuppressWarnings("java:S2222") // Locks should be released 
```

szetszwo (commented):
@OneSizeFitsQuorum , thanks for working on this!  Just two minor comments inlined.

ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
OneSizeFitsQuorum: Good idea!

ratis-common/src/main/java/org/apache/ratis/util/AutoCloseableLock.java
OneSizeFitsQuorum: Fixed~

ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.apache.ratis.client.impl;
 
+import jdk.nashorn.internal.ir.annotations.Immutable;
```
szetszwo, line 20: @OneSizeFitsQuorum , Please use `javax.annotation.concurrent.Immutable`.

OneSizeFitsQuorum: It looks like this will fail to compile.
image
image

szetszwo: @OneSizeFitsQuorum , sorry that I was not clear.  The tag is for the class as below:
```java
+++ b/ratis-common/src/main/java/org/apache/ratis/protocol/RaftPeerId.java
@@ -22,6 +22,7 @@ import org.apache.ratis.thirdparty.com.google.protobuf.ByteString;
 
+import javax.annotation.concurrent.Immutable;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Objects;
@@ -33,6 +34,7 @@ import java.util.function.Supplier;
  * <p>
  * This is a value-based class.
  */
+@Immutable
 public final class RaftPeerId {
   private static final Map<ByteString, RaftPeerId> BYTE_STRING_MAP = new ConcurrentHashMap<>();
   private static final Map<String, RaftPeerId> STRING_MAP = new ConcurrentHashMap<>();
```

OneSizeFitsQuorum: Sorry, thank you for your patient guidance ~ now it is no problem! I learned something new!

szetszwo (approved):
+1 the change looks good.

PR #970, opened 2023-11-21
RATIS-1940. Introduce Pluggable Metrics to readme 

SzyWilliam:
see https://issues.apache.org/jira/browse/RATIS-1940


szetszwo (approved):
+1 the change looks good.

PR #983, opened 2023-12-09
RATIS-1960. Follower may be incorrectly marked as having caught up

adoroszlai:

1. Fix race condition that may cause follower to be incorrectly marked as having caught up with leader.
2. Fix calculation of `initializing` flag for `appendEntries` request sent to follower.

Please see RATIS-1960 for details on both.

## How was this patch tested?

CI:
https://github.com/adoroszlai/ratis/actions/runs/7151958459

Passed 10x50 `TestInstallSnapshotNotificationWithGrpc#testInstallSnapshotDuringBootstrap` (this was failing in 6% of runs after fixing only the race condition):
https://github.com/adoroszlai/ratis/actions/runs/7152037092


szetszwo (approved):
+1 the change looks good.

Good catch on both bugs!

adoroszlai:
Thansk @szetszwo for reviewing and merging this.

SzyWilliam:
Cool catch!

PR #991, opened 2023-12-13
RATIS-1649. Fix SF_SWITCH_NO_DEFAULT in PureJavaCrc32C.

nandakumar131:

Fixed SF_SWITCH_NO_DEFAULT findbugs issue in `PureJavaCrc32C`

RATIS-1649

## How was this patch tested?
Manually ran findbugs to make sure that the warning is gone after removing the `SuppressFBWarnings`


szetszwo (approved):
+1 the change looks good.

PR #999, opened 2023-12-29
RATIS-1984. Simplify Maven cache for dependencies in CI

adoroszlai:

 * Use explicit `cache/restore` in all checks except `build`.  This avoids the need for separate cache keys in concurrent checks, since only `build` will save cache.
 * Specify `path` exception to avoid the need to explicitly delete Ratis artifacts before cache upload.
 * Add step name where missing.

https://issues.apache.org/jira/browse/RATIS-1984

## How was this patch tested?

CI:
https://github.com/adoroszlai/ratis/actions/runs/7357421711

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -25,10 +25,11 @@ jobs:
       - name: Cache for maven dependencies
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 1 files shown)


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #1000, opened 2023-12-30
RATIS-1985. Calculate code coverage

adoroszlai:

* Setup Java 8 in `unit` check to match the version used in `build` check.
* Collect code coverage data from `unit` check.
* Rename `sonar` check to `coverage`.
* Combine coverage data from `unit` check splits.
* Generate HTML and XML report from combined coverage data.

Sonar Cloud, which currently shows 0% coverage, will hopefully pick up the same data.  (We may need an additional hack [like this](https://github.com/apache/ozone/blob/58572aaa736bb4c9ef859020c2a4726c7f1de97a/hadoop-ozone/dev-support/checks/sonar.sh#L26-L31).  But first let's try without that one.)

https://issues.apache.org/jira/browse/RATIS-1985

## How was this patch tested?

Temporarily enabled `coverage` check in forks.  Downloaded `coverage` artifact, checked HTML report.

https://github.com/adoroszlai/ratis/actions/runs/7364058733

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -16,6 +16,8 @@ name: build-branch
 on:
   - push
   - pull_request
+env:
+  WITH_COVERAGE: true
 jobs:
   build:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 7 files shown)


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #1004, opened 2024-01-07
RATIS-1991. Optimize the reconfig judgment logic in the ReconfigurationBase class.

huangzhaobo99:
Change the judgment logic: 
It is necessary to determine whether the `effective` is null, rather than `newValue`.

Because the parameter of the `properties.set` method is `effective`.

```java
final String effective = reconfigureProperty(property, newValue);
LOG.info("{}: changed property {} to {} (effective {})", name, property, newValue, effective);
if (newValue != null) {
  properties.set(property, effective);
} else {
  properties.unset(property);
}
```

https://issues.apache.org/jira/browse/RATIS-1991

## How was this patch tested?
No testing required.

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/conf/ReconfigurationBase.java
+++ b/ratis-common/src/main/java/org/apache/ratis/conf/ReconfigurationBase.java
@@ -172,7 +172,7 @@ public abstract class ReconfigurationBase implements Reconfigurable {
     }
     final String effective = reconfigureProperty(property, newValue);
     LOG.info("{}: changed property {} to {} (effective {})", name, property, newValue, effective);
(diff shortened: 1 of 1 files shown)


huangzhaobo99:
Hi @szetszwo, Could you kindly help to take a review when you have time, Thanks.

szetszwo (commented):
@huangzhaobo99 , thanks a lot for working on this!   I agree that we should check `effective`.  Please see the comments inlined.

huangzhaobo99:
Thanks @szetszwo your nice comments, have fixed.

szetszwo (approved):
+1 the change looks good.

PR #1008, opened 2024-01-10
RATIS-1992. Update website with Ratis 3.0.1 release

adoroszlai:

Regenerate Ratis website from latest sources to publish the post about 3.0.1.

Please double-check `post/3.0.0.html`: its content is slightly different from the version currently published, due to how the source was updated in 36317bdb.  Is it OK, or should we first update the source to more closely match the current website?

https://issues.apache.org/jira/browse/RATIS-1992

## How was this patch tested?

```
open index.html
open downloads.html
open post/3.0.1.html
```

### DIFF
--- a/categories.html
+++ b/categories.html
@@ -115,7 +115,7 @@
     <div class="container">
 
         <div class="col-md-12 trademark">
-            <p>&copy; 2023 <a href="http://apache.org">The Apache Software Foundation</a>,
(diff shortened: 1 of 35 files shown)


szetszwo (approved):
+1 the change looks good.

SzyWilliam:
> or should we first update the source to more closely match the current website?

The source contains all the features, changes and bug fixes, how about just use this source?

adoroszlai:
Thanks @szetszwo, @SzyWilliam for the review.

PR #1012, opened 2024-01-15
RATIS-1989. Intermittent timeout in TestStreamObserverWithTimeout

adoroszlai:

Fix intermittent timeout in `TestStreamObserverWithTimeout#testWithTimeout`.

1. The first request may take long to reach the server, probably due to some initialization overhead.  The patch adds a "warmup" request, for which timeout is not applied.
2. Client and server seem to share `grpc-default-executor` thread pool.  This may result in delay at client due to server sleeping.  The patch uses a thread per request in `GrpcTestServer`.
3. With these, we can restore test settings prior to 29d032dde65cfb441e7c912b1621cfff6c266674.

https://issues.apache.org/jira/browse/RATIS-1989

## How was this patch tested?

Passed 500x in [repeated run](https://github.com/adoroszlai/ratis/actions/runs/7528357858) (although it [also passed](https://github.com/adoroszlai/ratis/actions/runs/7529250422) 500x on `master`).

Since it's passing more frequently in regular CI, repeated full `unit (grpc)` 10 times.  `TestStreamObserverWithTimeout` failed 4 times [on `master`](https://github.com/adoroszlai/ratis/actions/runs/7529643511), passed in all 10 runs [with the patch](https://github.com/adoroszlai/ratis/actions/runs/7529628120).

Also passed 500x locally.

Regular CI:
https://github.com/adoroszlai/ratis/actions/runs/7528350705


szetszwo (approved):
+1 the change looks good.  Thanks,  a lot for fixing this!

adoroszlai:

PR #1016, opened 2024-01-17
RATIS-2003. Fix IT_NO_SUCH_ELEMENT in InstallSnapshotRequests

adoroszlai:

Remove `@SuppressFBWarnings("IT_NO_SUCH_ELEMENT")` in `InstallSnapshotRequests`, and ensure spotbugs does not complain.

https://issues.apache.org/jira/browse/RATIS-2003

## How was this patch tested?

CI:
https://github.com/adoroszlai/ratis/actions/runs/7562098256

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/leader/InstallSnapshotRequests.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/leader/InstallSnapshotRequests.java
@@ -62,13 +62,6 @@ class InstallSnapshotRequests implements Iterable<InstallSnapshotRequestProto> {
   /** The total number of snapshot files. */
   private final int numFiles;
 
-  /** The index of the current request. */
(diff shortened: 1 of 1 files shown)


szetszwo (commented):
@adoroszlai Thanks a lots for working on this!

Let's take this chance to move the iterator code to an inner class; see https://issues.apache.org/jira/secure/attachment/13066117/1016_review.patch

adoroszlai:
> Let's take this chance to move the iterator code to an inner class

Thanks @szetszwo for the review.  I agree that extracting to inner class improves the code.

There is one piece from the review patch that I omitted: extracting the `FileChunkProto readFileChunk(FileInfo info)` method.  The rest of the refactoring is straightforward, but I don't think I understand the reason `fileIndex++` can be incremented outside the `if (chunk.getDone())` block.

PR #1054, opened 2024-03-22
RATIS-2047 Avoid unnecessary warn log when creating raft group

OneSizeFitsQuorum:
see [jira](https://issues.apache.org/jira/browse/RATIS-2047)


OneSizeFitsQuorum:
@szetszwo PTAL

szetszwo (approved):
+1 the change looks good.

PR #1063, opened 2024-04-09
RATIS-2053. Notify term index change when configuration changes

SzyWilliam:

Statemachine will not update LastApplied TermIndex when handling configuration change logs. This will lead problems for linearizable reads here https://github.com/apache/ratis/blob/bc6221b32fff9022cebb02bb243aea9fff35e290/ratis-server/src/main/java/org/apache/ratis/server/impl/ReadRequests.java#L102-L105.

Linearizable read will compare the `apply index` (of state machine instead of RaftServer) with `read index`. So we need to update the state machines' `apply index` when applying configuration logs.

https://issues.apache.org/jira/browse/RATIS-2053

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/statemachine/impl/BaseStateMachine.java
+++ b/ratis-server/src/main/java/org/apache/ratis/statemachine/impl/BaseStateMachine.java
@@ -18,7 +18,7 @@
 
 package org.apache.ratis.statemachine.impl;
 
-import org.apache.ratis.proto.RaftProtos;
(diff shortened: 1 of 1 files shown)


szetszwo (approved):
+1 the change looks good.

SzyWilliam:
@szetszwo Thanks for reviewing this! I found that RaftServer would update TermIndex when applying non-statemachine logs here https://github.com/apache/ratis/blob/8638a2ffa802de576aae546c22160da3d9339a8f/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java#L1809. 
So this patch is duplicate. Let me close this PR.

PR #1067, opened 2024-04-17
RATIS-2059. Missing reference count when putting log entries to cache on follower

duongkame:

See [RATIS-2059](https://issues.apache.org/jira/browse/RATIS-2059).

https://issues.apache.org/jira/browse/RATIS-2059

## How was this patch tested?

CI.


szetszwo (approved):
+1 the change looks good.

PR #1075, opened 2024-05-03
RATIS-2069. RaftMetaConf command sets incorrect peerAddress pattern.

sadanand48:
[RATIS-2040](https://issues.apache.org/jira/browse/RATIS-2040) made some changes in the command to add an option to provide peerID. While doing so , it updated the address string to be of pattern "hostname/ip:port" which was previously hostname:port. This PR sets it back to "host:port" pattern.

This causes connectionRefused messages in the application as the wrong address string is used.

https://issues.apache.org/jira/browse/RATIS-2069

## How was this patch tested?
Unit test added


szetszwo (approved):
+1 the change looks good.

PR #1085, opened 2024-05-09
[RATIS-2082] RaftPeers equal should also check address

AngersZhuuuu:

We have two raft cluster,

A cluster IP is

A1:7082,A2:7082,A3:7082

B cluster master ip is 

B1:7082,B2:7082,B3:7082

When we start A cluster we use a wrong config 

A1:7082,A2:7082,B3:7082

After we start the cluster we found we use wrong config.

But this info has been write to raft-meta.conf.

Then after we restart A cluster, when client connect to A cluster.

It connect to A cluster follower then return the cached leader address, the leader peer id is 3, but in the meta conf, the peer rpc link is B3:7082.

Then cause it connect to cluster B.

Its caused by RaftPeer object equal only compare peerId cause it won't refresh raft-meta.conf
image

https://issues.apache.org/jira/browse/RATIS-2082

## How was this patch tested?


AngersZhuuuu:
ping @codings-dan @szetszwo

szetszwo (commented):
`RaftPeer` is a public API so changing it becomes incompatible.

Let's fix `RaftMetaConfCommand` instead?

AngersZhuuuu:
The problem should be here?
截屏2024-05-10 09 52 08

AngersZhuuuu:
I think https://issues.apache.org/jira/browse/RATIS-1860 meets the same issue and can fix the issue by ratis-shell.

PR #1088, opened 2024-05-10
RATIS-2086. Autolink Ozone issues in PRs

adoroszlai:

Ratis PR descriptions sometimes reference Ozone Jira issues. It would be nice to automatically link these, similar to how it is already done for RATIS.

https://issues.apache.org/jira/browse/RATIS-2086

## How was this patch tested?

Ozone has similar config:
https://github.com/apache/ozone/blob/6487de7defc87f6456e986100de07620f88b1015/.asf.yaml#L25-L29

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -24,7 +24,9 @@ github:
     squash:  true
     merge:   false
     rebase:  false
-  autolink_jira: RATIS
+  autolink_jira:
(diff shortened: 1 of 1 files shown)


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #1102, opened 2024-05-24
RATIS-2100. The `closeFuture` never completed while closing from the `NEW` state

chungen0126:

Currently, the closeFuture only completes after the LogAppenderDaemon has started. However, when closing from the NEW state, the transition is NEW -> CLOSED, and the LogAppenderDaemon was not started.

https://issues.apache.org/jira/browse/RATIS-2100


szetszwo (approved):
+1 the change looks good.

PR #1104, opened 2024-05-25
RATIS-2102. AsyncApi#send() is not handling retry and reply correctly for replication levels higher than MAJORITY

smengcl:

Background: [RATIS-1994](https://issues.apache.org/jira/browse/RATIS-1994) adds a feature allowing `AsyncApi#send()` to return only when a specified replication level is reached. This was done by adding `waitForReplication()` in RaftServer.

However, two bugs are found with [RATIS-1994](https://issues.apache.org/jira/browse/RATIS-1994) when the (write) replication level is set to higher than `MAJORITY` (the default):

1. When the request is retried (e.g. when timed out), the future retrieved from cache is not correctly waited on with `waitForReplication()` again. This causes divergence in the desired behavior depending on if the request is retried or not. This is fixed by invoking waitForReplication() in `writeAsync()` rather than `appendTransaction()`.

2. `waitForReplication()`'s watch request reply was incorrectly ignored, this is fixed by combining the reply from write and watch together using `combineReplies()`.

Thanks @szetszwo for the help during debugging and making the patch.

https://issues.apache.org/jira/browse/RATIS-2102

## How was this patch tested?

- Tested manually during HDDS-10108 dev. Maybe it is a good idea to add some test cases around this in Ratis.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -832,20 +832,14 @@ class RaftServerImpl implements RaftServer.Division,
       leaderState.notifySenders();
     }
 
-    final CompletableFuture<RaftClientReply> future = pending.getFuture();
-    if (request.is(TypeCase.WRITE)) {
-      // check replication
-      final ReplicationLevel replication = request.getType().getWrite().getReplication();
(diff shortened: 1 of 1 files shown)


szetszwo (approved):
+1 the change looks good.

smengcl:
Thanks @szetszwo for reviewing this!

PR #1105, opened 2024-05-27
RATIS-2089. Add CommitInfoProto in NotReplicatedException

ivandika3:

In Ozone's XceiverClientRatis#watchForCommit, there are two watch commits request with different ReplicationLevel

1. Watch for ALL_COMMITTED 
2. Watch for MAJORITY_COMMITTED (If the previous watch threw an exception)

Based on the second watch request, the client will remove some failed datanode UUID from the commitInfoMap.

The second watch might not be necessary since the entries in AbstractCommitWatcher.commitIndexMap implies that the PutBlock request has been committed to the majority of the servers. Therefore, another MAJORITY_COMMITTED watch might not be necessary. From my understanding, the second MAJORITY_COMMITTED only serves to gain information to remove entries from commitInfoMap.

If the first watch failed with NotReplicatedException, we might be able to remove the need to a second watch request. Since NotReplicatedException is a Raft server exception, we can include the CommitInfoProtos in the NotReplicatedException. The client can use this CommitInfoProtos to remove the entry from commitInfoMap without sending another WATCH request. 

This CommitInfoProto is returned for every RaftClientReply (RaftClientReply.commitInfos), but if there is an exception, it seems the RaftClientReply is not accessible to the client.

However, if the exception is a client exception (e.g. due to Raft client watch timeout configuration), the client might have no choice but to send another watch request.

So in this patch, I propose to include CommitInfoProto into NotReplicatedException. 

This only introduces a client-side change by reusing the commitInfos in RaftClientReply.

https://issues.apache.org/jira/browse/RATIS-2089

## How was this patch tested?

Add a unit test to ensure commitInfoProto is not null during NotReplicateException.

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/ClientProtoUtils.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/ClientProtoUtils.java
@@ -398,7 +398,7 @@ public interface ClientProtoUtils {
       e = new NotLeaderException(serverMemberId, suggestedLeader, peers);
     } else if (replyProto.getExceptionDetailsCase() == NOTREPLICATEDEXCEPTION) {
       final NotReplicatedExceptionProto nre = replyProto.getNotReplicatedException();
-      e = new NotReplicatedException(nre.getCallId(), nre.getReplication(), nre.getLogIndex());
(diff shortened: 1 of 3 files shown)


ivandika3:
cc: @smengcl @szetszwo

szetszwo (approved):
+1 the change looks good.

smengcl:
Thanks a lot @ivandika3 for working on this one!

PR #1109, opened 2024-06-05
RATIS-2109. RaftLogBase#updateCommitIndex should return true only if commitIndex is increased

ivandika3:

RaftLogBase#updateCommitIndex returns true regardless whether commit index is actually increased (RaftLogIndex#updateIncreasingly returns true). This can cause unnecessary StateMachineUpdater notification

We can change it so that RaftLogBase#updateCommitIndex only returns true if commit index is actually increased. 

Note: For leader ServerState#updateCommitIndex is called only if there are entries to commit (i.e. majority > oldLasCommitted), so RaftLogBase#updateCommitIndex should increase the commit index of leader.

https://issues.apache.org/jira/browse/RATIS-2109

## How was this patch tested?

Existing tests.


szetszwo (approved):
+1 the change looks good.

PR #1113, opened 2024-06-20
RATIS-2114. Corruption due to SegmentedRaftLogWorker queues LogEntry without reference counter

duongkame:

`SegmentedRaftLogWorker` can queue LogEntries without retaining a reference counter. This is caught during testing zero-copy with Apache Ozone. Datanode crashes to corruption caused by SegmentedRaftLogWorker threads.

https://issues.apache.org/jira/browse/RATIS-2114

## How was this patch tested?

CI.


duongkame:
c.c @szetszwo

szetszwo (approved):
+1 the change looks good.

PR #1115, opened 2024-06-25
RATIS-2117. No need for manual assembly:single execution

adoroszlai:

RATIS-1660 added explicit executions of `maven-assembly-plugin` to build binary and source tarballs.  Since they are bound to the `package` phase, we no longer need to manually include `assembly:single` in the command to build Ratis.

https://issues.apache.org/jira/browse/RATIS-2117

## How was this patch tested?

Verified that both source and binary tarballs are still created:

https://github.com/adoroszlai/ratis/actions/runs/9660881857/job/26647380234#step:5:1950
https://github.com/adoroszlai/ratis/actions/runs/9660881857/job/26647380234#step:5:1964

Additional validation by `compile` check, which builds Ratis using contents from the source tarball.

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -44,7 +44,7 @@ jobs:
           distribution: 'temurin'
           java-version: 8
       - name: Run a full build
-        run: ./dev-support/checks/build.sh -Prelease assembly:single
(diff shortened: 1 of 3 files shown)


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #1123, opened 2024-07-13
RATIS-2125. javadoc jars should not include non-doc items

adoroszlai:

Ratis build generates javadoc when using `-Prelease`.  The javadoc jars have unexpected content, e.g. other jars, `classes` dir (with all `.class` files), etc.

Steps to reproduce:

```
mvn -DskipTests -Prelease clean package
unzip -d before ratis-server-api/target/ratis-server-api-3.2.0-SNAPSHOT-javadoc.jar
ls -1sh before
```

The problem happens because [outputDirectory](https://maven.apache.org/plugins/maven-javadoc-plugin/jar-mojo.html#outputDirectory) is set to `project.build.directory`, which is `target/`.  This directory is also where Maven build puts other items, which then get included in the javadoc jar.

The default location in `maven-javadoc-plugin` is `target/apidocs`, we can use that.

https://issues.apache.org/jira/browse/RATIS-2125

## How was this patch tested?

Steps described in repro, but using `after` as directory.

Result:

```
$ diff -r before/ after/
Only in before: classes
Only in before: generated-sources
Only in before: javadoc-bundle-options
Only in before: maven-archiver
Only in before: maven-javadoc-plugin-stale-data.txt
Only in before: maven-shared-archive-resources
Only in before: .plxarc
Only in before: ratis-server-api-3.2.0-SNAPSHOT.jar
Only in before: ratis-server-api-3.2.0-SNAPSHOT-sources.jar
Only in before: ratis-server-api-3.2.0-SNAPSHOT-tests.jar
Only in before: rat.txt
Only in before: test-classes
```

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -91,9 +91,9 @@ jobs:
           distribution: 'temurin'
           java-version: ${{ matrix.java }}
       - name: Run a full build
-        run: ./dev-support/checks/build.sh -Prelease -Djavac.version=${{ matrix.java }}
+        run: ./dev-support/checks/build.sh -Djavac.version=${{ matrix.java }}
       - name: Test reproducibility
-        run: ./dev-support/checks/repro.sh -Prelease -Djavac.version=${{ matrix.java }}
+        run: ./dev-support/checks/repro.sh -Djavac.version=${{ matrix.java }}
   rat:
     name: rat
     runs-on: ubuntu-20.04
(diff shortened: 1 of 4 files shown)


adoroszlai:
Reproducibility check for javadoc works on Java 8, but fails with newer Java.  The problem is in the search index JSON files.  So it's reverted for now.

szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #1125, opened 2024-07-25
RATIS-2131: Surround with [] only if hostName is a IPv6 string

mridulm:

Currently, if the address is IPv6, NetUtils.address2String always surrounds it with '[]'.
This should be done only when the input is an IP string, and not for hostnames (which have an IPv6 address)

RATIS-2131

## How was this patch tested?

Manual testing


mridulm:
Verified with Apache Celeborn, which uses 3.1.0

mridulm:
+CC @szetszwo based on git blame.

szetszwo (approved):
+1 the change looks good.

mridulm:
Thanks @szetszwo !

PR #1137, opened 2024-08-21
RATIS-2140. Thread wait when installing snapshot.

z-bb:
see https://issues.apache.org/jira/browse/RATIS-2140


szetszwo (approved):
+1 the change looks good.

PR #1138, opened 2024-08-22
RATIS-2144 SegmentedRaftLogWorker should close the stream before releasing the buffer.

OneSizeFitsQuorum:
see [jira](https://issues.apache.org/jira/browse/RATIS-2144)


OneSizeFitsQuorum:
@szetszwo PTAL

szetszwo (approved):
+1 the change looks good.

PR #1139, opened 2024-08-24
RATIS-2145. Follower hangs until the next trigger to take a snapshot.

z-bb:
see: https://issues.apache.org/jira/browse/RATIS-2145


szetszwo (approved):
+1 the change looks good.

PR #1142, opened 2024-09-03
RATIS-2147. Md5 mismatch when snapshot install

133tosakarin:
see https://issues.apache.org/jira/browse/RATIS-2147


ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
```diff
@@ -114,7 +115,8 @@ public void installSnapshot(InstallSnapshotRequestProto request, StateMachine st
 
     // TODO: Make sure that subsequent requests for the same installSnapshot are coming in order,
     // and are not lost when whole request cycle is done. Check requestId and requestIndex here
-
+    MessageDigest digester = newMd5Digest();
+    digester.reset();
```
szetszwo, line 119: Let's use `MD5Hash.getDigester()`.

szetszwo (commented):
@133tosakarin , thanks a lot for working on this!  Please remove the unused code instead of commenting them out.  See also the comment inlined and below
```java
diff --git a/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java b/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
index 794604d66b..88a7555dcb 100644
--- a/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
@@ -27,7 +27,6 @@ import org.apache.ratis.statemachine.SnapshotInfo;
 import org.apache.ratis.statemachine.StateMachine;
 import org.apache.ratis.statemachine.StateMachineStorage;
 import org.apache.ratis.util.FileUtils;
-import org.apache.ratis.util.JavaUtils;
 import org.apache.ratis.util.MD5FileUtil;
 import org.apache.ratis.util.MemoizedSupplier;
@@ -63,7 +62,6 @@ public class SnapshotManager {
   private final Supplier<File> snapshotDir;
   private final Supplier<File> snapshotTmpDir;
   private final Function<FileChunkProto, String> getRelativePath;
-  private final Supplier<MessageDigest> digester = JavaUtils.memoize(MD5Hash::getDigester);
 
   SnapshotManager(RaftPeerId selfId, Supplier<RaftStorageDirectory> dir, StateMachineStorage smStorage) {
     this.selfId = selfId;
@@ -88,7 +86,6 @@ public class SnapshotManager {
       }
       // create the temp snapshot file and put padding inside
       out = FileUtils.newFileChannel(tmpSnapshotFile, StandardOpenOption.WRITE, StandardOpenOption.CREATE);
-      digester.get().reset();
     } else {
       if (!exists) {
         throw new FileNotFoundException("Chunk offset is non-zero but file is not found: " + tmpSnapshotFile
@@ -114,7 +111,7 @@ public class SnapshotManager {
 
     // TODO: Make sure that subsequent requests for the same installSnapshot are coming in order,
     // and are not lost when whole request cycle is done. Check requestId and requestIndex here
-
+    final MessageDigest digester = MD5Hash.getDigester();
     for (FileChunkProto chunk : snapshotChunkRequest.getFileChunksList()) {
       SnapshotInfo pi = stateMachine.getLatestSnapshot();
       if (pi != null && pi.getTermIndex().getIndex() >= lastIncludedIndex) {
@@ -128,7 +125,7 @@ public class SnapshotManager {
 
       try (FileChannel out = open(chunk, tmpSnapshotFile)) {
         final ByteBuffer data = chunk.getData().asReadOnlyByteBuffer();
-        digester.get().update(data.duplicate());
+        digester.update(data.duplicate());
 
         int written = 0;
         for(; data.remaining() > 0; ) {
@@ -144,7 +141,7 @@ public class SnapshotManager {
             new MD5Hash(chunk.getFileDigest().toByteArray());
         // calculate the checksum of the snapshot file and compare it with the
         // file digest in the request
-        final MD5Hash digest = new MD5Hash(digester.get().digest());
+        final MD5Hash digest = new MD5Hash(digester.digest());
         if (!digest.equals(expectedDigest)) {
           LOG.warn("The snapshot md5 digest {} does not match expected {}",
               digest, expectedDigest);
```

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
133tosakarin: Since I am not sure if it is a problem with JavaUtils.memoize, but I can confirm that it is a problem with digest, my idea is to use a new digest each time, which is more secure and reliable, and also avoids the situation where the digest calculation buffer is destroyed and the final result is incorrect

szetszwo: `MD5Hash.getDigester()` is safe since it is a `Threadlocal` and reset the `digest` every time, i.e. the same thread will always use the same `digest` object.

133tosakarin: Yes, ThreadLocal variables should be safe. However, we have stably reproduced this issue in multiple different testing environments, and we are also identifying whether there is a problem with the JDK version we are using. During this process, we have made an important discovery that even if we receive an empty file, Corrupt will still appear.

image

szetszwo: I see.  For an empty file, we probably need to handle it differently.

133tosakarin: well! I am testing in our environment to see if there are still problems. 

Before that, I will convert this PR to draft

OneSizeFitsQuorum:
We spent a lot of time troubleshooting this issue, and what we were able to confirm in the end is that there seems to be some problem with `digester.reset`, which causes the final MD5 calculation to be incorrect. We searched for related issues on the OpenJDK official [website](https://bugs.openjdk.org/browse/JDK-8300416?jql=project%20%3D%20JDK%20AND%20issuetype%20%3D%20Bug%20AND%20affectedVersion%20in%20releasedVersions()%20AND%20text%20~%20%22MD5%22%20ORDER%20BY%20created%20DESC) but couldn't find an exact cause.

However, we found that instead of using reset, creating a new digester each time a new file appears should solve the issue (our testing process can verify this). Therefore, I recommend making this change directly. 

What's your opinion? @szetszwo

133tosakarin:
@szetszwo What do you think of this way?

133tosakarin:
private String randomString(int length) {
        StringBuilder sb = new StringBuilder();
        for (int i = 0; i < length; i++) {
            sb.append((char) (Math.random() * 26 + 'a'));
        }
        return sb.toString();
    }

    @Benchmark
    public void md5Test(Blackhole blackhole) throws NoSuchAlgorithmException {
        int length = 1000;
        md5 = MessageDigest.getInstance("MD5");
        for (int i = 0; i < 1000; ++i) {
            md5.reset();
            md5.update(randomString(length).getBytes());
            blackhole.consume(md5.digest());
            length += 100;
        }
    }

    @Benchmark
    public void md5Test2(Blackhole blackhole) throws NoSuchAlgorithmException {
        int length = 10000;
        for (int i = 0; i < 1000; ++i) {
            MessageDigest md5 = MessageDigest.getInstance("MD5");
            md5.reset();
            md5.update(randomString(length).getBytes());
            blackhole.consume(md5.digest());
            length += 100;
        }

    }

    @Benchmark
    public void md5DoNothingTest() throws NoSuchAlgorithmException {
        MessageDigest md5 = MessageDigest.getInstance("MD5");
        for (int i = 0; i < 1000000; ++i) {
            md5.reset();
        }
    }

    @Benchmark
    public void md5DoNothingTest2() throws NoSuchAlgorithmException {
        for (int i = 0; i < 1000000; ++i) {
            MessageDigest md5 = MessageDigest.getInstance("MD5");
            md5.reset();
        }
    }

133tosakarin:
Screenshot 2024-09-10 at 11 59 04

133tosakarin:
image

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
```diff
@@ -88,7 +89,8 @@ private FileChannel open(FileChunkProto chunk, File tmpSnapshotFile) throws IOEx
       }
       // create the temp snapshot file and put padding inside
       out = FileUtils.newFileChannel(tmpSnapshotFile, StandardOpenOption.WRITE, StandardOpenOption.CREATE);
-      digester.get().reset();
+      digester = newMd5Digest();
+      digester.reset();
```
OneSizeFitsQuorum, line 93: We can move this reset to line182 ?

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
ivandika3: Just taking a quick look. I think simply removing the `MemoizedSupplier` (just calling `MD5Hash#getDigester` every time) should technically work. From what I see, `MemoizedSupplier` only call the `MessageDigest#reset` at the first lazy instantiation. The subsequent `MemoizedSupplier` invocation will just return the `MessageDigest` without calling the `MessageDigest#reset`.

Ozone encountered some thread local MD5 issue (https://github.com/apache/ozone/pull/6435), might be related.

133tosakarin: > Just taking a quick look. I think simply removing the `MemoizedSupplier` (just calling `MD5Hash#getDigester` every time) should technically work. From what I see, `MemoizedSupplier` only call the `MessageDigest#reset` at the first invocation. The subsequent `MemoizedSupplier` invocation will just return the `MessageDigest` instance without calling the `MessageDigest#reset` which might cause the mismatch issue.
> 
> Ozone encountered some thread local MD5 issue ([apache/ozone#6435](https://github.com/apache/ozone/pull/6435)), might be related.

Thank you for your comment!

I also think that whether to remove MemoizedSupplier or not is irrelevant to the problem. My idea is to use a new digester for each file to solve this problem well.

I have pushed the code to this PR, you can take a look

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
133tosakarin: That's ok

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
ivandika3: Yes, instantiating a new digester is should also fine, but there might be more memory overhead and short lived objects compared to the ThreadLocal implementation (which might also be fine if the operation is infrequent).

If the issue can be resolved by simply removing the `MemoizedSupplier`, I think we can consider this instead of instantiating a new digester every time.

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
```diff
@@ -173,9 +172,10 @@ public void installSnapshot(InstallSnapshotRequestProto request, StateMachine st
     }
   }
 
-  private MessageDigest newMd5Digest() throws IOException {
+  private void renewMd5Digest() throws IOException {
     try {
-      return MessageDigest.getInstance("MD5");
+      digester =  MessageDigest.getInstance("MD5");
+      digester.reset();
```
ivandika3, line 178: I think calling `MessageDigest#reset` here is unnecessary since a newly instantiated `MessageDigest` should already be "reset".

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
133tosakarin: > Just taking a quick look. I think simply removing the `MemoizedSupplier` (just calling `MD5Hash#getDigester` every time) should technically work. From what I see, `MemoizedSupplier` only call the `MessageDigest#reset` at the first invocation. The subsequent `MemoizedSupplier` invocation will just return the `MessageDigest` instance without calling the `MessageDigest#reset` which might cause the mismatch issue.
> 
> Ozone encountered some thread local MD5 issue ([apache/ozone#6435](https://github.com/apache/ozone/pull/6435)), might be related.



> Just taking a quick look. I think simply removing the `MemoizedSupplier` (just calling `MD5Hash#getDigester` every time) should technically work. From what I see, `MemoizedSupplier` only call the `MessageDigest#reset` at the first invocation. The subsequent `MemoizedSupplier` invocation will just return the `MessageDigest` instance without calling the `MessageDigest#reset` which might cause the mismatch issue.

image
In fact, digester.get ().reset () will be executed every time the first chunk is received.

In addition, I have done JMH testing, and it takes 0.5s for new 1000000 digests, but in fact, we do not send so many files so frequently, and this overhead is not a problem

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
133tosakarin: Thank you for your reminder

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
```diff
@@ -63,7 +64,7 @@ public class SnapshotManager {
   private final Supplier<File> snapshotDir;
   private final Supplier<File> snapshotTmpDir;
   private final Function<FileChunkProto, String> getRelativePath;
-  private final Supplier<MessageDigest> digester = JavaUtils.memoize(MD5Hash::getDigester);
+  private MessageDigest digester;
```
szetszwo, line 67: Let's move it to a local variable?

szetszwo (commented):
@133tosakarin , thanks for working on this!

Could we covert the `digester` to a local variable?
```java
diff --git a/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java b/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
index 794604d66b..6f36757189 100644
--- a/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
@@ -27,7 +27,6 @@ import org.apache.ratis.statemachine.SnapshotInfo;
 import org.apache.ratis.statemachine.StateMachine;
 import org.apache.ratis.statemachine.StateMachineStorage;
 import org.apache.ratis.util.FileUtils;
-import org.apache.ratis.util.JavaUtils;
 import org.apache.ratis.util.MD5FileUtil;
 import org.apache.ratis.util.MemoizedSupplier;
@@ -63,7 +62,6 @@ public class SnapshotManager {
   private final Supplier<File> snapshotDir;
   private final Supplier<File> snapshotTmpDir;
   private final Function<FileChunkProto, String> getRelativePath;
-  private final Supplier<MessageDigest> digester = JavaUtils.memoize(MD5Hash::getDigester);
 
   SnapshotManager(RaftPeerId selfId, Supplier<RaftStorageDirectory> dir, StateMachineStorage smStorage) {
     this.selfId = selfId;
@@ -88,7 +86,6 @@ public class SnapshotManager {
       }
       // create the temp snapshot file and put padding inside
       out = FileUtils.newFileChannel(tmpSnapshotFile, StandardOpenOption.WRITE, StandardOpenOption.CREATE);
-      digester.get().reset();
     } else {
       if (!exists) {
         throw new FileNotFoundException("Chunk offset is non-zero but file is not found: " + tmpSnapshotFile
@@ -112,9 +109,9 @@ public class SnapshotManager {
     LOG.info("Installing snapshot:{}, to tmp dir:{}",
         ServerStringUtils.toInstallSnapshotRequestString(request), tmpDir);
 
+    final MessageDigest digester = MD5Hash.getDigester();
     // TODO: Make sure that subsequent requests for the same installSnapshot are coming in order,
     // and are not lost when whole request cycle is done. Check requestId and requestIndex here
-
     for (FileChunkProto chunk : snapshotChunkRequest.getFileChunksList()) {
       SnapshotInfo pi = stateMachine.getLatestSnapshot();
       if (pi != null && pi.getTermIndex().getIndex() >= lastIncludedIndex) {
@@ -128,7 +125,7 @@ public class SnapshotManager {
 
       try (FileChannel out = open(chunk, tmpSnapshotFile)) {
         final ByteBuffer data = chunk.getData().asReadOnlyByteBuffer();
-        digester.get().update(data.duplicate());
+        digester.update(data.duplicate());
 
         int written = 0;
         for(; data.remaining() > 0; ) {
@@ -144,7 +141,7 @@ public class SnapshotManager {
             new MD5Hash(chunk.getFileDigest().toByteArray());
         // calculate the checksum of the snapshot file and compare it with the
         // file digest in the request
-        final MD5Hash digest = new MD5Hash(digester.get().digest());
+        final MD5Hash digest = new MD5Hash(digester.digest());
         if (!digest.equals(expectedDigest)) {
           LOG.warn("The snapshot md5 digest {} does not match expected {}",
               digest, expectedDigest);
```

szetszwo:
> For each received file, the difference between using a new md5 object and using a member variable is almost negligible

I am fine to use a new digester.   Then, let's move the new MD5 method to `MD5Hash.java`.

```java
diff --git a/ratis-common/src/main/java/org/apache/ratis/io/MD5Hash.java b/ratis-common/src/main/java/org/apache/ratis/io/MD5Hash.java
index e60bef9652..8aacd9a65d 100644
--- a/ratis-common/src/main/java/org/apache/ratis/io/MD5Hash.java
+++ b/ratis-common/src/main/java/org/apache/ratis/io/MD5Hash.java
@@ -29,14 +29,15 @@ import java.util.Arrays;
 public class MD5Hash {
   public static final int MD5_LEN = 16;
 
-  private static final ThreadLocal<MessageDigest> DIGESTER_FACTORY =
-      ThreadLocal.withInitial(() -> {
+  private static final ThreadLocal<MessageDigest> DIGESTER_FACTORY = ThreadLocal.withInitial(MD5Hash::newDigester);
+
+  public static MessageDigest newDigester() {
     try {
       return MessageDigest.getInstance("MD5");
     } catch (NoSuchAlgorithmException e) {
-      throw new RuntimeException(e);
+      throw new IllegalStateException("Failed to create MessageDigest for MD5", e);
     }
-  });
+  }
 
   private byte[] digest;
 
diff --git a/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java b/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
index 794604d66b..5b5158aebd 100644
--- a/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
@@ -112,9 +109,9 @@ public class SnapshotManager {
     LOG.info("Installing snapshot:{}, to tmp dir:{}",
         ServerStringUtils.toInstallSnapshotRequestString(request), tmpDir);
 
+    final MessageDigest digester = MD5Hash.newDigester();
     // TODO: Make sure that subsequent requests for the same installSnapshot are coming in order,
     // and are not lost when whole request cycle is done. Check requestId and requestIndex here
-
     for (FileChunkProto chunk : snapshotChunkRequest.getFileChunksList()) {
       SnapshotInfo pi = stateMachine.getLatestSnapshot();
       if (pi != null && pi.getTermIndex().getIndex() >= lastIncludedIndex) {
```

133tosakarin:
> > For each received file, the difference between using a new md5 object and using a member variable is almost negligible
> 
> Thanks for running the benchmarks! Let's use a new digester. How about moving the new MD5 method to `MD5Hash`?
> 
> ```java
> diff --git a/ratis-common/src/main/java/org/apache/ratis/io/MD5Hash.java b/ratis-common/src/main/java/org/apache/ratis/io/MD5Hash.java
> index e60bef9652..8aacd9a65d 100644
> --- a/ratis-common/src/main/java/org/apache/ratis/io/MD5Hash.java
> +++ b/ratis-common/src/main/java/org/apache/ratis/io/MD5Hash.java
> @@ -29,14 +29,15 @@ import java.util.Arrays;
>  public class MD5Hash {
>    public static final int MD5_LEN = 16;
>  
> -  private static final ThreadLocal<MessageDigest> DIGESTER_FACTORY =
> -      ThreadLocal.withInitial(() -> {
> +  private static final ThreadLocal<MessageDigest> DIGESTER_FACTORY = ThreadLocal.withInitial(MD5Hash::newDigester);
> +
> +  public static MessageDigest newDigester() {
>      try {
>        return MessageDigest.getInstance("MD5");
>      } catch (NoSuchAlgorithmException e) {
> -      throw new RuntimeException(e);
> +      throw new IllegalStateException("Failed to create MessageDigest for MD5", e);
>      }
> -  });
> +  }
>  
>    private byte[] digest;
>  
> diff --git a/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java b/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
> index 794604d66b..5b5158aebd 100644
> --- a/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
> +++ b/ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
> @@ -112,9 +109,9 @@ public class SnapshotManager {
>      LOG.info("Installing snapshot:{}, to tmp dir:{}",
>          ServerStringUtils.toInstallSnapshotRequestString(request), tmpDir);
>  
> +    final MessageDigest digester = MD5Hash.newDigester();
>      // TODO: Make sure that subsequent requests for the same installSnapshot are coming in order,
>      // and are not lost when whole request cycle is done. Check requestId and requestIndex here
> -
>      for (FileChunkProto chunk : snapshotChunkRequest.getFileChunksList()) {
>        SnapshotInfo pi = stateMachine.getLatestSnapshot();
>        if (pi != null && pi.getTermIndex().getIndex() >= lastIncludedIndex) {
> ```

Good idea, but there are some problems in the provided code, we should construct digester when receiving the first chunk and calculate md5 after receiving a complete file.

Therefore, we add a MessageDigest member variable in SnapshotManager and assign it to the first chunk every time

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
133tosakarin: I explained it in the code snippet you provided, please take a look.

szetszwo (commented):
@133tosakarin , thanks for the update!  The change looks good.  Could you fix the checkstyle issues?

szetszwo (approved):
+1 the change looks good.

OneSizeFitsQuorum:
Maybe we need a atomicReference for this digester？ @szetszwo

szetszwo:
@OneSizeFitsQuorum , That's a good point!  We should have an `AtomicReference`.

PR #1149, opened 2024-09-14
RATIS-2157 Enhance make_rc.sh for non-first rc at release time

OneSizeFitsQuorum:
see [jira](https://issues.apache.org/jira/browse/RATIS-2157)


OneSizeFitsQuorum:
@szetszwo PTAL

szetszwo (approved):
+1 the change looks good.

PR #1165, opened 2024-10-09
RATIS-2170. Improve some logging in GrpcLogAppender

jianghuazhu:
The purpose of this pr is to improve the log in GrpcLogAppender.

https://issues.apache.org/jira/browse/RATIS-2170

## How was this patch tested?
Need to make sure the unit tests pass.


jianghuazhu:
@szetszwo , can you help review this PR?
Thanks.

szetszwo (commented):
@jianghuazhu , let's don't change this assertion message.  Using the same name seems better.

BTW, `Preconditions` is for bug detection but not for logging.  So the messages are relatively unimportant since we won't see them unless there is a bug.

jianghuazhu:
Ok, thanks @szetszwo .

PR #1174, opened 2024-11-08
RATIS-2177. Delete segmentLog from small to large according to logIndex

133tosakarin:
see #RATIS-2177

In addition, I made some improvements to TruncateSegment.maxEndIndex (), because segments are sorted by index from largest to smallest, so returning the endIndex of the first segment is enough.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
@@ -151,8 +151,8 @@ public class SegmentedRaftLogCache {
       if (toTruncate != null) {
         max = toTruncate.endIndex;
       }
-      for(SegmentFileInfo d : toDelete) {
-        max = Math.max(max, d.endIndex);
+      if (toDelete != null && toDelete.length > 0) {
+        max = Math.max(max, toDelete[0].endIndex);
       }
       return max;
     }
@@ -359,7 +359,7 @@ public class SegmentedRaftLogCache {
       try (AutoCloseableLock writeLock = writeLock()) {
         int segmentIndex = binarySearch(index);
         List<LogSegment> list = new LinkedList<>();
-
+        LOG.info("purgeLog are: {}", segments);
         if (segmentIndex == -segments.size() - 1) {
           list.addAll(segments);
           segments.clear();
@@ -383,6 +383,7 @@ public class SegmentedRaftLogCache {
         list.forEach(LogSegment::evictCache);
         List<SegmentFileInfo> toDelete = list.stream().map(SegmentFileInfo::newClosedSegmentFileInfo)
             .collect(Collectors.toList());
+        LOG.info("final toDelete: {}", toDelete);
         return list.isEmpty() ? null : new TruncationSegments("purge(" + index + ")", null, toDelete);
       }
     }
--- a/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
@@ -151,8 +151,8 @@ public class SegmentedRaftLogCache {
       if (toTruncate != null) {
         max = toTruncate.endIndex;
       }
-      if (toDelete != null && toDelete.length > 0) {
-        max = Math.max(max, toDelete[0].endIndex);
+      for(SegmentFileInfo d : toDelete) {
+        max = Math.max(max, d.endIndex);
       }
       return max;
     }
@@ -359,7 +359,7 @@ public class SegmentedRaftLogCache {
       try (AutoCloseableLock writeLock = writeLock()) {
         int segmentIndex = binarySearch(index);
         List<LogSegment> list = new LinkedList<>();
-        LOG.info("purgeLog are: {}", segments);
+
         if (segmentIndex == -segments.size() - 1) {
           list.addAll(segments);
           segments.clear();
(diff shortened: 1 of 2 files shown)


OneSizeFitsQuorum (commented):
image
Maybe we need to adjust the order here and delete them in order from smallest to largest to avoid bugs altogether?

133tosakarin:
> image Maybe 
we need to adjust the order here and delete them in order from smallest to largest to avoid bugs altogether?

these segments that will be deleted  are sorted at constructing TruncateSegment, so we may not need to adjust the order here. 

BTW, segments are sorted in descending order  in TruncateSegments.

So, in this PR, I use reverse order traversal to delete segments

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
```diff
@@ -151,8 +151,8 @@ long maxEndIndex() {
       if (toTruncate != null) {
         max = toTruncate.endIndex;
       }
-      for(SegmentFileInfo d : toDelete) {
-        max = Math.max(max, d.endIndex);
+      if (toDelete != null && toDelete.length > 0) {
+        max = Math.max(max, toDelete[0].endIndex);
```
szetszwo, line 155: Let's keep the origin code since it is more robust -- it does not depend on the ordering.

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
```diff
@@ -359,7 +359,7 @@ TruncationSegments purge(long index) {
       try (AutoCloseableLock writeLock = writeLock()) {
         int segmentIndex = binarySearch(index);
         List<LogSegment> list = new LinkedList<>();
-
+        LOG.info("purgeLog are: {}", segments);
```
szetszwo, line 362: Print also the name and the index:
```java
        LOG.info("{}: purge to index {} from closed segments {}", name, index, segments);
```

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
```diff
@@ -383,6 +383,7 @@ TruncationSegments purge(long index) {
         list.forEach(LogSegment::evictCache);
         List<SegmentFileInfo> toDelete = list.stream().map(SegmentFileInfo::newClosedSegmentFileInfo)
             .collect(Collectors.toList());
+        LOG.info("final toDelete: {}", toDelete);
```
szetszwo, line 386: Print `LogSegment` instead of `SegmentFileInfo`.  It is easier to read.
```java
        LOG.info("{}: toDelete segment: {}", name, list);
```

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogWorker.java
```diff
@@ -694,7 +695,9 @@ void execute() throws IOException {
 
       if (segments.getToDelete() != null && segments.getToDelete().length > 0) {
         long minStart = segments.getToDelete()[0].getStartIndex();
-        for (SegmentFileInfo del : segments.getToDelete()) {
+        SegmentFileInfo[] toDelete = segments.getToDelete();
+        for (int i = toDelete.length - 1; i >= 0; i--) {
+          final SegmentFileInfo del = toDelete[i];
```
szetszwo, line 700: `TruncateLog` should not be changed -- it should delete from high to low.

szetszwo (commented):
@133tosakarin , thanks for working on this!  Purging was fixed by RATIS-1763 but then reintroduced by RATIS-1887 for truncation.  The ordering should be
- Purging: low to high.
- Truncation: high to low.

Please see the comments inlined.

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogWorker.java
133tosakarin: Deleting from high to low may cause gaps, such as power outages, kill -9, etc. during the deletion process.

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
133tosakarin: I have reverted this line

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
133tosakarin: I have reverted this line

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogCache.java
133tosakarin: I have reverted this line

szetszwo (approved):
+1 the change looks good.

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLogWorker.java
szetszwo: Log is a list of entries with consecutive indices.
- Truncate is an operation to delete the log from higher end -- higher indices should be deleted first.
- Purge is an operation to delete the log from the lower end -- lower index should be deleted first.

Suppose we have the following log files (RATIS-2186).
log-88363996241_88364000257, 
log-88364000258_88364004199, 
log-88364004200_88364008231, 
log-88364008232_88364012246, 
log-88364012247_88364016452, 
log-88364016453_88364020483. 

Truncate(88364012246) should delete the last two files log-88364012247_88364016452 and log-88364016453_88364020483.   We should delete the higher index first in order to avoid holes.

PR #1178, opened 2024-11-15
RATIS-2189. Use ByteBufAllocator#ioBuffer in NettyDataStreamUtils

ivandika3:

See https://issues.apache.org/jira/browse/RATIS-2189

## How was this patch tested?

Checked using `jmap -histo:live` that `DirectByteBuffer` are not created when using the following arguments:

```
-Dorg.apache.ratis.thirdparty.io.netty.noPreferDirect=true -Dorg.apache.ratis.thirdparty.io.netty.noUnsafe=true -Dorg.apache.ratis.thirdparty.io.netty.allocator.numDirectArenas=0
```


szetszwo (approved):
@ivandika3 , thanks a lot for working on this!  

+1 The change looks good.

PR #1180, opened 2024-11-22
RATIS-2191. Add ResourceLeakDetector to Netty tests

ivandika3:

Similar to LeakDetector#assertNoLeaks in MiniRaftCluster, we can also add the Netty leak detector in tests as specified in https://netty.io/wiki/reference-counted-objects.html#wiki-h2-9 to ensure that all ByteBuf are released.

https://issues.apache.org/jira/browse/RATIS-2191

## How was this patch tested?

Intentionally add leak add check that `IllegalStateException` is thrown: https://github.com/ivandika3/ratis/actions/runs/11871018319/job/33121223662


szetszwo (approved):
+1 the change looks good.

PR #1183, opened 2024-11-25
RATIS-2194. FileLock didn't unlock properly

chungen0126:

Found OverlappingFileLockExceptio when restarting RaftServer. 

When closing `RaftServerProxy`, all `RaftServerImpl` in `implMap` should be closed. However, if there was a new `RaftServerImpl` started by `RaftServerProxy#groupManagementAsync`, it would not be closed properly. The log will say that the new `RaftServerImpl` has been closed, but its raft store is still being initialised by `implExecutor`.

Things change:
- Closing implExecutor before closing implMap prevents RaftServerImpl from starting after it is closed by implMap.

https://issues.apache.org/jira/browse/RATIS-2194

## How was this patch tested?
https://github.com/chungen0126/ratis/actions/runs/12008556158


szetszwo (approved):
+1 the change looks good.

PR #1189, opened 2024-12-14
RATIS-2200. Make Ratis example log at INFO level

kiruthiga1793:

made `ratis-example` logs as info level

https://issues.apache.org/jira/browse/RATIS-2200

## How was this patch tested?

### DIFF
--- a/ratis-examples/src/main/resources/log4j.properties
+++ b/ratis-examples/src/main/resources/log4j.properties
@@ -16,8 +16,6 @@
 
 
 log4j.rootLogger=INFO, stdout
-log4j.logger.org.apache.ratis.server.impl.RatisServerImpl=INFO
-log4j.logger.org.apache.ratis.client.RaftClient=INFO
 
 log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.Target=System.out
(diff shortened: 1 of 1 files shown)


ratis-examples/src/main/resources/log4j.properties
```diff
@@ -16,10 +16,10 @@
 
 
 log4j.rootLogger=INFO, stdout
-log4j.logger.org.apache.ratis.server.impl.RatisServerImpl=DEBUG
-log4j.logger.org.apache.ratis.client.RaftClient=DEBUG
+log4j.logger.org.apache.ratis.server.impl.RatisServerImpl=INFO
+log4j.logger.org.apache.ratis.client.RaftClient=INFO
```
adoroszlai, line 20: Since root level is INFO, I think these two class-specific settings can be removed.

```suggestion
```

adoroszlai (commented):
Thanks @kiruthiga1793 for the patch.

ratis-examples/src/main/resources/log4j.properties
kiruthiga1793: @adoroszlai Thanks for the suggestion. I will update that and commit again.

adoroszlai (approved):

adoroszlai:
Thanks @kiruthiga1793 for updating the patch.

PR #1195, opened 2024-12-22
RATIS-2210. Reduce duplication in CI workflow

adoroszlai:

Reduce duplication in CI workflow.

- Create a reusable workflow (`check.yml`) for a single, generic check.
- Refactor existing checks in `post-commit.yml` (except `coverage`, which needs extra tokens) to call the reusable one.  Minor changes as implementation detail:
  - Separate `repro` check from `compile`.
  - Create separate `compile.sh` for `compile` check.
- Group simple checks (`author`, `checkstyle`, etc.) with matrix strategy as `basic`.
- Split existing `post-commit.yml` into:
  - `ci.yml`, has most of the logic, is triggered by workflow call
  - `post-commit.yml`, triggered by `push`/`pull_request`, simply calls `ci.yml`
  This split is not strictly required, but I think the way GitHub UI displays workflow runs is better this way (runs: [before](https://github.com/adoroszlai/ratis/actions/runs/12341153183) / [after](https://github.com/adoroszlai/ratis/actions/runs/12347469076) for comparison).

https://issues.apache.org/jira/browse/RATIS-2210

## How was this patch tested?

CI:
https://github.com/adoroszlai/ratis/actions/runs/12452887900

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,206 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #1203, opened 2025-01-02
RATIS-2232. Improve log for debugging on RaftLog / TransactionManager

SzyWilliam:
See https://issues.apache.org/jira/browse/RATIS-2232.
Thanks to the patch provided by @szetszwo


szetszwo (approved):
+1 the change looks good.

PR #1205, opened 2025-01-02
RATIS-2234. Remove lock race between heartbeat and append log channels

SzyWilliam:
See https://issues.apache.org/jira/browse/RATIS-2234.
See discussion at https://issues.apache.org/jira/browse/RATIS-2208

Thanks @szetszwo for the patch provided!


szetszwo (approved):
+1 the change looks good.

PR #1210, opened 2025-01-10
RATIS-2237 Add getFollowerMatchIndices interface for DivisionInfo

OneSizeFitsQuorum:
see https://issues.apache.org/jira/browse/RATIS-2237


OneSizeFitsQuorum:
@szetszwo PTAL

szetszwo (approved):
+1 the change looks good.

PR #1223, opened 2025-02-19
RATIS-2250. Add RAT exclusions for bootstrap and jquery

adoroszlai:
## What changes were proposed in this pull request?

Addendum for #1222.  Exclude bootstrap and jquery minimized scripts for RAT check.

## How was this patch tested?

```
./build.sh ../ratis-site
```

PR #1222, opened 2025-02-19
RATIS-2250. Website broken due to using remote resources

adoroszlai:
## What changes were proposed in this pull request?

Ratis website is broken due to new Content Security Policy.

before

Fix:

- Add bootstrap and jquery scripts to `static` to avoid using remote resources
- Remove broken embedded slides (keep links)

https://issues.apache.org/jira/browse/RATIS-2250

## How was this patch tested?

```
hugo serve
open http://localhost:1313
```

after

PR #1224, opened 2025-02-19
RATIS-2250. Website broken due to using remote resources

adoroszlai:

Generate website after #1222 (locally using change from #1223).

This is the first update after 3.1.2, so it also includes content from #1211 and changes done in #1213.


SzyWilliam:
@adoroszlai Thanks for working on fixing the website. Just tried #1222 locally and everything looks good to me.
image

adoroszlai:
Thanks @SzyWilliam for the review.

szetszwo:
Just found that the diagrams in the Feature section do not show up:
image

adoroszlai:
Thanks @szetszwo for finding it.  Looks like these come from `bootstrap.min.css` referencing `../fonts/`, which we didn't copy.

PR #1231, opened 2025-02-27
RATIS-2022. Replace some Guava usage with builtins

adoroszlai:

Replace few usages of Guava with equivalents from JDK or Ratis.

https://issues.apache.org/jira/browse/RATIS-2022

## How was this patch tested?

CI:
https://github.com/adoroszlai/ratis/actions/runs/13572312975


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #1246, opened 2025-04-05
RATIS-2274. Newly added peer may retain outdated configuration after membership change, causing election failure.

RkGrit:

When the leader has no snapshot (forcing log-based catch-up), the new peer might lack timely configuration change logs due to stagingCatchupGap. This leaves the new peer with outdated configurations, violating Raft's membership change safety guarantees.
To solve this problem, the Leader is essentially required to ensure that the configuration of the Follower is the latest when the Follower completes the catch-up.
We add one more condition to the CAUGHTUP judgment logic. That is, CAUGHTUP is complete only after a follower receives the latest conf logs from the leader.

https://issues.apache.org/jira/browse/RATIS-2274

## How was this patch tested?

unit tests


RkGrit:
@szetszwo
Please review this PR

szetszwo (approved):
+1 the change looks good.

PR #1247, opened 2025-04-07
RATIS-2278. Follower Fails to Append Entries Due to Index Validation in NavigableIndices

RkGrit:

When a follower receives multiple appendEntries requests from the leader, it may throw an IllegalStateException during log index validation. Specifically, the NavigableIndices class fails to handle concurrent log appends properly, causing the follower to incorrectly reject valid entries with an "index mismatch" error.

The reason is that removeExisting can be called out of order. So we changing the Preconditions to check startIndex.

https://issues.apache.org/jira/browse/RATIS-2278

## How was this patch tested?

unit tests


RkGrit:
@szetszwo
Please review this PR

szetszwo (approved):
+1 the change looks good.

RkGrit:
@szetszwo 
please merge the PR

PR #1259, opened 2025-05-13
RATIS-2292. Remove some imports of '.*'

jianghuazhu:
Remove some '.*' imports in Ratis project.

https://issues.apache.org/jira/browse/RATIS-2292

## How was this patch tested?
ci: 
https://github.com/jianghuazhu/ratis/actions/runs/14977684098


slfan1989:
The change is unnecessary. It will create difficulties for future backports.

Tips from steveloughran

```
-1 to all needless import changes, even if cleanup. import changes are the nightmare of cherrypicks, which is why we so rarely clean them up, even ongoing patches. given a choice between "messy imports in existing code" and "backport hell", i will take the mess.
```

cc: @adoroszlai @szetszwo

jianghuazhu:
Middleware should be kept simple enough. Additional introductions may increase instability.

jianghuazhu:
> @jianghuazhu , I agree with @slfan1989 that this kind of changes may not be helpful.
> 
> Let's spend more time on the code but not on the imports. You have filed [RATIS-2171](https://issues.apache.org/jira/browse/RATIS-2171). How is it?
> 
> There are also some test failures: [RATIS-2261](https://issues.apache.org/jira/browse/RATIS-2261), [RATIS-2262](https://issues.apache.org/jira/browse/RATIS-2262), [RATIS-2175](https://issues.apache.org/jira/browse/RATIS-2175), etc. See if you would like to work on them.

Thanks @szetszwo  for the comment and review.
I will try to fix these.

PR #1274, opened 2025-07-02
[RATIS-2314] Fix that SegmentedRaftLogWorker may append entry by itself

jt2594838:
Please refer to https://issues.apache.org/jira/browse/RATIS-2314 for details.


szetszwo (approved):
+1 teh change looks good

PR #1276, opened 2025-07-09
RATIS-2319. Use Concurrent Set for permits in PendingRequests.

nandakumar131:
Use Concurrent Set for `permits` in `PendingRequests`.

Instead of using `HashMap` to track the permits in `PendingRequests#RequestMap` we can use Concurrent Set and avoid explicit synchronization.

RATIS-2319

## How was this patch tested?
Existing unit tests.


nandakumar131:
Thanks @szetszwo for the feedback.

PR #1280, opened 2025-08-13
[RATIS-2321] Fix NPE in ServerState after continuous leader changes

jt2594838:
Please refer to https://issues.apache.org/jira/browse/RATIS-2321 for details.


szetszwo (approved):
+1 the change looks good.

PR #1285, opened 2025-09-16
RATIS-2329. NettyRpcProxy should support handling netty channel exception

leixm:
NettyRpcProxy should support handling netty channel exception

https://issues.apache.org/jira/browse/RATIS-2329

## How was this patch tested?
Manual tests


leixm:
image

leixm:
1. Network anomaly, persistent TimeoutIOException
```
25/09/15 15:39:46,745 WARN [1@group-47BEDE733167->3-LogAppenderDefault-LogAppenderDaemon] LogAppender: 1@group-47BEDE733167->3-LogAppenderDefault: Failed to appendEntries (retry=1): org.apache.ratis.protocol.exceptions.TimeoutIOException
.......
25/09/15 17:14:31,450 WARN [1@group-47BEDE733167->3-LogAppenderDefault-LogAppenderDaemon] LogAppender: 1@group-47BEDE733167->3-LogAppenderDefault: Failed to appendEntries (retry=631): org.apache.ratis.protocol.exceptions.TimeoutIOException 
```

2. Network recovery, append entries stuck like above stack, I think the connection has been closed, but the SimpleChannelInboundHandler implemented by NettyRpcProxy does not implement exceptionCaught.

Env: we use celeborn 0.4.1 version, which depends on ratis 2.5.1 version and configure netty mode.

leixm:
@szetszwo  Can you help review?

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
```diff
@@ -121,6 +124,35 @@ protected void channelRead0(ChannelHandlerContext ctx,
...
+              replies.size(),
+              peer);
+            failOutstandingRequests(new IOException("Connection to " + peer + " is closed."));
+          }
+        }
+
+        private void failOutstandingRequests(Throwable cause) {
+          replies.forEach(f -> f.completeExceptionally(cause));
+          replies.clear();
+        }
```
szetszwo: Let's move it out.  Then, it can also be used in `close()`.

BTW, could you also change the IOException to AlreadyClosedException?
```diff
@@ -153,9 +169,14 @@ public class NettyRpcProxy implements Closeable {
     @Override
     public synchronized void close() {
       client.close();
+      failOutstandingRequests(new AlreadyClosedException("Closing connection to " + peer));
+    }
+
+    private void failOutstandingRequests(Throwable cause) {
       if (!replies.isEmpty()) {
-        final IOException e = new IOException("Connection to " + peer + " is closed.");
-        replies.stream().forEach(f -> f.completeExceptionally(e));
+        LOG.warn("Still have {} requests outstanding from {} connection: {}",
+            replies.size(), peer, cause.toString());
+        replies.forEach(f -> f.completeExceptionally(cause));
         replies.clear();
       }
     }
```

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
```diff
@@ -121,6 +124,35 @@ protected void channelRead0(ChannelHandlerContext ctx,
...
+        @Override
+        public void channelInactive(ChannelHandlerContext ctx) {
+          if (!replies.isEmpty()) {
+            LOG.error(
+              "Still have {} requests outstanding when connection from {} is closed",
+              replies.size(),
+              peer);
+            failOutstandingRequests(new IOException("Connection to " + peer + " is closed."));
+          }
+        }
```
szetszwo: Then, these two methods become very short.  BTW, let's
- Wrap the cause with IOException in the first method so it could include the peer.
- Use `AlreadyClosedException` for the second method and use a different message.
```java
        @Override
        public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
          failOutstandingRequests(new IOException("Caught an exception for the connection to " + peer, cause));
          client.close();
        }

        @Override
        public void channelInactive(ChannelHandlerContext ctx) {
          failOutstandingRequests(new AlreadyClosedException("Channel to " + peer + " is inactive."));
        }
```

leixm:
@szetszwo   Thank you very much for your guidance. I have two questions to discuss:
1. ExceptionCaught is set to IOException, which doesn't cause a reconnection. Refer to IOUtils#shouldReconnect, so this may still cause issues.
2. https://github.com/apache/ratis/blob/master/ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java#L201, when a TimeoutException is caught, should we attempt to reconnect to avoid persistent timeouts in some exceptional cases?

Can we add a new exception type, such as NettyRpcException, to wrap both of the above cases into a NettyRpcException, and then modify IOUtils#shouldReconnect to support reconnecting when a NettyRpcException is encountered?

szetszwo:
> Can we add a new exception type, such as NettyRpcException,  ...

That's a good idea. Let's add it.

ratis-common/src/main/java/org/apache/ratis/util/IOUtils.java
```diff
@@ -93,7 +94,7 @@ static <T> T getFromFuture(CompletableFuture<T> future, Supplier<Object> name, T
   static boolean shouldReconnect(Throwable e) {
     return ReflectionUtils.isInstance(e,
         SocketException.class, SocketTimeoutException.class, ClosedChannelException.class, EOFException.class,
-        AlreadyClosedException.class);
+        AlreadyClosedException.class, NettyRpcException.class);
```
szetszwo: @leixm , Actually, `NettyRpcException` is not needed.  Could we just use `TimeoutIOException` here?

leixm: Sure.

leixm: I have removed NettyRpcException, but we should call `failOutstandingRequests(cause);` in exceptionCaught, and we should not wrap it to IOException, because there may have been a SocketTimeoutException that required reconnection, but because we wrapped it into an IOException, it ultimately did not reconnect.

szetszwo: Let's check also the causes?
```java
  static boolean shouldReconnect(Throwable e) {
    for (; e != null; e = e.getCause()) {
      if (ReflectionUtils.isInstance(e,
          SocketException.class, SocketTimeoutException.class, ClosedChannelException.class, EOFException.class,
          AlreadyClosedException.class, TimeoutIOException.class)) {
        return true;
      }
    }
    return false;
  }
```

leixm:
> After adding `TimeoutIOException.class ` in `shouldReconnect`,`testRequestAsyncWithRetryFailure` failed, should we add new exception types to avoid affecting other code logic?

I can sometimes run it successfully locally, it doesn't seem to be due to this logic change. Can you help me rerun the CI?

szetszwo:
@leixm , thanks for the update.  There is a test failure.  Please take a look.  We probably have to update the test.
```java
[INFO] Running org.apache.ratis.netty.TestRaftAsyncWithNetty
Error:  Tests run: 14, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 83.67 s <<< FAILURE! -- in org.apache.ratis.netty.TestRaftAsyncWithNetty
Error:  org.apache.ratis.netty.TestRaftAsyncWithNetty.testRequestAsyncWithRetryFailureAfterInitialMessages -- Time elapsed: 5.590 s <<< FAILURE!
java.lang.AssertionError: The test "retry-failure-0" does not throw anything.
	at org.apache.ratis.BaseTest.testFailureCase(BaseTest.java:161)
	at org.apache.ratis.BaseTest.testFailureCase(BaseTest.java:169)
	at org.apache.ratis.RaftAsyncTests.runTestRequestAsyncWithRetryFailure(RaftAsyncTests.java:195)
	at org.apache.ratis.RaftAsyncTests.lambda$runTestRequestAsyncWithRetryFailure$0(RaftAsyncTests.java:134)
	at org.apache.ratis.server.impl.MiniRaftCluster$Factory$Get.runWithNewCluster(MiniRaftCluster.java:142)
	at org.apache.ratis.server.impl.MiniRaftCluster$Factory$Get.runWithNewCluster(MiniRaftCluster.java:125)
	at org.apache.ratis.RaftAsyncTests.runTestRequestAsyncWithRetryFailure(RaftAsyncTests.java:134)
	at org.apache.ratis.RaftAsyncTests.testRequestAsyncWithRetryFailureAfterInitialMessages(RaftAsyncTests.java:129)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
```

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
```diff
@@ -153,9 +168,14 @@ synchronized CompletableFuture<RaftNettyServerReplyProto> pollReply() {
     @Override
     public synchronized void close() {
       client.close();
+      failOutstandingRequests(new AlreadyClosedException("Closing connection to " + peer));
+    }
+
+    private void failOutstandingRequests(Throwable cause) {
```
OneSizeFitsQuorum: maybe we should synchronized for this method as it now maybe called by SimpleChannelInboundHandler, we need to ensure atomicity

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
```diff
@@ -121,6 +125,17 @@ protected void channelRead0(ChannelHandlerContext ctx,
...
+
+        @Override
+        public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+          failOutstandingRequests(new IOException("Caught an exception for the connection to " + peer, cause));
+          client.close();
+        }
+
+        @Override
+        public void channelInactive(ChannelHandlerContext ctx) {
+          failOutstandingRequests(new AlreadyClosedException("Channel to " + peer + " is inactive."));
```
OneSizeFitsQuorum, line 137: do we need to call super.channelInactive(ctx) here?

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
```diff
@@ -121,6 +125,17 @@ protected void channelRead0(ChannelHandlerContext ctx,
             future.complete(proto);
           }
         }
+
+        @Override
+        public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+          failOutstandingRequests(new IOException("Caught an exception for the connection to " + peer, cause));
```
OneSizeFitsQuorum: We could keep the same order with close() in line170-171

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
leixm: Should not be added. If super.channelInactive(ctx) is called here, it will cause infinite recursion because super.channelInactive will call ctx.fireChannelInactive();

leixm: Spark also has a similar implementation, https://github.com/apache/spark/blob/master/common/network-common/src/main/java/org/apache/spark/network/client/TransportResponseHandler.java#L144, which does not call super.channelInactive(ctx).

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
```diff
@@ -121,6 +125,17 @@ protected void channelRead0(ChannelHandlerContext ctx,
...
+          client.close();
```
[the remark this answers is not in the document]
leixm: ctx.close() and client.close() are equivalent and do not need to be called repeatedly.

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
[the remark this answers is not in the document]
szetszwo: Copilot seems to have a legitimate suggestion.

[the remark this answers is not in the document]
OneSizeFitsQuorum: @leixm said that spark did the same way in https://github.com/apache/spark/blob/master/common/network-common/src/main/java/org/apache/spark/network/client/TransportResponseHandler.java#L144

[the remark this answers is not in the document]
szetszwo: @leixm , @OneSizeFitsQuorum , I traced the code
super.channelInactive(ctx) is ChannelInboundHandlerAdapter.channelInactive(ctx)
-> ChannelHandlerContext..fireChannelInactive()
-> AbstractChannelHandlerContext.fireChannelInactive()
->  AbstractChannelHandlerContext.invokeChannelInactive(this.findContextInbound(16));
-> next.invokeChannelInactive()
So, Copilot seems correct.  It works only if there is no next handler.
```java
// AbstractChannelHandlerContext
    static void invokeChannelInactive(final AbstractChannelHandlerContext next) {
        EventExecutor executor = next.executor();
        if (executor.inEventLoop()) {
            next.invokeChannelInactive();
        } else {
            executor.execute(new Runnable() {
                @Override
                public void run() {
                    next.invokeChannelInactive();
                }
            });
        }
    }
```
https://github.com/netty/netty/blob/d9a8b00cb15e54a1f167025bbd9ab53d466a04d3/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java#L278-L290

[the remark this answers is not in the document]
szetszwo: Google AI says something similar:
```java
// Important: Call super.channelInactive(ctx) to propagate the event
// to the next handler in the ChannelPipeline.
super.channelInactive(ctx);
```
image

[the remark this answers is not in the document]
szetszwo: Filed https://issues.apache.org/jira/browse/SPARK-53761 ; see if they would change the code.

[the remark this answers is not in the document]
szetszwo: If the current code is working, I am fine if we are not adding `super.channelInactive(ctx)`.  We may add it later if there is a problem.

OneSizeFitsQuorum:
@leixm any updates?

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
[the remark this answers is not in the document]
leixm: I think @szetszwo  is right, I will add `super.channelInactive(ctx)`.

ratis-netty/src/main/java/org/apache/ratis/netty/NettyRpcProxy.java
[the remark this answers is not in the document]
leixm: client.close() 
-> AbstractChannel.close
-> DefaultChannelPipeline.close
-> AbstractChannelHandlerContext.close

ctx.close()
-> AbstractChannelHandlerContext.close

leixm:
Sorry for the late reply, I made some minor changes to the code, please help review.

szetszwo:
@leixm , thanks a lot for working on this!

@OneSizeFitsQuorum , thanks for reviewing this!

OneSizeFitsQuorum:
Well, it seems the ci become unstable after this pr's mergement(https://github.com/apache/ratis/actions/runs/18405493613/job/52498525024)

@szetszwo @leixm

leixm:
> Well, it seems the ci become unstable after this pr's mergement(https://github.com/apache/ratis/actions/runs/18405493613/job/52498525024)
> 
> @szetszwo @leixm

Let me check it.

leixm:
I rolled back this PR locally and ran testWithLoadAsync with netty, and it still failed occasionally. It seems to be a flaky test.
image

szetszwo:
> Well, it seems the ci become unstable after this pr's mergement(https://github.com/apache/ratis/actions/runs/18405493613/job/52498525024)

It failed with timeout.  Not sure if the machine is slow.  Let's try a 10x10 run to see if it can reproduce it.

szetszwo:
It has 3/100 failures.  All failed with timeout.  Let's try increasing the timeout.
- https://github.com/apache/ratis/actions/runs/18450620115/job/52563900327

OneSizeFitsQuorum:
@szetszwo Let's increase the timeout and  I will start releasing rc2~

szetszwo:
>  Let's try increasing the timeout.

Unfortunately, increasing the timeout does not help.  The failure seems to be related to the test but not the product code.  

The test keeps changing leader by blocking the old leader.  Observed below in the logs:
1. The requestVote from the old leader will not be delivered since it is blocked intentionally.
2. If the old leader has a higher index than the followers.  The old leader won't vote for the followers.
3. One of the follower is dead (not sure why).

So, both the old leader and and the remaining follower cannot be elected.  The test timed out if the above 3 conditions occur together.

> ...  will start releasing rc2~

~~The test problem seems there for a long time.  I agree that it should not block rc2.~~

@OneSizeFitsQuorum Just found that the follower has a deadlock for # 3 above.  We should fix it before rolling rc2.

szetszwo:
See RATIS-2345 for more details on the deadlock bug.  Of course, the deadlock bug is not related to this PR.

PR #1293, opened 2025-09-28
RATIS-2339. Avoid "commit does not belong to any branch" during release

OneSizeFitsQuorum:
see https://issues.apache.org/jira/browse/RATIS-2339

### DIFF
--- a/dev-support/make_rc.sh
+++ b/dev-support/make_rc.sh
@@ -145,7 +145,7 @@ mvnFun() {
 
 5-publish-git(){
   cd "$projectdir"
-  git push apache "release-${RATISVERSION}"
+  git push apache HEAD:"release-${RATISVERSION}"
   git push apache "ratis-${RATISVERSION}${RC}"
 }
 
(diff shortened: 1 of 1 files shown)


dev-support/make_rc.sh
```diff
@@ -145,6 +145,7 @@ mvnFun() {
 
 5-publish-git(){
   cd "$projectdir"
+  git push apache "release-${RATISVERSION}"
```
adoroszlai: I think this only works if your local branch is also named `release-${RATISVERSION}`.

```suggestion
  git push apache HEAD:"release-${RATISVERSION}"
```

adoroszlai (commented):
Thanks @OneSizeFitsQuorum for the patch.

dev-support/make_rc.sh
OneSizeFitsQuorum: Yes! Your suggesstion is more general!

OneSizeFitsQuorum:
@adoroszlai @szetszwo PTAL

adoroszlai (approved):
Thanks @OneSizeFitsQuorum for updating the patch, LGTM.

PR #1297, opened 2025-10-03
RATIS-2342. Remove NIO specification from GrpcStubPool.

symious:

In https://issues.apache.org/jira/browse/RATIS-2325, a GrpcStubPool was added, and it was specified to use NIOEventLoopGroup, but on server side it's using epoll by default, this ticket is to remove the NIO specification.

https://issues.apache.org/jira/browse/RATIS-2342

## How was this patch tested?

Local test, no more warning logs like "WARN org.apache.ratis.thirdparty.io.netty.bootstrap.Bootstrap: Unknown channel option 'org.apache.ratis.thirdparty.io.netty.channel.epoll.EpollChannelOption#TCP_USER_TIMEOUT' for channel '[id: 0xe291c830]'"


symious:
@szetszwo Thank you for the review.

PR #1339, opened 2026-02-03
RATIS-2385. Don't keep failed requests in the RetryCache.

ss77892:
If the leader has changed after we placed the request in the RetryCache but before we wrote it to the log file, we should remove it from the cache.

https://issues.apache.org/jira/browse/RATIS-2385
## How was this patch tested?
Manual tests with Ozone. Under load with a continuously changing leader, NPEs occur during retry operations. No exception occurs; the patch and retry operations complete successfully.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -772,9 +772,8 @@ class RaftServerImpl implements RaftServer.Division,
    */
   private CompletableFuture<RaftClientReply> checkLeaderState(RaftClientRequest request, CacheEntry entry) {
     if (!getInfo().isLeader()) {
-      NotLeaderException exception = generateNotLeaderException();
-      final RaftClientReply reply = newExceptionReply(request, exception);
-      return RetryCacheImpl.failWithReply(reply, entry);
+      return retryCache.failWithReplyAndInvalidate(
+          newExceptionReply(request, generateNotLeaderException()), entry);
     }
     if (!getInfo().isLeaderReady()) {
       final CacheEntry cacheEntry = retryCache.getIfPresent(ClientInvocationId.valueOf(request));
@@ -832,8 +831,8 @@ class RaftServerImpl implements RaftServer.Division,
 
     final LeaderStateImpl unsyncedLeaderState = role.getLeaderState().orElse(null);
     if (unsyncedLeaderState == null) {
-      final RaftClientReply reply = newExceptionReply(request, generateNotLeaderException());
-      return RetryCacheImpl.failWithReply(reply, cacheEntry);
+      return retryCache.failWithReplyAndInvalidate(
+          newExceptionReply(request, generateNotLeaderException()), cacheEntry);
     }
     final PendingRequests.Permit unsyncedPermit = unsyncedLeaderState.tryAcquirePendingRequest(request.getMessage());
     if (unsyncedPermit == null) {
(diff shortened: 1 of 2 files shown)


szetszwo (commented):
@ss77892 , thanks a lot for working on this!

You are right that we should clean up the cache entries for NotLeaderException.
- Indeed, we should also clean up all the cache entries for non-StateMachineException.
- BTW, the current code does not check the return value from CompletableFuture.complete(..)/completeExceptionally(..).

Could you also fix the problems above?

See also https://issues.apache.org/jira/secure/attachment/13080683/1339_review.patch

szetszwo:
@ss77892 , on second thought, we should not invalidate any cache entries.  Ratis retry cache is designed to have unique cache entries.
- If a call has replied normally or exceptionally, the calls with the same `ClientInvocationId` should always receive the same result.
- For retires, the client always uses a different `ClientInvocationId`.

Such design makes thing simple since all the replies are always fixed.

In contrast, if we invalidate a NotLeaderException entry.  It could cause problem like below:

1. Client sends a call.
2. Server replies NotLeaderException and invalidates the entry.
3. Client has not received the NotLeaderException reply (due to slow network).  It retries with the same ClientInvocationId.
4. Server replies replies again with a different result.
5. Client receives the second result.
6. Client receives the NotLeaderException

ss77892:
@szetszwo We have another leadership check at the beginning, so we don't store the entry in the cache if we are not the leader at the moment. So, it would behave exactly in the same way even without the patch. The patch removes an inconsistency in behavior that might occur when leadership changes during the event processing (the node was the leader when it received the event, but it's not a leader when it wants to add the event to the log).

szetszwo:
> ... We have another leadership check at the beginning,  ...

Could you point out the code?

ss77892:
https://github.com/apache/ratis/blob/b940b65004bea5f15593e2bbcb5ce925c0580b39/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java#L991
Here we call checkLeaderState that would throw an exception if the node is not the leader. To the cache, we add it 5 lines later.

ivandika3:
> on second thought, we should not invalidate any cache entries. Ratis retry cache is designed to have unique cache entries.

+1 on this, we should not invalidate cache entries even if there is an exception. This will break the idempotency guarantee for a single unique `ClientInvocationId` (`ClientId` + invocation ID or callId) .

If a client wants to retry the same operation after they receive an exception, they should use higher a CallId.

ss77892:
@ivandika3 @szetszwo 
Please check this.
https://github.com/apache/ratis/blob/b940b65004bea5f15593e2bbcb5ce925c0580b39/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java#L991
We fail with an exception at the beginning of  writeAsyncImpl before the request lands to RetryCache. So, there is no idempotency guarantee. Also the client might go to another node that is the leader now with the same callId and that node doesn't have it in the RetryCache.

szetszwo:
@ss77892 , thanks for pointing the code.

HDDS-13621 only has the exception but not the details.  Could you explain how invalidating the cache entry could solve the problem in Ozone?  Why Ozone cannot retry with a new callId?

ss77892:
@szetszwo The workflow was following:
1. client send the request and it successfully pass the first check and lands to the RetryCache. 
2. leadership changed
3. before landing the request to the log we find that we are not leader anymore and fail the request
4. OM wants to retry the request but it checks the RetryCache directly and tries to get the response. Since the request failed with an exception it fails with the NPE.

HDDS-13621 adds the check that request from the cache is successful. If it's not the regular logic with retry will be triggered. 
So, from Ozone perspective the issue is already resolved. In other hand Ratis behavior is not consistent. Under similar circumstances (the leadership has been changed) the request might or might not be in the RetryCache, depending on the time when leadership has been changed.

szetszwo:
@ss77892 ,  I agree that it is inconsistent for NotLeaderException for different cases.  A consistent approach is to add all the  calls to the RetryCache.  However, it adds overhead but not really solving any problems.

if Ozone has already been fixed by [HDDS-13621](https://issues.apache.org/jira/browse/HDDS-13621), how about not changing Ratis?

PR #1363, opened 2026-03-06
RATIS-2421. Gracefully cancel stream after complete in GrpcLogAppender

symious:

When onCompleted() is called, the client waits for the server response to finish the RPC and release the underlying resources.

However, in some cases the server response may never arrive, which leaves the RPC stream open and prevents resources from being released.

This ticket introduces a client-side cancel() after a short grace period following onCompleted(), to ensure the RPC stream is forcefully closed and resources are cleaned up.

https://issues.apache.org/jira/browse/RATIS-2421

## How was this patch tested?

Test locally.

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -360,7 +360,8 @@ public class GrpcLogAppender extends LogAppenderBase {
     StreamObservers(GrpcServerProtocolClient client, AppendLogResponseHandler handler, boolean separateHeartbeat,
         TimeDuration waitTimeMin, TimeDuration completeGracePeriod) {
       this.appendLog = (ClientCallStreamObserver<AppendEntriesRequestProto>) client.appendEntries(handler, false);
-      this.heartbeat = separateHeartbeat? (ClientCallStreamObserver<AppendEntriesRequestProto>) client.appendEntries(handler, true): null;
+      this.heartbeat = separateHeartbeat?
+          (ClientCallStreamObserver<AppendEntriesRequestProto>) client.appendEntries(handler, true): null;
       this.waitForReady = waitTimeMin.isPositive()? waitTimeMin: TimeDuration.ONE_MILLISECOND;
       this.completeGracePeriod = completeGracePeriod.isPositive()? completeGracePeriod : TimeDuration.ONE_SECOND;
     }
@@ -485,7 +486,8 @@ public class GrpcLogAppender extends LogAppenderBase {
       increaseNextIndex(pending);
       if (appendLogRequestObserver == null) {
         appendLogRequestObserver = new StreamObservers(
-            getClient(), new AppendLogResponseHandler(), useSeparateHBChannel, getWaitTimeMin(), getCompleteGracePeriod());
+            getClient(), new AppendLogResponseHandler(), useSeparateHBChannel, getWaitTimeMin(),
+            getCompleteGracePeriod());
       }
     }
 
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -35,6 +35,7 @@ import org.apache.ratis.server.raftlog.RaftLog;
 import org.apache.ratis.server.util.ServerStringUtils;
 import org.apache.ratis.thirdparty.io.grpc.StatusRuntimeException;
 import org.apache.ratis.thirdparty.io.grpc.stub.CallStreamObserver;
+import org.apache.ratis.thirdparty.io.grpc.stub.ClientCallStreamObserver;
 import org.apache.ratis.thirdparty.io.grpc.stub.StreamObserver;
 import org.apache.ratis.proto.RaftProtos.AppendEntriesReplyProto;
 import org.apache.ratis.proto.RaftProtos.AppendEntriesReplyProto.AppendResult;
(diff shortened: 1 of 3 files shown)


ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -57,10 +58,8 @@
 import java.util.Optional;
 import java.util.Queue;
 import java.util.UUID;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
spacemonkd: Is a wildcard import required? I would prefer if we can import specific classes only.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -341,20 +340,37 @@ private boolean haveTooManyPendingRequests() {
...
     private volatile boolean running = true;
 
+    private final ScheduledExecutorService closer =
+        Executors.newSingleThreadScheduledExecutor(r -> {
+          Thread t = new Thread(r, "grpc-log-appender-stream-closer");
+          t.setDaemon(true);
+          return t;
+        });
+
+//    private final AtomicBoolean closing = new AtomicBoolean(false);
```
spacemonkd: nit: can be removed?

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -366,16 +382,73 @@ void onNext(AppendEntriesRequestProto proto)
...
     void onCompleted() {
-      appendLog.onCompleted();
-      Optional.ofNullable(heartbeat).ifPresent(StreamObserver::onCompleted);
+      if (completed.compareAndSet(false, true)) {
+        completeStreamGracefully(appendLog, "appendLog");
+        Optional.ofNullable(heartbeat)
+            .ifPresent(s -> completeStreamGracefully(s, "heartbeat"));
+      }
+      final long delayMs = Math.max(1L, completeGracePeriod.toLong(TimeUnit.MILLISECONDS));
+      closer.schedule(this::cancelIfStillNeeded, delayMs, TimeUnit.MILLISECONDS);
```
spacemonkd, line 410: This schedules `cancelIfStillNeeded()` outside the `completed.compareAndSet(...)` block. Repeated onCompleted() calls can queue redundant cancel tasks.

Can this be avoided?
Also I think if it’s invoked again after `shutdownCloser()` it may throw exception.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -366,16 +382,73 @@ void onNext(AppendEntriesRequestProto proto)
...
+        String name) {
+      try {
+        stream.onCompleted();
+      } catch (Exception e) {
+        LOG.warn("Failed to call onCompleted on {}", name, e);
+      }
+    }
+
+    private void cancelStream(
+        ClientCallStreamObserver<AppendEntriesRequestProto> stream,
```
spacemonkd, line 443: We are doing explicit cast to `ClientCallStreamObserver<>`, but I am thinking can there be a case where `appendEntries()` can return only `CallStreamObserver`?
That is the method signature in GrpcServerProtocolClient.

In such a case this can throw error.
Better to accept CallStreamObserver<> and then do an instanceof check?

```
try {
    if (stream instanceof ClientCallStreamObserver) {
      ((ClientCallStreamObserver<AppendEntriesRequestProto>) stream).cancel(reason, cause);
    } else {
      ...handle cancel when it is not clientcallstreamobserver, something like stream.onError(...)?
    }
  } catch (Exception e) {
    LOG.warn("Failed to cancel {}", name, e);
  }
```

spacemonkd (commented):
Thanks for the patch @symious.
Just a few comments, I am not too sure in some cases so definitely another set of eyes would be good to have.
But apart from these it looks good to me

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
symious: Fixed.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
symious: Fixed.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
symious: The cancel and shutdown parts can be called multiple times.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
symious: As a client, this one should always be a ClientCallStreamObserver.

symious:
@devabhishekpal Thank you for the review. Updated and PTAL.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
spacemonkd: But if it is completed already aren't subsequent cancel calls queuing tasks which are no longer required?

symious: If onComplete() is invoked, the scheudler should be called and shutdown later, later requests should be rejected then.

spacemonkd (approved):
Thanks for addressing the comments @symious. Looks good to me, +1

spacemonkd:
@szetszwo could you take a look at this PR?

spacemonkd:
@szetszwo this looks good to me, should we merge it? Your review would be great

PR #1368, opened 2026-03-09
RATIS-2426. Fix memory leak in ServerRequestStreamObserver

ivandika3:

Please refer to https://issues.apache.org/jira/browse/RATIS-2426

https://issues.apache.org/jira/browse/RATIS-2426

## How was this patch tested?

UT


ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
```diff
@@ -366,16 +374,46 @@ void onNext(AppendEntriesRequestProto proto)
...
       stream.onNext(proto);
     }
 
-    void stop() {
+    void stop(Event event) {
       running = false;
+      if (event == Event.COMPLETE) {
+        onCompleted();
+      } else {
+        cancelStream("stop due to " + event);
```
szetszwo, line 388: We have the following call sequence:
- AppendLogResponseHandler.onError(..) / onCompleted() -> resetClient(..) -> stop()

In the [gRPC example](https://grpc.io/docs/languages/java/basics/#client-side-streaming-rpc),
- onCompleted() implementation will response onCompleted() but
- onError(Throwable t) only calls logger.log(Level.WARNING, ...)

So, I think stop() should not call cancelStream(..)

ratis-test/src/test/java/org/apache/ratis/grpc/TestLogAppenderWithGrpc.java
```diff
@@ -148,4 +159,207 @@ private void runTestRestartLogAppender(MiniRaftClusterWithGrpc cluster) throws E
...
   }
+
+  /**
+   * Verify that old LogAppender instances are properly cleaned up (gRPC streams terminated)
+   * after restartLogAppenders. Without the fix, gRPC holds references to unterminated
+   * stream response handlers, preventing old LogAppender instances from being collected.
+   */
+  @ParameterizedTest
+  @MethodSource("data")
+  public void testLogAppenderStreamCleanupOnRestart(Boolean separateHeartbeat) throws Exception {
```
szetszwo, line 170: I tried to revert GrpcLogAppender and GrpcServerProtocolService (except for GRPC_SERVER_HANDLE_ERROR).  Both testLogAppenderStreamCleanupOnRestart and testFollowerHandleErrorCleanup can pass.  So, they probably cannot test the code change.

szetszwo (commented):
@ivandika3 , thanks for working on this!  Please see the comment inlined. (Sorry that I missed this PR earlier.)

ivandika3:
@szetszwo Thanks for the review. I'm still not 100% sure about this fix yet. I'll revisit this when I have time.

PR #1372, opened 2026-03-09
RATIS-2430. Write snapshot to temporary path until finish

spacemonkd:

Today when [SnapshotInstallatioHandler#checkAndInstallSnapshot](https://github.com/apache/ratis/blob/1433b4cbf7350afcf9b2f871dd48c48e17a91a1d/ratis-server/src/main/java/org/apache/ratis/server/impl/SnapshotInstallationHandler.java#L167) calls `state.installSnapshot(request)` - it pauses the state machine via `ServerState.installSnapshot()`.

However this means that in case later checks fail or IO fails or any such scenario occurs, then there is no clear rollback option. Followers in this scenario can be left in a partial installation state.

One way to mitigate this is in appendChunk we can write to a temp file without pausing StateMachine. When this is done we can atomically apply the snapshot and reload the statemachine log.

https://issues.apache.org/jira/browse/RATIS-2430

## How was this patch tested?
Patch was tested using unit tests.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java
@@ -470,11 +470,16 @@ class ServerState {
     return raftStorage.getUnchecked();
   }
 
-  void installSnapshot(InstallSnapshotRequestProto request) throws IOException {
+  void appendSnapshot(InstallSnapshotRequestProto request) throws IOException {
     // TODO: verify that we need to install the snapshot
-    StateMachine sm = server.getStateMachine();
-    sm.pause(); // pause the SM to prepare for install snapshot
-    snapshotManager.installSnapshot(request, sm);
+    snapshotManager.appendSnapshot(request, server.getStateMachine());
+  }
+
+  void finalizeSnapshot(InstallSnapshotRequestProto request) throws IOException {
+    final StateMachine sm = server.getStateMachine();
+    sm.pause(); // pause the SM right before publishing the snapshot atomically
+    // TODO: if there is a failure here, we need to rollback the snapshot installation.
+    snapshotManager.finalizeSnapshot(request);
   }
 
   private SnapshotInfo getLatestSnapshot() {
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java
@@ -31,7 +31,6 @@ import org.apache.ratis.server.raftlog.RaftLog;
 import org.apache.ratis.server.raftlog.memory.MemoryRaftLog;
 import org.apache.ratis.server.raftlog.segmented.SegmentedRaftLog;
 import org.apache.ratis.server.storage.*;
-import org.apache.ratis.proto.RaftProtos.InstallSnapshotRequestProto;
 import org.apache.ratis.proto.RaftProtos.LogEntryProto;
 import org.apache.ratis.statemachine.SnapshotInfo;
 import org.apache.ratis.statemachine.StateMachine;
@@ -69,7 +68,6 @@ class ServerState {
   private final MemoizedSupplier<StateMachineUpdater> stateMachineUpdater;
   /** local storage for log and snapshot */
   private final MemoizedCheckedSupplier<RaftStorageImpl, IOException> raftStorage;
-  private final SnapshotManager snapshotManager;
   private final AtomicReference<Timestamp> lastNoLeaderTime;
   private final TimeDuration noLeaderTimeout;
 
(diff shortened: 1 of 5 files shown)


spacemonkd:
@szetszwo could you take a look at this change?

ratis-server/src/main/java/org/apache/ratis/server/impl/ServerState.java
```diff
@@ -470,11 +470,16 @@ RaftStorageImpl getStorage() {
...
-    StateMachine sm = server.getStateMachine();
-    sm.pause(); // pause the SM to prepare for install snapshot
-    snapshotManager.installSnapshot(request, sm);
+    snapshotManager.appendSnapshot(request, server.getStateMachine());
+  }
+
+  void finalizeSnapshot(InstallSnapshotRequestProto request) throws IOException {
+    final StateMachine sm = server.getStateMachine();
+    sm.pause(); // pause the SM right before publishing the snapshot atomically
+    // TODO: if there is a failure here, we need to rollback the snapshot installation.
```
szetszwo: What do you mean by "rollback the snapshot installation" ?

ratis-server/src/main/java/org/apache/ratis/server/impl/SnapshotInstallationHandler.java
```diff
@@ -214,17 +214,18 @@ private CompletableFuture<InstallSnapshotReplyProto> checkAndInstallSnapshot(Ins
...
         }
+        // Append chunks to a temporary location first. Publish only when done=true.
+        state.appendSnapshot(request);
+        nextChunkIndex.incrementAndGet();
         // update the committed index
         // re-load the state machine if this is the last chunk
         if (snapshotChunkRequest.getDone()) {
+          state.finalizeSnapshot(request);
           state.reloadStateMachine(lastIncluded);
           chunk0CallId.set(-1);
```
szetszwo, line 239: Let's move snapshotManager to SnapshotInstallationHandler. It is only used here.
```java
        final int expectedChunkIndex = nextChunkIndex.get();
        if (expectedChunkIndex != snapshotChunkRequest.getRequestIndex()) {
          throw new IOException("Unexpected request chunk index: " + snapshotChunkRequest.getRequestIndex()
              + " (the expected index is " + expectedChunkIndex + ")");
        }
        // Append chunks to a temporary location first. Publish only when done=true.
        final StateMachine stateMachine = server.getStateMachine();
        snapshotManager.appendSnapshot(request, stateMachine);
        nextChunkIndex.incrementAndGet();
        // update the committed index
        // re-load the state machine if this is the last chunk
          stateMachine.pause(); // pause the SM right before publishing the snapshot atomically
          snapshotManager.finalizeSnapshot(request);

          state.reloadStateMachine(lastIncluded);
          chunk0CallId.set(-1);
        }
```

ratis-server/src/main/java/org/apache/ratis/server/storage/SnapshotManager.java
```diff
@@ -164,9 +168,16 @@ public void installSnapshot(InstallSnapshotRequestProto request, StateMachine st
...
-    if (snapshotChunkRequest.getDone()) {
-      rename(tmpDir, snapshotDir.get());
+  }
+
+  public void finalizeSnapshot(InstallSnapshotRequestProto request) throws IOException {
+    final InstallSnapshotRequestProto.SnapshotChunkProto snapshotChunkRequest = request.getSnapshotChunk();
+    if (!snapshotChunkRequest.getDone()) {
+      throw new IOException("Cannot finalize incomplete snapshot request: "
+          + ServerStringUtils.toInstallSnapshotRequestString(request));
     }
```
szetszwo: Let's use Preconditions:
```java
    Preconditions.assertTrue(snapshotChunkRequest.getDone());
```

ratis-server/src/test/java/org/apache/ratis/server/storage/SnapshotManagerTest.java
```diff
@@ -0,0 +1,174 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.security.MessageDigest;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class SnapshotManagerTest {
```
szetszwo: To be consistent with the other tests, 
- rename it to "TestSnapshotManager"
- move it to ratis-test.

ratis-server/src/test/java/org/apache/ratis/server/storage/SnapshotManagerTest.java
```diff
@@ -0,0 +1,174 @@
...
+  private static final class TestRaftStorageDirectory implements RaftStorageDirectory {
```
szetszwo: Let's use RaftStorageDirectoryImpl.

szetszwo (commented):
@spacemonkd , thanks for working on this!  Please see the comments inlined and also https://issues.apache.org/jira/secure/attachment/13082753/1372_review.patch

(Sorry that I missed this PR.)

spacemonkd:
Unit test failure is flaky. CI has passed on fork: https://github.com/spacemonkd/ratis/actions/runs/27395469744

@szetszwo could you take another look?

szetszwo (approved):
+1 the change looks good.

PR #1428, opened 2026-04-15
RATIS-2398. Add opentelemetry-javaagent to ratis-examples and assembly

taklwu:
also fixed a bug that client need to issue the parent span in the client request

- make an example of how to use tracing in ratis-examples
- make sure the script could be automatically running and span shows in the logs and Jaegar UI
- fix the parent span issued from the client.

[RATIS-2389](https://issues.apache.org/jira/browse/RATIS-2389)

## How was this patch tested?
manual tests for running the https://github.com/apache/ratis/tree/master/ratis-examples locally, tested with below

1. logger mode that emits span in logs

```
# on the server 
[otel.javaagent 2026-04-15 15:40:24:419 -0700] [pool-3-thread-1] INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'raft.server.submitClientRequestAsync' : 2a9bfd5258de404d4c8dcd291408538f 98677058d4b31a8e SERVER [tracer: org.apache.ratis:3.3.0-SNAPSHOT] AttributesMap{data={thread.id=69, thread.name=n0-client-thread1, raft.member.id=n0@group-6F7570313233, raft.call.id=51198, raft.client.id=client-5813AB26C91D}, capacity=128, totalAddedValues=5}
[otel.javaagent 2026-04-15 15:40:24:419 -0700] [pool-3-thread-2] INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'raft.server.submitClientRequestAsync' : f56c5fe25544695b023cfcdcd4f32447 57fd34b224ee3d6a SERVER [tracer: org.apache.ratis:3.3.0-SNAPSHOT] AttributesMap{data={thread.id=69, thread.name=n0-client-thread1, raft.member.id=n0@group-6F7570313233, raft.call.id=51196, raft.client.id=client-5813AB26C91D}, capacity=128, totalAddedValues=5}

# from the client 
[otel.javaagent 2026-04-15 15:40:24:420 -0700] [grpc-default-executor-7] INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'Async::send' : 3326eebb2357274628d3dea57e586435 2715719eb0c38ec3 CLIENT [tracer: org.apache.ratis:3.3.0-SNAPSHOT] AttributesMap{data={thread.id=37, raft.peer.id=LEADER, thread.name=pool-1-thread-20, raft.operation.type=RW, raft.operation.name=Async::send}, capacity=128, totalAddedValues=5}
```

2. export traces to the third party otlp e.g. Jaegar 

Screenshot 2026-04-15 at 3 39 32 PM

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
@@ -26,6 +26,7 @@ import org.apache.ratis.client.api.SnapshotManagementApi;
 import org.apache.ratis.client.retry.ClientRetryEvent;
 import org.apache.ratis.conf.Parameters;
 import org.apache.ratis.conf.RaftProperties;
+import io.opentelemetry.context.Context;
 import org.apache.ratis.proto.RaftProtos.SlidingWindowEntry;
 import org.apache.ratis.protocol.ClientId;
 import org.apache.ratis.protocol.Message;
@@ -290,6 +291,9 @@ public final class RaftClientImpl implements RaftClient {
     } else {
       b.setLeaderId(getLeaderId())
        .setRepliedCallIds(repliedCallIds.get(callId));
+    }
+    if (TraceUtils.getGlobalTracer() != null) {
+      b.setSpanContext(TraceUtils.injectContextToProto(Context.current()));
     }
     return b.setClientId(clientId)
         .setGroupId(groupId)
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
@@ -292,6 +292,9 @@ public final class RaftClientImpl implements RaftClient {
       b.setLeaderId(getLeaderId())
        .setRepliedCallIds(repliedCallIds.get(callId));
     }
+    if (TraceUtils.isEnabled()) {
+      b.setSpanContext(TraceUtils.injectContextToProto());
+    }
     return b.setClientId(clientId)
         .setGroupId(groupId)
         .setCallId(callId)
--- a/ratis-common/src/main/java/org/apache/ratis/trace/TraceConfigKeys.java
+++ b/ratis-common/src/main/java/org/apache/ratis/trace/TraceConfigKeys.java
@@ -18,6 +18,7 @@
 package org.apache.ratis.trace;
 
 import org.apache.ratis.conf.RaftProperties;
+import org.apache.ratis.util.StringUtils;
 
 import java.util.function.Consumer;
 
@@ -30,8 +31,20 @@ public interface TraceConfigKeys {
   String ENABLED_KEY = PREFIX + ".enabled";
   boolean ENABLED_DEFAULT = false;
 
+  /**
+   * Whether Ratis should emit OpenTelemetry spans. When the key is absent from
+   * {@link RaftProperties}, the JVM system property {@value #ENABLED_KEY} is consulted so
+   * {@code -Draft.otel.tracing.enabled=true} works with empty example configs.
+   */
   static boolean enabled(RaftProperties properties, Consumer<String> logger) {
-    return getBoolean(properties::getBoolean, ENABLED_KEY, ENABLED_DEFAULT, logger);
+    if (properties.getRaw(ENABLED_KEY) != null) {
+      return getBoolean(properties::getBoolean, ENABLED_KEY, ENABLED_DEFAULT, logger);
+    }
+    final String fromSystem = System.getProperty(ENABLED_KEY);
+    if (fromSystem != null) {
+      return StringUtils.string2boolean(fromSystem.trim(), ENABLED_DEFAULT);
+    }
+    return ENABLED_DEFAULT;
   }
 
   static boolean enabled(RaftProperties properties) {
--- a/ratis-examples/src/main/bin/client.sh
+++ b/ratis-examples/src/main/bin/client.sh
@@ -23,7 +23,8 @@ if [ "$#" -lt 2 ]; then
   exit 1
 fi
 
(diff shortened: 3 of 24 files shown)


taklwu:
@szetszwo @OneSizeFitsQuorum any comments?

ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.ratis.client.retry.ClientRetryEvent;
 import org.apache.ratis.conf.Parameters;
 import org.apache.ratis.conf.RaftProperties;
+import io.opentelemetry.context.Context;
```
szetszwo: We should not import io.opentelemetry in raft-client.

szetszwo (commented):
@taklwu , thanks for working on this!   Just have a quick comment inlined.

Let me try to run an example with tracing.

ratis-client/src/main/java/org/apache/ratis/client/impl/RaftClientImpl.java
taklwu: removed and make the call of Context.current() within TraceUtils

taklwu:
those failure are not related, but let's see if we can retrigger...

org.apache.ratis.grpc.TestLinearizableReadWithGrpc
org.apache.ratis.grpc.TestLinearizableReadAppliedIndexLeaderLeaseReadWithGrpc
org.apache.ratis.netty.TestRaftAsyncWithNetty (this one keeps failing now)

taklwu:
@adoroszlai @szetszwo  the PR github tests looks green now, do you have any other comments?

ratis-common/src/main/java/org/apache/ratis/trace/TraceConfigKeys.java
```diff
@@ -30,8 +31,20 @@ public interface TraceConfigKeys {
...
+   * Whether Ratis should emit OpenTelemetry spans. When the key is absent from
+   * {@link RaftProperties}, the JVM system property {@value #ENABLED_KEY} is consulted so
+   * {@code -Draft.otel.tracing.enabled=true} works with empty example configs.
+   */
   static boolean enabled(RaftProperties properties, Consumer<String> logger) {
-    return getBoolean(properties::getBoolean, ENABLED_KEY, ENABLED_DEFAULT, logger);
+    if (properties.getRaw(ENABLED_KEY) != null) {
+      return getBoolean(properties::getBoolean, ENABLED_KEY, ENABLED_DEFAULT, logger);
+    }
+    final String fromSystem = System.getProperty(ENABLED_KEY);
```
szetszwo: Let's don't fallback to JVM system properties.

Ratis currently do not handle how the confs are stored (e.g. in a file, in a DB, in an external service, or JVM system properties).  The application uses Ratis has to decide how to do it and pass it to Ratis via a RaftProperties object.

Also, using JVM system properties could cause problem for applications using Ratis for multiple components -- e.g. there could be two different services running in the same JVM using Ratis.  They may use different confs.

ratis-examples/src/main/bin/client.sh
```diff
@@ -32,4 +33,4 @@ shift
 subcommand="$1"
 shift
 
-java ${LOGGER_OPTS} -jar $ARTIFACT "$example" "$subcommand" "$@"
+java ${OTEL_OPTS} ${LOGGER_OPTS} -jar $ARTIFACT "$example" "$subcommand" "$@"
```
szetszwo, line 36: For the example command, we could change it to get the JVM property and then set it to RaftProperties.

One idea is to add a new ExampleLuancher class, which will read JVM property and then pass it to the examples  (e.g. "filestore" or "arithmetic").

szetszwo (commented):
@taklwu , thanks for working on this!  Please see the comments inlined.

(Sorry for the late reply.)

ratis-common/src/main/java/org/apache/ratis/trace/TraceConfigKeys.java
taklwu: sorry for coming back on this. 

it's fine to remove JVM or system properties.

When we want to use this tracing with Ozone, does it mean we need to create a special keys in Ozone e.g. `hdds.ratis.client.*` or `hdds.ratis.server.*` such that the open telemetry could be enabled?

szetszwo: For Ozone, we may add a prefix to set any Ratis conf; see https://github.com/apache/ozone/blob/master/hadoop-hdds/docs/content/feature/multi-raft-support.md#advanced-ratis-configuration

szetszwo:
@taklwu , please wait for RATIS-2558, which will  move OpenTelemetry code and its dependency to a separated module. 

Or, you may update this with https://github.com/apache/ratis/compare/master...szetszwo:ratis:RATIS-2558

taklwu:
> @taklwu , please wait for [RATIS-2558](https://issues.apache.org/jira/browse/RATIS-2558), which will move OpenTelemetry code and its dependency to a separated module.
> 
> Or, you may update this with [master...szetszwo:ratis:RATIS-2558](https://github.com/apache/ratis/compare/master...szetszwo:ratis:RATIS-2558)

got it, I will have another update once RATIS-2558 completes

ratis-examples/src/main/bin/client.sh
taklwu: added [ExampleLauncher.java](https://github.com/apache/ratis/pull/1428/changes#diff-e2d589ee6a85aab22b3ed47dd6292d9857035feb1da1bb99cc7b87fb20285ae7)

spacemonkd:
@szetszwo is this PR good to merge?

szetszwo (approved):
+1 the change looks good.

PR #1433, opened 2026-04-18
RATIS-2497. Implement dummy request as a noop request instead of watch request

ivandika3:

Currently the dummy request is a WATCH(0) which will trigger client leader failover. In a follower read client scenario, this might cause client to always failover to the leader which reduces the follower read effectiveness.

We can try implement a new noop request (some kind of ping request) which simply returns a successful response regardless of the Raft peer role.

https://issues.apache.org/jira/browse/RATIS-2497

## How was this patch tested?

UT.

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/ClientProtoUtils.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/ClientProtoUtils.java
@@ -144,6 +144,8 @@ public interface ClientProtoUtils {
         return RaftClientRequest.Type.valueOf(p.getDataStream());
       case FORWARD:
         return RaftClientRequest.Type.valueOf(p.getForward());
+      case NOOP:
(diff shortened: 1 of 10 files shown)


ivandika3:
Closing this, I think it's better to address this TODO.

```
// TODO: this is a work around, it is better to fix the underlying RPC implementation
```

szetszwo:
I agree it is better to fix the underlying RPC implementation but it may not be easy.  😅

How about passing server to the dummy watch request? See https://github.com/apache/ratis/pull/1438

ivandika3:
> I agree it is better to fix the underlying RPC implementation but it may not be easy. 😅

Yes I thought about this yesterday and it probably requires a breaking change on the underlying protocol.

> How about passing server to the dummy watch request? See https://github.com/apache/ratis/pull/1438

Thanks for the idea. I think this might be better. Let me check this.

PR #1438, opened 2026-04-21
RATIS-2497. Pass server to the dummy watch request in OrderedAsync

szetszwo:
See RATIS-2497

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/AsyncImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/AsyncImpl.java
@@ -40,7 +40,7 @@ class AsyncImpl implements AsyncRpcApi {
   CompletableFuture<RaftClientReply> send(
       RaftClientRequest.Type type, Message message, RaftPeerId server) {
     return TraceClient.asyncSend(
-        () -> client.getOrderedAsync().send(type, message, server), type, server);
+        () -> client.getOrderedAsync(server).send(type, message, server), type, server);
   }
 
   @Override
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/OrderedAsync.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/OrderedAsync.java
@@ -116,12 +116,12 @@ public final class OrderedAsync {
     }
   }
 
-  static OrderedAsync newInstance(RaftClientImpl client, RaftProperties properties) {
+  static OrderedAsync newInstance(RaftClientImpl client, RaftPeerId server, RaftProperties properties) {
     final OrderedAsync ordered = new OrderedAsync(client, properties);
     // send a dummy watch request to establish the connection
     // TODO: this is a work around, it is better to fix the underlying RPC implementation
     if (RaftClientConfigKeys.Async.Experimental.sendDummyRequest(properties)) {
-      ordered.send(RaftClientRequest.watchRequestType(), null, null);
+      ordered.send(RaftClientRequest.watchRequestType(), null, server);
     }
     return ordered;
   }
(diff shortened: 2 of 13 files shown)


ratis-client/src/main/java/org/apache/ratis/client/impl/AsyncImpl.java
```diff
@@ -40,7 +40,7 @@ class AsyncImpl implements AsyncRpcApi {
   CompletableFuture<RaftClientReply> send(
       RaftClientRequest.Type type, Message message, RaftPeerId server) {
     return TraceClient.asyncSend(
-        () -> client.getOrderedAsync().send(type, message, server), type, server);
+        () -> client.getOrderedAsync(server).send(type, message, server), type, server);
```
ivandika3, line 43: Currently, if the `server` is not a leader, it will throw `NotLeaderException` since `watchAsync` calls `checkLeaderState`. This might cause `AsyncStreamObservers#completeReplyExceptionally` which will trigger client to invalidate all the pending replies.

One idea is to allow follower to accept the watch request if the RaftClientRequest serverId matches, but currently `WatchRequests` is stored in the `LeaderStateImpl`. So we might want to make `WATCH(0)` to be a special "bootstrap" request that can be handled by follower, otherwise the watch should throw `NotLeaderException` to preserve the watch semantics. 

Also, could you help to add some tests to test to behavior similar to https://github.com/apache/ratis/pull/1433? Set the dummy request to true for `LinearizableReadTests` should be enough.

ivandika3 (commented):
Thanks @szetszwo for the patch. Left one comment.

ratis-client/src/main/java/org/apache/ratis/client/impl/AsyncImpl.java
szetszwo: > ... watchAsync calls checkLeaderState ...

You are right -- we should use staleRead instead.

szetszwo: > ... Set the dummy request to true for LinearizableReadTests should be enough.

Sure.  Let me include the change.

szetszwo:
> Currently, if the server is not a leader, it will throw NotLeaderException ...

> ... we should use staleRead instead.

@ivandika3 , Tried to set a special minIndex (e.g. Long.MAX_VALUE/MIN_VALUE) to staleRead but it became an incompatible change since users might also set it.

Then, found that watch request passed a null message and users could not set it.  So, we may send a watch request with a "DUMMY" message.  The server can detect it and return immediately without leader check.

LinearizableReadTests can pass.

ivandika3:
@szetszwo Thanks for the patch. LGTM +1.

> Tried to set a special minIndex (e.g. Long.MAX_VALUE/MIN_VALUE) to staleRead but it became an incompatible change since users might also set it.

Yes, we should not use `staleRead` since the behavior might be coupled with `StateMachine#queryStale` which can cause unpredictability.

> Then, found that watch request passed a null message and users could not set it. So, we may send a watch request with a "DUMMY" message. The server can detect it and return immediately without leader check.

This is a good workaround.

> LinearizableReadTests can pass.

Thanks, do you think we should include tests like `runTestFollowerLinearizableReadStaysOnFollower` in https://github.com/apache/ratis/pull/1433 to be more explicit? I'm OK if you choose to leave it be.

ivandika3:
@OneSizeFitsQuorum Would you like to take a look?

ivandika3 (approved):

OneSizeFitsQuorum:
LGTM~
The DUMMY request only establishes the RPC connection and bypasses leader state checks, while real read requests still follow the full linearizability path. 
I think this is a good workaround!

OneSizeFitsQuorum (approved):

szetszwo:
@ivandika3 , @OneSizeFitsQuorum , thanks for reviewing this!
- Checked the test that the read-from-follower client only talked to a that follower and there were no failovers.
- This is a compatible change:
  - Old client -> New server:  the old client sends a normal watch request to the new server.  It will processed normally.
  - New client -> Old server: the new client sends a DUMMY watch request to the old server.  The DUMMY message will be ignored  and the watch request will be processed as a normal watch request.

PR #1439, opened 2026-04-24
RATIS-2507. Fix java.lang.IllegalStateException: gap between entries

ChenSammi:
Fail RatisServer when RaftLog end log index is smaller than last snapshot index during startup

https://issues.apache.org/jira/browse/RATIS-2507

## How was this patch tested?

existing unit test

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLog.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLog.java
@@ -430,6 +430,7 @@ public final class SegmentedRaftLog extends RaftLogBase {
       // If the entry has state machine data, then the entry should be inserted
       // to statemachine first and then to the cache. Not following the order
       // will leave a spurious entry in the cache.
+      cache.verifyAppendEntryIndex(entry);
       CompletableFuture<Long> writeFuture =
           fileLogWorker.writeLogEntry(entry, context).getFuture();
       if (stateMachineCachingEnabled) {
--- a/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/LogSegment.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/LogSegment.java
@@ -353,9 +353,9 @@ public final class LogSegment {
     return CorruptionPolicy.get(storage, RaftStorage::getLogCorruptionPolicy);
   }
 
-  void appendToOpenSegment(LogEntryProto entry, Op op) {
+  void appendToOpenSegment(LogEntryProto entry, Op op, boolean verifyEntryIndex) {
     Preconditions.assertTrue(isOpen(), "The log segment %s is not open for append", this);
-    append(true, entry, op, false);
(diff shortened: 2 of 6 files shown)


ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/SegmentedRaftLog.java
```diff
@@ -264,15 +264,13 @@ private void loadLogSegments(long lastIndexInSnapshot,
         }
       }
 
-      // if the largest index is smaller than the last index in snapshot, we do
-      // not load the log to avoid holes between log segments. This may happen
-      // when the local I/O worker is too slow to persist log (slower than
-      // committing the log and taking snapshot)
+      // If the end index is smaller than lastIndexInSnapshot, it means the state machine state is inconsistent
+      // with raft log state, fail the RaftServerImpl.start() to keep the state untacked.
```
szetszwo: > ... state machine state is inconsistent with raft log state ...

They don't have to be consistent:
1. leader's log index is 1000, but a (slow) follower only has log index 500
2. the follower dies
3. leader purges the log before 700
4. the follower restarts and reads log up to index 500
5. the follower installs a snapshot at index 1000
6. the follower dis again <--- snapshot 1000 but log 500

ChenSammi: @szetszwo , above is exactly the case as what I saw.  For this snapshot 1000 but log 500 case, current ratis will still start the RaftServer, which I think is not a good choice.   
For Ozone's case, it leads to OM shutdown due to "java.lang.IllegalStateException: gap between entry term"  , https://issues.apache.org/jira/browse/HDDS-15103.  Do you have any better suggestion for the fix solution?

szetszwo: @ChenSammi , For HDDS-15103, let's try to create a unit test to reproduce it.  Then, it would be easier to see how to fix it.

ChenSammi: Sorry, not [HDDS-15103](https://issues.apache.org/jira/browse/HDDS-15103), but [HDDS-15068](https://issues.apache.org/jira/browse/HDDS-15068).   

[HDDS-15103](https://issues.apache.org/jira/browse/HDDS-15103) is fixed.

szetszwo: For [HDDS-15068](https://issues.apache.org/jira/browse/HDDS-15068), Ratis should not write to the RaftLog when there is a gap.  Would like to fix it in this PR (or filing a new JIRA) ?  If not, I am happy to do it.

ChenSammi: Sure. We can reuse this jira.

szetszwo (approved):
+1 the change looks good.

ChenSammi:
Not sure the UT failed reason, under check.

PR #1440, opened 2026-04-25
RATIS-2508. appendEntries log messages improvement.

szetszwo:
See RATIS-2508

Due to checkstyle line number limit (2000), this PR also refactors RaftServerImpl.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/FollowerState.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/FollowerState.java
@@ -158,8 +158,8 @@ class FollowerState extends Daemon {
         }
         synchronized (server) {
           if (roleChangeChecking(electionTimeout)) {
-            LOG.info("{}: change to CANDIDATE, lastRpcElapsedTime:{}, electionTimeout:{}",
-                this, lastRpcTime.elapsedTime(), electionTimeout);
+            LOG.info("{}: change to CANDIDATE, lastRpcElapsedTime:{}ms, electionTimeout:{}",
+                this, lastRpcTime.elapsedTimeMs(), electionTimeout);
             server.getLeaderElectionMetrics().onLeaderElectionTimeout(); // Update timeout metric counters.
             // election timeout, should become a candidate
             server.changeToCandidate(false);
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -1699,7 +1699,7 @@ class RaftServerImpl implements RaftServer.Division,
       if  (t != null) {
         LOG.warn("{}: appendEntries* failed: {}", getMemberId(), toLogEntryTermIndexString(entries), t);
       } else if (LOG.isDebugEnabled()) {
-        LOG.debug("{}: appendEntries* succeeded {}", getMemberId(), toLogEntryTermIndexString(entries));
+        LOG.debug("{}: appendEntries* succeeded: {}", getMemberId(), toLogEntryTermIndexString(entries));
       }
       followerState.ifPresent(fs -> fs.updateLastRpcTime(FollowerState.UpdateType.APPEND_COMPLETE));
       timer.stop();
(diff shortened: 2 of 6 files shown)


ratis-server/src/main/java/org/apache/ratis/server/raftlog/LogProtoUtils.java
```diff
@@ -87,9 +87,23 @@ public static String toLogEntriesString(List<LogEntryProto> entries) {
...
+  public static String toLogEntryTermIndexString(List<LogEntryProto> entries) {
+    final int n = entries.size();
+    return n == 0 ? toLogEntryTermIndexString(n, null, null)
+        : toLogEntryTermIndexString(n, TermIndex.valueOf(entries.get(0)), TermIndex.valueOf(entries.get(n - 1)));
+  }
+
+  public static String toLogEntryTermIndexString(int n, TermIndex first, TermIndex last) {
+    return n == 0 ? "HEARTBEAT"
+        : n == 1 ? "entry" + first
+        : n + " entries:" + first + "..." + last;
```
adoroszlai: Missing `:`?

```suggestion
        : n == 1 ? "entry:" + first
        : n + " entries:" + first + "..." + last;
```

Also, would `=` instead of `:` be better?  Since `AppendEntriesRequest#toString` uses `:` as separator.

https://github.com/apache/ratis/blob/3e927ccbf2c44cc62e77ee3e8e4bee74328ee83a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java#L894-L896

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1703,6 +1696,11 @@ leaderId, getMemberId(), currentTerm, followerCommit, inconsistencyReplyNextInde
     final long commitIndex = effectiveCommitIndex(proto.getLeaderCommit(), previous, entries.size());
     final long matchIndex = isHeartbeat? RaftLog.INVALID_LOG_INDEX: entries.get(entries.size() - 1).getIndex();
     return appendFuture.whenCompleteAsync((r, t) -> {
+      if  (t != null) {
+        LOG.warn("{}: appendEntries* failed: {}", getMemberId(), toLogEntryTermIndexString(entries), t);
+      } else if (LOG.isDebugEnabled()) {
+        LOG.debug("{}: appendEntries* succeeded {}", getMemberId(), toLogEntryTermIndexString(entries));
```
adoroszlai: nit: `failed: {}` has `:` but `succeeded {}` does not.

adoroszlai (commented):

szetszwo:
@adoroszlai , thanks for reviewing this!  Just pushed a change to address your comments.

adoroszlai (approved):
Thanks @szetszwo for updating the patch, LGTM.

ratis-server/src/main/java/org/apache/ratis/server/impl/FollowerState.java
```diff
@@ -158,8 +158,8 @@ private void runImpl() {
         }
         synchronized (server) {
           if (roleChangeChecking(electionTimeout)) {
-            LOG.info("{}: change to CANDIDATE, lastRpcElapsedTime:{}, electionTimeout:{}",
-                this, lastRpcTime.elapsedTime(), electionTimeout);
+            LOG.info("{}: change to CANDIDATE, lastRpcElapsedTime:{}ms, electionTimeout:{}",
```
OneSizeFitsQuorum, line 161: why we add ms for lastRpcElapsedTime but not electionTimeout

OneSizeFitsQuorum (approved):
LGTM~ Only a minor problem

ratis-server/src/main/java/org/apache/ratis/server/impl/FollowerState.java
szetszwo: Before this change
- electionTimeout is a TimeDuration in ms
- elapsedTime() returns a TimeDuration in ns.

So, I changed elapsedTime() to elapsedTimeMs() and added ms.

slfan1989 (approved):

szetszwo:
@adoroszlai , @OneSizeFitsQuorum , @slfan1989 , thanks a lot for reviewing this!

PR #1443, opened 2026-04-30
RATIS-2397. Add trace support for Log Appender

taklwu:

Adding tracing support for AppendEntries / AppendEntriesAsync

https://issues.apache.org/jira/browse/RATIS-2397

## How was this patch tested?

1. added unit tests
2. build with #1428 and test locally with Jaegar UI to see the spans have been capture.

Screenshot 2026-04-29 at 5 04 41 PM

Screenshot 2026-04-29 at 4 54 36 PM

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/trace/TraceServer.java
+++ b/ratis-common/src/main/java/org/apache/ratis/trace/TraceServer.java
@@ -20,9 +20,14 @@ package org.apache.ratis.trace;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.context.Context;
+import org.apache.ratis.proto.RaftProtos.AppendEntriesRequestProto;
+import org.apache.ratis.proto.RaftProtos.RaftRpcRequestProto;
+import org.apache.ratis.proto.RaftProtos.SpanContextProto;
 import org.apache.ratis.protocol.RaftClientRequest;
+import org.apache.ratis.protocol.RaftPeerId;
 import org.apache.ratis.util.function.CheckedSupplier;
 
+import java.io.IOException;
 import java.util.concurrent.CompletableFuture;
 
 /** Server-side OpenTelemetry helpers. */
@@ -55,5 +60,35 @@ public final class TraceServer {
     span.setAttribute(RatisAttributes.CALL_ID, String.valueOf(request.getCallId()));
     span.setAttribute(RatisAttributes.MEMBER_ID, memberId);
     return span;
+  }
+
+  /**
+   * Traces follower handling of {@link AppendEntriesRequestProto} when the leader attached trace
+   * context (client-originated) for replication.
+   */
+  public static <T> CompletableFuture<T> traceAppendEntriesAsync(
+      CheckedSupplier<CompletableFuture<T>, IOException> action,
+      AppendEntriesRequestProto request, String memberId) throws IOException {
+    if (!TraceUtils.isEnabled()) {
+      return action.get();
+    }
+    final RaftRpcRequestProto rpc = request.getServerRequest();
+    final SpanContextProto spanContext = rpc.getSpanContext();
+    // If the leader sent no parent span context, still trace as a root span
+    // rather than skipping tracing entirely.
+    final Context remoteContext = (spanContext == null || spanContext.getContextMap().isEmpty())
+        ? Context.root()
+        : TraceUtils.extractContextFromProto(spanContext);
+    return TraceUtils.traceAsyncMethod(action, () -> {
+      final Span span = TraceUtils.getGlobalTracer()
+          .spanBuilder(SpanNames.APPEND_ENTRIES_ASYNC)
+          .setParent(remoteContext)
+          .setSpanKind(SpanKind.INTERNAL)
+          .startSpan();
+      span.setAttribute(RatisAttributes.MEMBER_ID, memberId);
+      span.setAttribute(RatisAttributes.PEER_ID, String.valueOf(RaftPeerId.valueOf(rpc.getRequestorId())));
+      span.setAttribute(RatisAttributes.APPEND_ENTRIES_COUNT, (long) request.getEntriesCount());
+      return span;
+    });
   }
 }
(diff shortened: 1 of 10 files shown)


taklwu:
@szetszwo so , I found the checkstyle is complaining about max lines in a single java class, does it mean we need to split this file into separate classes ? 

```
ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
 1: File length is 2,002 lines (max allowed is 2,000).
```

ratis-common/src/main/java/org/apache/ratis/trace/TraceServer.java
```diff
@@ -56,4 +61,34 @@ private static Span createServerSpanFromClientRequest(RaftClientRequest request,
...
+    final RaftRpcRequestProto rpc = request.getServerRequest();
+    final SpanContextProto spanContext = rpc.getSpanContext();
+    // If the leader sent no parent span context, still trace as a root span
+    // rather than skipping tracing entirely.
+    final Context remoteContext = (spanContext == null || spanContext.getContextMap().isEmpty())
+        ? Context.root()
+        : TraceUtils.extractContextFromProto(spanContext);
+    return TraceUtils.traceAsyncMethod(action, () -> {
+      final Span span = TraceUtils.getGlobalTracer()
+          .spanBuilder("raft.server.appendEntriesAsync")
```
szetszwo: Let's add a static constant for "raft.server.appendEntriesAsync".

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -733,6 +765,7 @@ void submitStepDownEvent(long term, StepDownReason reason) {
   }
 
   private void stepDown(long term, StepDownReason reason) {
+    replicationTraceByLogIndex.clear();
```
szetszwo: This is not needed since stop() will be callled.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -558,7 +568,28 @@ PendingRequest addPendingRequest(PendingRequests.Permit permit, RaftClientReques
...
+      ConcurrentHashMap<Long, SpanContextProto> traceByIndex) {
+    if (entries == null || entries.isEmpty()) {
+      return null;
+    }
+    for (LogEntryProto e : entries) {
+      final SpanContextProto sc = traceByIndex.get(e.getIndex());
+      if (sc != null && !sc.getContextMap().isEmpty()) {
+        return sc;
+      }
+    }
```
szetszwo: Since entries is a list, there could be multiple PendingRequest(s).  Should we return multiple SpanContextProto(s)?

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -558,7 +568,28 @@ PendingRequest addPendingRequest(PendingRequests.Permit permit, RaftClientReques
...
+    return pending;
+  }
+
+  private static SpanContextProto tracingContextForReplication(List<LogEntryProto> entries,
+      ConcurrentHashMap<Long, SpanContextProto> traceByIndex) {
+    if (entries == null || entries.isEmpty()) {
+      return null;
+    }
+    for (LogEntryProto e : entries) {
+      final SpanContextProto sc = traceByIndex.get(e.getIndex());
```
szetszwo: Since it needs to loop for entries in consecutive indices, it is more efficient to use a NavigableMap.  I suggest to use TreeMap with a read-write lock.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -354,6 +356,13 @@ boolean isApplied() {
   private final long followerMaxGapThreshold;
   private final PendingStepDown pendingStepDown;
 
+  /**
+   * Client-originated trace context keyed by log index; attached to {@link AppendEntriesRequestProto}
+   * so follower {@code appendEntries} spans join the same trace as the client write.
+   */
+  private final ConcurrentHashMap<Long, SpanContextProto> replicationTraceByLogIndex =
+      new ConcurrentHashMap<>();
```
szetszwo: Let's create a new class, say LeaderTracer, and move all the tracing code there.

szetszwo (commented):
@taklwu , thanks for working on this!

Please see the comments inlined and also https://issues.apache.org/jira/secure/attachment/13082105/1443_review.patch

ratis-common/src/main/java/org/apache/ratis/trace/TraceServer.java
taklwu: I made a new class SpanNames and add this into it, also it comes with other refactoring

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
taklwu: AFAIK that AppendEntriesRequestProto is a single RPC with a batch of append entries, so, we should only have a single parent spancontext for this batch.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
taklwu: ack, I followed your change.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
taklwu: so does that mean we only add trace for the actual leader ?

taklwu (commented):
thanks for your time! and please review again.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
taklwu: nvm, I have removed it.

taklwu:
`TestRaftAsyncWithNetty` failure should not be related, I may need to retrigger the execution

szetszwo (approved):
+1 the change looks good.

szetszwo:
@taklwu , could you merge rebase with master?  The are some recent commits fixing the test failures.

taklwu:
@szetszwo `TestCacheEviction` and `TestRaftAsyncWithNetty` are still failing after rebase

szetszwo:
@taklwu , could you sync with the master one more time for "RATIS-2498. Mark TestRaftAsyncWithNetty as flaky"?

adoroszlai:
Thanks @taklwu for the patch, @szetszwo for the review.

PR #1444, opened 2026-05-02
RATIS-2511. Follower should throw ReadException if it is installing snapshot

ivandika3:

When Ratis follower is installing snapshot, its appliedIndex will not advance until the snapshot is installed. This can cause all linearizable follower read on this node to timeout due to the client timeout (default is 3s).

Instead of waiting until the client timeout, Ratis follower can check whether it is currently installing snapshot and trigger failover immediately if it does. Any new or pending read requests in the follower need to return ReadException.

This decision is based on the assumption that snapshot will take a longer time and might stall the follower read, which I think is a reasonable assumption. However, if snapshot is very small and InstallSnapshot can be done quickly, throwing ReadException might trigger unnecessary failover.

https://issues.apache.org/jira/browse/RATIS-2511

## How was this patch tested?

Integration test.

Clean CI: https://github.com/ivandika3/ratis/actions/runs/25156971002

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -1092,8 +1092,11 @@ class RaftServerImpl implements RaftServer.Division,
     }
     return processQueryFuture(stateMachine.queryStale(request.getMessage(), minIndex), request);
   }
-
   private CompletableFuture<ReadIndexReplyProto> sendReadIndexAsync(RaftClientRequest clientRequest) {
+    final Throwable snapshotInstallation = snapshotInstallationHandler.getInProgressInstallSnapshotReadException();
+    if (snapshotInstallation != null) {
+      return JavaUtils.completeExceptionally(snapshotInstallation);
+    }
     final RaftPeerId leaderId = getInfo().getLeaderId();
     if (leaderId == null) {
       return JavaUtils.completeExceptionally(new ReadIndexException(getMemberId() + ": Leader is unknown."));
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -1093,10 +1093,6 @@ class RaftServerImpl implements RaftServer.Division,
     return processQueryFuture(stateMachine.queryStale(request.getMessage(), minIndex), request);
   }
 
-  ReadRequests getReadRequests() {
-    return getState().getReadRequests();
-  }
-
   private CompletableFuture<ReadIndexReplyProto> sendReadIndexAsync(RaftClientRequest clientRequest) {
     final RaftPeerId leaderId = getInfo().getLeaderId();
     if (leaderId == null) {
@@ -1146,7 +1142,8 @@ class RaftServerImpl implements RaftServer.Division,
       }
 
       return replyFuture
-          .thenCompose(readIndex -> getReadRequests().waitToAdvance(readIndex))
+          .thenCompose(readIndex -> getState().getReadRequests().waitToAdvance(readIndex,
+              snapshotInstallationHandler::getInProgressInstallSnapshotReadException))
           .thenCompose(readIndex -> queryStateMachine(request))
           .exceptionally(e -> readException2Reply(request, e));
     } else {
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -1092,10 +1092,14 @@ class RaftServerImpl implements RaftServer.Division,
     }
     return processQueryFuture(stateMachine.queryStale(request.getMessage(), minIndex), request);
   }
+  ReadException newReadException(String op, long installSnapshot, boolean started) {
+    return new ReadException(getMemberId() + ": Failed to " + op + " readIndex as snapshot (" + installSnapshot
+        + ") installation is " + (started ? "started" : "in progress"));
+  }
   private CompletableFuture<ReadIndexReplyProto> sendReadIndexAsync(RaftClientRequest clientRequest) {
-    final Throwable snapshotInstallation = snapshotInstallationHandler.getInProgressInstallSnapshotReadException();
-    if (snapshotInstallation != null) {
-      return JavaUtils.completeExceptionally(snapshotInstallation);
+    final long installSnapshot = snapshotInstallationHandler.getInProgressInstallSnapshotIndex();
+    if (installSnapshot != RaftLog.INVALID_LOG_INDEX) {
+      return JavaUtils.completeExceptionally(newReadException("get", installSnapshot, false));
     }
     final RaftPeerId leaderId = getInfo().getLeaderId();
     if (leaderId == null) {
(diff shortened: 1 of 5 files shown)


ratis-server/src/main/java/org/apache/ratis/server/impl/ReadRequests.java
```diff
@@ -88,6 +101,14 @@ private void handleTimeout(long readIndex) {
       removed.completeExceptionally(new ReadException("Read timeout " + readTimeout + " for index " + readIndex));
     }
 
+    void fail(Throwable cause) {
+      final Collection<CompletableFuture<Long>> futures;
+      synchronized (this) {
+        futures = new ArrayList<>(sorted.values());
+        sorted.clear();
```
szetszwo: - Let's make sorted non-final in order to avoid the copying.
- We may return the futures here and fail them in ReadRequests.fail(..)..
```java
//ReadIndexQueue
    synchronized Collection<CompletableFuture<Long>> clear(Throwable cause) {
      final Collection<CompletableFuture<Long>> futures = sorted.values();
      sorted = new TreeMap<>();
      return futures;
    }
```

```java
//ReadRequests
  void fail(Throwable cause) {
    for(CompletableFuture<Long> f : readIndexQueue.clear(cause)) {
      f.completeExceptionally(cause);
    }
  }
```

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1092,12 +1092,11 @@ private CompletableFuture<RaftClientReply> staleReadAsync(RaftClientRequest requ
     }
     return processQueryFuture(stateMachine.queryStale(request.getMessage(), minIndex), request);
   }
-
-  ReadRequests getReadRequests() {
-    return getState().getReadRequests();
-  }
-
   private CompletableFuture<ReadIndexReplyProto> sendReadIndexAsync(RaftClientRequest clientRequest) {
+    final Throwable snapshotInstallation = snapshotInstallationHandler.getInProgressInstallSnapshotReadException();
```
szetszwo: Let's create the ReadException with a different message.  It will be easier to debug later.
```java
  ReadException newReadException(String op, long installSnapshot, boolean started) {
    return new ReadException(getMemberId() + ": Failed to " + op + " readIndex as snapshot (" + installSnapshot
        + ") installation is " + (started ? "started" : "in progress"));
  }

  private CompletableFuture<ReadIndexReplyProto> sendReadIndexAsync(RaftClientRequest clientRequest) {
    final long installSnapshot = snapshotInstallationHandler.getInProgressInstallSnapshotIndex();
    if (installSnapshot != RaftLog.INVALID_LOG_INDEX) {
      return JavaUtils.completeExceptionally(newReadException("get", installSnapshot, false));
    }
```

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1146,7 +1145,8 @@ private CompletableFuture<RaftClientReply> readAsync(RaftClientRequest request)
       }
 
       return replyFuture
-          .thenCompose(readIndex -> getReadRequests().waitToAdvance(readIndex))
+          .thenCompose(readIndex -> getState().getReadRequests().waitToAdvance(readIndex,
+              snapshotInstallationHandler::getInProgressInstallSnapshotReadException))
```
szetszwo: Let's add a waitReadIndex method and check there.
```java
          .thenCompose(this::waitReadIndex)
```
```java
  private CompletableFuture<Long> waitReadIndex(long readIndex) {
    final long installSnapshot = snapshotInstallationHandler.getInProgressInstallSnapshotIndex();
    if (installSnapshot != RaftLog.INVALID_LOG_INDEX) {
      return JavaUtils.completeExceptionally(newReadException("start waiting for", installSnapshot, false));
    }
    return getState().getReadRequests().waitToAdvance(readIndex);
  }
```

ratis-server/src/main/java/org/apache/ratis/server/impl/ReadRequests.java
```diff
@@ -20,21 +20,30 @@
...
+import java.util.function.Supplier;
 
 /** For supporting linearizable read. */
 class ReadRequests {
   private static final Logger LOG = LoggerFactory.getLogger(ReadRequests.class);
 
+  static ReadException newException(Object server, long installSnapshot) {
+    return new ReadException(server + ": Failed read as snapshot (" + installSnapshot
+        + ") installation is in progress");
+  }
```
szetszwo: Let's move this method to `RaftServerImpl.newReadException` as mentioned earlier.

ratis-server/src/main/java/org/apache/ratis/server/impl/ReadRequests.java
```diff
@@ -52,10 +61,14 @@ static class ReadIndexQueue {
...
 
-    CompletableFuture<Long> add(long readIndex) {
+    CompletableFuture<Long> add(long readIndex, Supplier<Throwable> failureSupplier) {
       final CompletableFuture<Long> returned;
       final boolean create;
       synchronized (this) {
+        final Throwable failure = failureSupplier.get();
+        if (failure != null) {
+          return JavaUtils.completeExceptionally(failure);
+        }
```
szetszwo: This can be checked in RaftServerImpl.

szetszwo (commented):
@ivandika3 , thanks for working not this!

See the comments inlined and also https://issues.apache.org/jira/secure/attachment/13082227/1444_review.patch

BTW, we current have a RaftServerImpl 2000 line checkstyle limit; filed RATIS-2532.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
ivandika3: Thanks for the review. Updated, but there might be a small chance of interleaving here since the snapshot checking is not protected by ReadRequests object lock.

1. waitReadIndex() checks snapshot state and sees no snapshot.
1. Snapshot notification sets inProgressInstallSnapshotIndex.
1. Snapshot notification calls ReadRequests.fail() and clears the current queue.
1. Read then enqueues itself in ReadRequests.waitToAdvance() after the failure sweep.

We can probably synchronized on RaftServerImpl lock, but I'm afraid this might increase follower read contention.

szetszwo: @ivandika3 , That is a good point!  Then, let's use your previous approach then; see https://issues.apache.org/jira/secure/attachment/13082234/1444_review2.patch

szetszwo:
@ivandika3 , not sure if you have seen [this comment](https://github.com/apache/ratis/pull/1444#discussion_r3244182726)?

ivandika3:
@szetszwo Thanks for checking, forgot to push it. I have addressed the comment in the latest commit.

szetszwo (approved):
+1 the change looks good.

szetszwo:
> This branch has not been deployed
> No deployments

Not sure why it did not run the checks and the tests?  Cc @adoroszlai

ivandika3:
> Not sure why it did not run the checks and the tests?

Thanks @szetszwo for the review. You can refer to https://github.com/apache/ratis/pull/1448#issuecomment-4393394029 , my GH action is still blocked so any commit written by me won't trigger Github actions (raised tickets but no update from GH side so it might be a while).

One way is you can help to trigger an empty commit to trigger CI.

ivandika3:
Thanks @szetszwo for the review and triggering the CI.

PR #1448, opened 2026-05-05
RATIS-2509. Introduce local read API to reduce serde cost

ivandika3:

See https://issues.apache.org/jira/browse/RATIS-2509

OM code integration (`OzoneManagerRatisRequest#submitRequest`)

```java
    public OMResponse submitReadRequest(OMRequest omRequest, Function<OMRequest, OMResponse> handler)
      throws ServiceException {
    try {
      return serverDivision.readOnlyAsync(() -> CompletableFuture.completedFuture(handler.apply(omRequest))).get();
    } catch (ExecutionException | IOException ex) {
      throw new ServiceException(ex.getMessage(), ex);
    } catch (InterruptedException ex) {
      throw new ServiceException(ex.getMessage(), ex);
    }

  }

```

https://issues.apache.org/jira/browse/RATIS-2509

## How was this patch tested?

OM Leader read performance testing, the Ratis read performance is on par with the direct read.
- Before patch: 235020 QPS -> 180433 QPS (24% QPS reduction)
- After patch: 235020 QPS -> 228362 (3% QPS reduction)

Follower read throughput remains unchanged, so might not be the main overhead.

### DIFF
--- a/ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
+++ b/ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
@@ -24,12 +24,16 @@ import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
 import org.apache.ratis.conf.Parameters;
 import org.apache.ratis.conf.RaftProperties;
 import org.apache.ratis.proto.RaftProtos.CommitInfoProto;
 import org.apache.ratis.proto.RaftProtos.RaftPeerRole;
+import org.apache.ratis.proto.RaftProtos.ReadRequestTypeProto;
 import org.apache.ratis.protocol.AdminAsynchronousProtocol;
 import org.apache.ratis.protocol.AdminProtocol;
+import org.apache.ratis.protocol.ClientId;
 import org.apache.ratis.protocol.RaftClientAsynchronousProtocol;
 import org.apache.ratis.protocol.RaftClientProtocol;
 import org.apache.ratis.protocol.RaftGroup;
@@ -101,6 +105,21 @@ public interface RaftServer extends Closeable, RpcType.Get,
 
     /** @return the {@link StateMachine} for this division. */
     StateMachine getStateMachine();
+
+    /**
+     * Execute a local read-only query after applying the configured
+     * {@link RaftServerConfigKeys.Read.Option} consistency checks.
+     *
+     * <p>This API is intended for embedded users that already have a local
+     * server division and want to avoid serializing application read requests
+     * and responses through {@link org.apache.ratis.protocol.Message}. Remote
+     * clients should continue to use
+     * {@link org.apache.ratis.client.api.AsyncApi#sendReadOnly(org.apache.ratis.protocol.Message)}.
+     */
+    default <T> CompletableFuture<T> readOnlyAsync(ClientId clientId, ReadRequestTypeProto readRequestType,
+        Supplier<CompletableFuture<T>> query) throws IOException {
+      throw new UnsupportedOperationException("readOnlyAsync is not supported");
+    }
 
     /** @return the raft log of this division. */
     RaftLog getRaftLog();
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -37,12 +37,14 @@ import org.apache.ratis.proto.RaftProtos.RaftPeerRole;
 import org.apache.ratis.proto.RaftProtos.RaftRpcRequestProto;
 import org.apache.ratis.proto.RaftProtos.ReadIndexReplyProto;
 import org.apache.ratis.proto.RaftProtos.ReadIndexRequestProto;
+import org.apache.ratis.proto.RaftProtos.ReadRequestTypeProto;
 import org.apache.ratis.proto.RaftProtos.ReplicationLevel;
 import org.apache.ratis.proto.RaftProtos.RequestVoteReplyProto;
 import org.apache.ratis.proto.RaftProtos.RequestVoteRequestProto;
 import org.apache.ratis.proto.RaftProtos.RoleInfoProto;
 import org.apache.ratis.proto.RaftProtos.StartLeaderElectionReplyProto;
 import org.apache.ratis.proto.RaftProtos.StartLeaderElectionRequestProto;
+import org.apache.ratis.protocol.ClientId;
 import org.apache.ratis.protocol.ClientInvocationId;
 import org.apache.ratis.protocol.GroupInfoReply;
 import org.apache.ratis.protocol.GroupInfoRequest;
@@ -1099,7 +1101,12 @@ class RaftServerImpl implements RaftServer.Division,
     if (leaderId == null) {
       return JavaUtils.completeExceptionally(new ReadIndexException(getMemberId() + ": Leader is unknown."));
     }
-    final ReadIndexRequestProto request = toReadIndexRequestProto(clientRequest, getMemberId(), leaderId);
+    return sendReadIndexAsync(clientRequest.getClientId(), clientRequest.getType().getRead(), leaderId);
+  }
+
+  private CompletableFuture<ReadIndexReplyProto> sendReadIndexAsync(
+      ClientId clientId, ReadRequestTypeProto readRequestType, RaftPeerId leaderId) {
+    final ReadIndexRequestProto request = toReadIndexRequestProto(clientId, readRequestType, getMemberId(), leaderId);
     try {
       return getServerRpc().async().readIndexAsync(request);
     } catch (IOException e) {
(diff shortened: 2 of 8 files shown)


ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -102,6 +104,21 @@ default RaftGroup getGroup() {
...
+     * {@link RaftServerConfigKeys.Read.Option} consistency checks.
+     *
+     * <p>This API is intended for embedded users that already have a local
+     * server division and want to avoid serializing application read requests
+     * and responses through {@link org.apache.ratis.protocol.Message}. Remote
+     * clients should continue to use
+     * {@link org.apache.ratis.client.api.AsyncApi#sendReadOnly(org.apache.ratis.protocol.Message)}.
+     */
+    default <T> CompletableFuture<T> readOnlyAsync(
+        CheckedSupplier<CompletableFuture<T>, IOException> query) throws IOException {
```
szetszwo: - Since the query method in StateMachine does not throw IOException, let's just use Supplier here.  It will force the StateMachine to handle IOException, if there is any.
```java
//StateMachine
  CompletableFuture<Message> query(Message request);
```
- Also, we need clientId and readRequestType to support read-after-write-consistency.
```java
    default <T> CompletableFuture<T> readOnlyAsync(ClientId clientId, ReadRequestTypeProto readRequestType,
        Supplier<CompletableFuture<T>> query) throws IOException {
```

szetszwo (commented):
@ivandika3 , thanks for working on this!  Please see the comments inlined and also https://issues.apache.org/jira/secure/attachment/13082106/1448_review.patch

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
ivandika3: Good point, updated.

szetszwo (approved):
+1 the change looks good.

ivandika3:
@szetszwo Thanks for the review. Currently I'm trying to wait until my GH actions are enabled before merging it since I haven't seen any green CI for this patch. If we can find another way, that's good. Otherwise, we can wait until my GH actions situation is resolved.

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1107,6 +1121,68 @@ private CompletableFuture<Long> getReadIndex(RaftClientRequest request, LeaderSt
...
+    Objects.requireNonNull(readRequestType, "readRequestType == null");
+    Objects.requireNonNull(query, "query == null");
+    assertLifeCycleState(LifeCycle.States.RUNNING);
+    if (readRequestType.getPreferNonLinearizable() || readOption == RaftServerConfigKeys.Read.Option.DEFAULT) {
+      final CompletableFuture<T> reply = checkLeaderStateForReadOnly();
+      return reply != null ? reply : supplyReadOnly(query);
+    } else if (readOption == RaftServerConfigKeys.Read.Option.LINEARIZABLE) {
+      return getReadIndexForReadOnly(clientId, readRequestType)
+          .thenCompose(readIndex -> getReadRequests().waitToAdvance(readIndex))
+          .thenCompose(readIndex -> supplyReadOnly(query));
```
ivandika3, line 1181: I wonder whether it is better to use thenComposeAsync to improve performance.

```java
return getReadIndexForReadOnly(clientId, readRequestType)
          .thenComposeAsync(readIndex -> getReadRequests().waitToAdvance(readIndex), clientExecutor)
          .thenComposeAsync(readIndex -> supplyReadOnly(query), clientExecutor);
```

or just the supplyReadOnly

```java
return getReadIndexForReadOnly(clientId, readRequestType)
          .thenCompose(readIndex -> getReadRequests().waitToAdvance(readIndex))
          .thenComposeAsync(readIndex -> supplyReadOnly(query), clientExecutor);
```

Seems currently with `thenCompose`, the actual read is going to be executed sequentially in gRPC `SerializingExecutor`.

Let me benchmark it. Can also be applied to `readAsync`.

szetszwo: We should use clientExecutor as earliest as possible.  It will keep using the executor for the later calls; see below:
```
//output
                                    main: main
                             supplyAsync: pool-A-thread1
        thenCompose after pool-A-thread1: pool-A-thread1
   thenComposeAsync after pool-A-thread1: pool-B-thread1
  thenCompose again after pool-B-thread1: pool-B-thread1
```
```java
  public static void main(String[] args) {
    final ExecutorService poolA = Executors.newCachedThreadPool(ConcurrentUtils.newThreadFactory("pool-A"));
    final ExecutorService poolB = Executors.newCachedThreadPool(ConcurrentUtils.newThreadFactory("pool-B"));

    printCurrentThread("main");
    CompletableFuture.supplyAsync(() -> printCurrentThread("supplyAsync"), poolA)
        .thenCompose(s -> CompletableFuture.completedFuture(printCurrentThread("thenCompose after " + s)))
        .thenComposeAsync(s -> CompletableFuture.completedFuture(printCurrentThread("thenComposeAsync after " + s)), poolB)
        .thenCompose(s -> CompletableFuture.completedFuture(printCurrentThread("thenCompose again after " + s)))
        ;
  }

  static String printCurrentThread(String label) {
    final String name = Thread.currentThread().getName();
    System.out.printf("%40s: %s%n", label, name);
    return name;
  }
```

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
```diff
@@ -1109,6 +1116,74 @@ private CompletableFuture<ReadIndexReplyProto> sendReadIndexAsync(RaftClientRequ
...
+        throw new CompletionException(new ReadIndexException(getId()
+            + ": Failed to get read index from the leader: " + reply));
+      }
+    });
+  }
+
+  private CompletableFuture<Long> getReadIndexForReadOnly(ClientId clientId, ReadRequestTypeProto readRequestType) {
+    final LeaderStateImpl leader = role.getLeaderState().orElse(null);
+    if (leader != null) {
+      return leader.getReadIndex(null);
```
spacemonkd, line 1133: Isn't this going to bypass the `writeIndexCache`?
Say if a caller constructs a `ReadRequestTypeProto` with `readAfterWriteConsistent` as true and passes it to `readOnlyAsync`, isn't the flag getting ignored when the node is the leader?

Maybe we can add a Javadoc saying `readAfterWriteConsistent` is not supported here, or we can throw an exception like `IllegalArgumentException` in case `readAfterWriteConsistent` flag is set so that we get some visibility into this behaviour?

ratis-server/src/test/java/org/apache/ratis/ReadOnlyRequestTests.java
```diff
@@ -218,6 +236,20 @@ static int retrieve(RaftClientReply reply) {
     return Integer.parseInt(reply.getMessage().getContent().toString(StandardCharsets.UTF_8));
   }
 
+  static long getCount(RaftServer.Division server) {
+    return ((CounterStateMachine) server.getStateMachine()).getCount();
+  }
+
+  static <T> CompletableFuture<T> readOnlyAsync(
+      RaftServer.Division server, Supplier<CompletableFuture<T>> query) throws IOException {
+    return server.readOnlyAsync(ClientId.randomId(), RaftClientRequest.readRequestType().getRead(), query);
```
spacemonkd, line 245: Just one suggestion. 
As far as I understand from the code `ClientId.randomId()` is not used at all for the default path. It goes  straight to `checkLeaderStateForReadOnly` and then `supplyReadOnly`.

Maybe we can overload the Division interface with something like:
```
 default <T> CompletableFuture<T> readOnlyAsync(Supplier<CompletableFuture<T>> query) throws IOException {
     return readOnlyAsync(ClientId.randomId(), ReadRequestTypeProto.getDefaultInstance(), query);
   }
```
so that callers can avoid importing `ClientId` and `RaftClientRequest`?
Right now it is only one call site so it is fine I guess, but if we have more callers this might be a better way.

spacemonkd (commented):
Thanks @ivandika3 for picking this up.
I had a few questions / suggestions and your inputs would be great, also it seems the checkstyle is failing. We should be good to merge once that is addressed.

Also on the test coverage I asked Claude to analyse the scenarios and it gave the following cases not covered yet.

Untested paths for readOnlyAsync:
- Follower in LINEARIZABLE mode when leader is unknown (the ReadIndexException at line 1143)
- Follower in LINEARIZABLE mode during snapshot installation (the ReadException at line 1138)
- Supplier that returns null (the Objects.requireNonNull at line 1161)
- Supplier that throws (the catch (Throwable t) at line 1163)
- Server not in RUNNING state (the assertLifeCycleState at line 1173)

Maybe it might be good to add these cases as well?

ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
ivandika3: Thanks, I have updated it so call `writeIndexCache`.

ratis-server/src/test/java/org/apache/ratis/ReadOnlyRequestTests.java
ivandika3: Thanks, updated.

ivandika3:
@spacemonkd Thanks for the review, I have addressed the reviews and added the tests.

@szetszwo Could you help take a look again?

ratis-server/src/main/java/org/apache/ratis/server/impl/WriteIndexCache.java
```diff
@@ -58,8 +59,13 @@ void add(ClientId key, CompletableFuture<Long> current) {
   }
 
   CompletableFuture<Long> getWriteIndexFuture(RaftClientRequest request) {
-    if (request != null && request.getType().getRead().getReadAfterWriteConsistent()) {
-      final AtomicReference<CompletableFuture<Long>> ref = cache.getIfPresent(request.getClientId());
+    return request != null ? getWriteIndexFuture(request.getClientId(), request.getType().getRead())
+        : CompletableFuture.completedFuture(null);
+  }
```
szetszwo: This becomes unused.  Please remove it.

ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -142,6 +146,33 @@ default RaftGroup getGroup() {
...
+    /**
+     * Execute a local read-only query after applying the configured
+     * {@link RaftServerConfigKeys.Read.Option} consistency checks.
+     *
+     * <p>The {@code clientId} and {@code readRequestType} identify the read
+     * semantics.  In particular, callers requesting read-after-write
+     * consistency must use the same client ID as their prior writes.
+     */
+    default <T> CompletableFuture<T> readOnlyAsync(ClientId clientId, ReadRequestTypeProto readRequestType,
+        Supplier<CompletableFuture<T>> query) throws IOException {
```
szetszwo: On a second thought, should we use Message instead of Supplier?
```java
    default CompletableFuture<Message> readOnlyAsync(ClientId clientId, ReadRequestTypeProto readRequestType,
        Message queryMessage) throws IOException {
```

szetszwo (commented):
@ivandika3 , thanks for the update!

- I found that we can restructure the code in order to reuse them: https://issues.apache.org/jira/secure/attachment/13084753/1448_review2.patch
- Please see also the comments inlined.

PR #1452, opened 2026-05-06
RATIS-2521. LifeCycle.startAndTransition(..) may cause illegal transition: CLOSING -> EXCEPTION.

szetszwo:
See RATIS-2521

### DIFF
--- a/ratis-test/src/test/java/org/apache/ratis/util/TestLifeCycle.java
+++ b/ratis-test/src/test/java/org/apache/ratis/util/TestLifeCycle.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


ratis-test/src/test/java/org/apache/ratis/util/TestLifeCycle.java
```diff
@@ -101,4 +104,68 @@ private static void testInvalidTransition(TriConsumer<LifeCycle.State, LifeCycle
...
+      try {
+        simulatedServer.start();
+        throw new AssertionError("start() should fail");
+      } catch (Exception e) {
+        return e.getCause();
+      }
+    });
+
+    Thread.sleep(100);
+    assertEquals(STARTING, simulatedServer.getLifeCycleState());
```
adoroszlai, line 122: Would using `JavaUtils.attemptUntilTrue` be better than `Thread.sleep`?

adoroszlai (approved):

ratis-test/src/test/java/org/apache/ratis/util/TestLifeCycle.java
szetszwo: Since the test is very simple, it is reliable to just use `Thread.sleep`.

An embarrassing reason is that TestRaftAsyncWithNetty often fails.  It is not easy to pass all the tests.  We probably should disable TestRaftAsyncWithNetty until it is fixed.

PR #1455, opened 2026-05-08
RATIS-2524. Implement ReadIndex batching

ivandika3:

See https://issues.apache.org/jira/browse/RATIS-2524

https://issues.apache.org/jira/browse/RATIS-2524

## How was this patch tested?

UT

### DIFF
--- a/ratis-docs/src/site/markdown/configurations.md
+++ b/ratis-docs/src/site/markdown/configurations.md
@@ -253,6 +253,31 @@ but there are tradeoffs (e.g. Write and Read performance) between different type
 | **Type**        | TimeDuration                                                                                                                                |
 | **Default**     | 10ms                                                                                                                                        |
 
+| **Property**    | `raft.server.read.read-index.batch.enabled`                                       |
+|:----------------|:----------------------------------------------------------------------------------|
+| **Description** | whether to batch follower-to-leader ReadIndex RPCs for plain linearizable reads    |
+| **Type**        | boolean                                                                           |
+| **Default**     | false                                                                             |
+
+| **Property**    | `raft.server.read.read-index.batch.interval`        |
+|:----------------|:----------------------------------------------------|
+| **Description** | maximum time to collect reads into one ReadIndex batch |
+| **Type**        | TimeDuration                                       |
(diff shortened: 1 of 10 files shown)


ivandika3:
Need to wire the ReadIndex batching to local read path as well (https://github.com/apache/ratis/pull/1448).

ivandika3:
Instead of having a certain fixed interval and size for batching which will increase the latency of read requests (even if the cluster load is small), we can try to follow the "Opportunistic ALR batching" from the paper so that we send ReadIndex when read are available. Therefore the batching will be adaptive depends on the read workloads. In small workloads, this means that the each read would trigger ReadIndex (batch size 1). However, for larger workloads, the batch can start to be bigger since there are more pending read requests. This "Opportunistic" batching is also called "Natural batching" (https://martinfowler.com/articles/mechanical-sympathy-principles.html).

This would probably involve a queue (maybe a circular buffer) that will hold the read requests. The serverExecutor(s) tried to consume the requests as much as possible and send a ReadIndex for that "batch". However only one serverExecutor can consume the queue at one time since one read requests should only be owned by one batch (i.e. one ReadIndex).

szetszwo:
@ivandika3 ,  Thanks for working on this!  Please let me know when this is ready for review.

ivandika3:
Thanks @szetszwo for checking this out. Did some benchmark for the first iteration where we batch by time and size and there are some improvements (16K - 18K), but need to reduce the batch time to 50 microseconds. However, even then, the pure follower reads performance still cannot beat the pure leader read performance (22K). 

I will try to benchmark the natural batching method. However, currently I'm still preoccupied with other tasks so it won't be done so soon. 

If even with natural batching it doesn't work, we might need to go one level deeper to reduce ReadIndex latency using things like IO_URING ([RATIS-2545](https://issues.apache.org/jira/browse/RATIS-2545)) or even changing the architecture to use SEDA similar to Sofa-Jraft (i.e. start to think about mechanical sympathy). I am also researching the implementation of low-latency systems like Disruptor, Aeron, Chronicle to reduce the latency even further. However, please let me know if there are some bottlenecks (low hanging fruits) that I missed.

PR #1459, opened 2026-05-14
RATIS-2534. Skip zizmor on dependabot push

adoroszlai:

Similar to RATIS-2070, for the `zizmor` check this time.

Background: dependabot pushes dependency version bumps to the main repo instead of its own fork. It then creates a pull request without waiting for the push workflow to succeed. Thus both workflows test the same state of code, which is unnecessary.

https://issues.apache.org/jira/browse/RATIS-2534

## How was this patch tested?

In use for `post-commit.yaml` since RATIS-2070.

https://github.com/adoroszlai/ratis/actions/runs/25862410063

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -17,6 +17,10 @@ name: zizmor
 
 on:
   push:
+    branches-ignore:
+      - 'dependabot/**'
+    tags:
+      - '**'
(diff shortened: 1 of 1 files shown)


szetszwo (commented):
+1 the change looks good.

adoroszlai:

PR #1462, opened 2026-05-19
RATIS-2537. Support configurable gRPC TLS provider and cipher suites

HTHou:
RATIS-2537

## What changed

This patch makes gRPC TLS settings configurable through `RaftProperties`:

- Netty SSL provider
- JSSE provider name
- enabled TLS protocols
- enabled cipher suites

It also allows generic JSSE providers, such as BCJSSE, to be used with Netty's JDK `SslContext` path and ALPN `h2`.

## Why

Ratis currently hardcodes most gRPC TLS behavior around Netty's defaults. Deployments that need custom JSSE providers or non-default protocol/cipher suites cannot configure them without code changes.

## Validation

```bash
./mvnw -pl ratis-grpc -Dtest=TestGrpcTlsConfig test
./mvnw -pl ratis-grpc checkstyle:check
./mvnw -pl ratis-grpc install -DskipTests
```

I also ran a local BCJSSE smoke test with `TLSv1.3`, `TLS_AES_128_GCM_SHA256`, and ALPN `h2`.

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcConfigKeys.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcConfigKeys.java
@@ -21,11 +21,15 @@ import org.apache.ratis.conf.Parameters;
 import org.apache.ratis.conf.RaftProperties;
 import org.apache.ratis.grpc.server.GrpcServices;
 import org.apache.ratis.server.RaftServerConfigKeys;
+import org.apache.ratis.thirdparty.io.netty.handler.ssl.SslProvider;
 import org.apache.ratis.util.SizeInBytes;
+import org.apache.ratis.util.StringUtils;
 import org.apache.ratis.util.TimeDuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Arrays;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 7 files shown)


ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcConfigKeys.java
```diff
@@ -87,6 +91,73 @@ static GrpcTlsConfig conf(Parameters parameters) {
...
+    String[] PROTOCOLS_DEFAULT = StringUtils.EMPTY_STRING_ARRAY;
+    static String[] protocols(RaftProperties properties) {
+      return StringUtils.getTrimmedStrings(properties.get(PROTOCOLS_KEY, null));
+    }
+    static void setProtocols(RaftProperties properties, String... protocols) {
+      properties.set(PROTOCOLS_KEY, String.join(",", protocols));
+    }
+
+    String CIPHER_SUITES_KEY = PREFIX + ".cipher.suites";
+    String[] CIPHER_SUITES_DEFAULT = StringUtils.EMPTY_STRING_ARRAY;
```
szetszwo: Since the GrpcTlsConfig objects are from the Parameters provided by the application, so we don't need the new confs in Ratis -- The application should set up everything in the GrpcTlsConfig objects before putting them in the Parameters.

ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcTlsConfig.java
```diff
@@ -34,6 +36,10 @@
  */
 public class GrpcTlsConfig extends TlsConf {
   private final boolean fileBasedConfig;
+  private final SslProvider sslProvider;
+  private final String jsseProviderName;
+  private final String[] protocols;
+  private final String[] cipherSuites;
```
szetszwo: Let's use List and make them unmodifiable. (The code in this PR actually copies the arrays in the get methods. Thus, using arrays do not make it more efficient.)

szetszwo (commented):
@HTHou , thanks for working on this!  It took me some time to understand the change.  Please see the comments inlined.

ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcConfigKeys.java
HTHou: Thanks for the review. Addressed in 900b8776 by removing the new RaftProperties TLS options and using the GrpcTlsConfig objects from Parameters directly.

ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcTlsConfig.java
HTHou: Done in 900b8776. protocols and cipherSuites now use unmodifiable List<String> values, while the builder still keeps the varargs setters for convenience.

ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcTlsConfig.java
```diff
@@ -34,6 +37,10 @@
  */
 public class GrpcTlsConfig extends TlsConf {
   private final boolean fileBasedConfig;
+  private final SslProvider sslProvider;
+  private final String jsseProviderName;
+  private final List<String> protocols;
+  private final List<String> cipherSuites;
```
szetszwo: Let's move them up to TlsConf so ratis-netty could also use it.

(Indeed, GrpcTlsConfig is a poor designed class.  TlsConf is a replacement of it.  Unfortunately, we cannot remove GrpcTlsConfig for compatibility.)

ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcFactory.java
```diff
@@ -91,9 +89,10 @@ private SslContexts(GrpcTlsConfig tlsConfig, GrpcTlsConfig adminTlsConfig,
   }
 
   private final GrpcServices.Customizer servicesCustomizer;
-
-  private final Supplier<SslContexts> forServerSupplier;
-  private final Supplier<SslContexts> forClientSupplier;
```
szetszwo, line 96: We should keep using MemoizedSupplier.  Otherwise, the same SslContexts will be built multiple times; see RATIS-2331.

Indeed, all the change in GrpcFactory are not needed.

ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcTlsConfig.java
```diff
@@ -131,4 +158,124 @@ private static Builder newBuilder(File privateKeyFile, File certChainFile, File
   private static Builder newBuilder(KeyManager keyManager, TrustManager trustManager, boolean mTlsEnabled) {
     return newBuilder().setMutualTls(mTlsEnabled).setKeyManager(keyManager).setTrustManager(trustManager);
   }
-}
\ No newline at end of file
+
+  public static Builder newBuilder(GrpcTlsConfig conf) {
```
szetszwo: This is only used in testNewBuilderCopiesGrpcTlsOptions().  So, we should remove both this method and also testNewBuilderCopiesGrpcTlsOptions().

ratis-grpc/src/test/java/org/apache/ratis/grpc/TestGrpcTlsConfig.java
```diff
@@ -0,0 +1,82 @@
...
+    Assertions.assertEquals(Collections.singletonList("TLSv1.2"), updated.getProtocols());
+    Assertions.assertEquals(Arrays.asList("TLS_AES_128_GCM_SHA256"), updated.getCipherSuites());
+  }
+
+  @Test
+  public void testUnknownJsseProviderUsesGenericJdkConfiguration() {
+    final SslContextBuilder builder = GrpcUtil.configureJsseProvider(
+        SslContextBuilder.forClient(), new TestProvider());
+    Assertions.assertNotNull(builder);
+  }
```
szetszwo, line 56: This test seems incomplete?  Should we build it and expect an exception?

szetszwo (commented):
@HTHou , thanks for the update!

Please see th comments inlined and also https://issues.apache.org/jira/secure/attachment/13082326/1462_review.patch

ratis-grpc/src/test/java/org/apache/ratis/grpc/TestGrpcTlsConfig.java
HTHou: Done in 12dc7472. The test now calls build() and expects SSLException, so it verifies the fallback path more completely.

HTHou:
Thanks for the review. Addressed the latest comments in 12dc7472:

- Moved the TLS provider/protocol/cipher suite options up to TlsConf so they are not tied to GrpcTlsConfig.
- Restored the MemoizedSupplier-based SslContext construction in GrpcFactory.
- Removed GrpcTlsConfig.newBuilder(GrpcTlsConfig) and the test which only covered that helper.
- Updated the JSSE-provider fallback test to call build() and expect SSLException.

I also reran TestGrpcTlsConfig, checkstyle, spotbugs, and git diff --check.

szetszwo (approved):
+1 the change looks good.

PR #1474, opened 2026-05-29
RATIS-2547. Advance repliedIndex before completing write replies

CRZbulabula:
RATIS-2547

When `raft.server.read.option` uses `REPLIED_INDEX`, `ReplyFlusher` holds write reply callbacks and advances the leader replied index in batches. The old flow invoked reply callbacks while computing the max flushed log index, so a client could receive a successful write reply before `repliedIndex` covered that write.

This PR stores the log index separately from the reply callback, advances `repliedIndex` first, and only then completes the held write replies. It also makes `LeaderTracer.removePendingRequest` tolerate a null pending request for cached-reply paths.

A read-after-write workload can issue a linearizable read immediately after receiving a successful write reply. With `REPLIED_INDEX`, that read depends on the leader replied index. Completing the write reply before advancing `repliedIndex` creates a race where the read can wait unnecessarily or time out even though the write reply was already released.

## How was this patch tested?

`mvn -pl ratis-server,ratis-test -am -Dtest=TestReplyFlusher,TestLinearizableReadRepliedIndexWithGrpc test`

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
@@ -1252,18 +1252,17 @@ class LeaderStateImpl implements LeaderState {
     final PendingRequest pending = pendingRequests.remove(termIndex);
     leaderTracer.removePendingRequest(pending);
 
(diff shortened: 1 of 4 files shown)


szetszwo (approved):
+1 the change looks good.

PR #1475, opened 2026-05-29
RATIS-2548. Stabilize timing-sensitive Ratis tests

CRZbulabula:
RATIS-2548

This PR stabilizes several timing-sensitive tests by replacing fixed sleeps or immediate assertions with waits for the concrete condition each test needs.

The changes include:

* Wait for restart futures before printing logs and continuing to log assertions in `RaftBasicTests`.
* Keep the kill-leader restart scheduled before client messages are sent, and assert that the expected append messages are present in order while tolerating an extra retry/failover state-machine entry.
* Track async append replies with `List<CompletableFuture<RaftClientReply>>` and join each reply directly.
* Wait for commit index / state machine count in linearizable read tests instead of relying on fixed sleeps.
* Wait for the transaction context map to become empty in `RaftLogTruncateTests`.
* Pause the current leader before election stepDown and wait until a different leader is elected.
* Allow more time for the install-snapshot follower next-index assertion on slow CI.

These tests can fail on slower CI machines when asynchronous restart, log cleanup, state machine application, leadership transition, or install snapshot progress completes later than the fixed delay assumed by the test.

## How was this patch tested?

`mvn -pl ratis-test -am -Dtest=TestLinearizableReadRepliedIndexWithGrpc,TestRaftAsyncWithGrpc#testBasicAppendEntriesAsyncKillLeader,TestElectionCommandIntegrationWithGrpc,TestRaftLogTruncateWithGrpc,TestRaftWithGrpc test`

`mvn -pl ratis-test -am -Dtest=TestRaftReconfigurationWithSimulatedRpc#testKillLeaderDuringReconf test`

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
@@ -1154,6 +1154,10 @@ class LeaderStateImpl implements LeaderState {
 
     final RaftConfigurationImpl conf = server.getRaftConf();
 
+    if (conf.isSingleMode(server.getId())) {
+      return true;
+    }
+
     if (conf.hasMajority(activePeers, server.getId())) {
       // leadership check passed
       return true;
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
@@ -1154,10 +1154,6 @@ class LeaderStateImpl implements LeaderState {
 
     final RaftConfigurationImpl conf = server.getRaftConf();
 
-    if (conf.isSingleMode(server.getId())) {
-      return true;
-    }
-
     if (conf.hasMajority(activePeers, server.getId())) {
       // leadership check passed
       return true;
--- a/ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
+++ b/ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
@@ -115,6 +115,41 @@ public abstract class RaftBasicTests<CLUSTER extends MiniRaftCluster>
     return future;
   }
 
+  private static List<LogEntryProto> findLogEntriesContaining(
+      RaftLog log, long expectedTerm, SimpleMessage[] expectedMessages) {
+    final List<LogEntryProto> entries = RaftTestUtil.getStateMachineLogEntries(log, s -> {});
+    final List<LogEntryProto> matched = new ArrayList<>(expectedMessages.length);
+    int e = 0;
+    for (SimpleMessage expected : expectedMessages) {
+      boolean found = false;
+      for (; e < entries.size(); e++) {
+        final LogEntryProto entry = entries.get(e);
+        if (entry.getTerm() >= expectedTerm
+            && expected.getContent().equals(entry.getStateMachineLogEntry().getLogData())) {
+          matched.add(entry);
+          e++;
+          found = true;
+          break;
+        }
+      }
+      Assertions.assertTrue(found, () -> "Failed to find " + expected + " in entries " + entries);
+    }
+    return matched;
+  }
+
+  private static void assertLogEntriesContaining(
+      RaftServer.Division server, long expectedTerm, SimpleMessage[] expectedMessages, int numAttempts, Logger log)
+      throws Exception {
+    final String name = server.getId() + " assertLogEntriesContaining";
+    JavaUtils.attempt(() -> {
+          RaftTestUtil.assertLogEntries(
+              findLogEntriesContaining(server.getRaftLog(), expectedTerm, expectedMessages),
+              expectedTerm, expectedMessages);
+          return null;
+        },
+        numAttempts, TimeDuration.ONE_SECOND, () -> name, log);
+  }
+
   static void runTestBasicAppendEntries(
       boolean async, boolean killLeader, int numMessages, MiniRaftCluster cluster, Logger log)
       throws Exception {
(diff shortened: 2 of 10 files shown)


ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -1154,6 +1154,10 @@ public boolean checkLeadership() {
 
     final RaftConfigurationImpl conf = server.getRaftConf();
 
+    if (conf.isSingleMode(server.getId())) {
+      return true;
+    }
```
szetszwo: Let's do this change separately.  Then, this PR changes only the test code.

ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
```diff
@@ -127,46 +128,41 @@ static void runTestBasicAppendEntries(
...
-          Assertions.assertTrue(reply.isSuccess());
         }
       }
-      if (async) {
-        f.join();
-        Assertions.assertEquals(messages.length, asyncReplyCount.get());
+      if (killLeader) {
+        log.info("killAndRestart leader " + leader.getId());
+        killAndRestartLeader = killAndRestartServer(leader.getId(), 0, 4000, cluster, log);
       }
```
szetszwo: > Wait for async append replies before injecting the kill-leader restart in RaftBasicTests.

Before this change, killLeader is in the beginning.  This change moves it to the end.  It makes the test easier to pass but not fixing a bug.

It is good to test killLeader before client sending messages.  So, let's don't make this change?

ratis-test/src/test/java/org/apache/ratis/shell/cli/sh/ElectionCommandIntegrationTest.java
```diff
@@ -150,13 +150,16 @@ public void testElectionStepDownCommand() throws Exception {
...
     final StringPrintStream out = new StringPrintStream();
     RatisShell shell = new RatisShell(out.getPrintStream());
-    Assertions.assertNotEquals(cluster.getLeader().getId(), newLeader.getId());
     Assertions.assertEquals(2, cluster.getFollowers().size());
-    int ret = shell.run("election", "stepDown", "-peers", sb.toString());
+    int ret = shell.run("election", "pause", "-peers", sb.toString(), "-address",
+        leader.getPeer().getAddress());
+    Assertions.assertEquals(0, ret);
+
+    ret = shell.run("election", "stepDown", "-peers", sb.toString());
```
szetszwo: This change is good.  Could you also remove the redundant toString() calls?
```java
    int ret = shell.run("election", "pause", "-peers", sb, "-address", leader.getPeer().getAddress());
    Assertions.assertEquals(0, ret);

    ret = shell.run("election", "stepDown", "-peers", sb);
    Assertions.assertEquals(0, ret);
```

ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
```diff
@@ -127,46 +128,41 @@ static void runTestBasicAppendEntries(
...
+      Thread.sleep(cluster.getTimeoutMax().toIntExact(TimeUnit.MILLISECONDS) + 100);
+    } finally {
+      CompletableFuture.allOf(killAndRestartFollower, killAndRestartLeader).join();
     }
-    Thread.sleep(cluster.getTimeoutMax().toIntExact(TimeUnit.MILLISECONDS) + 100);
     log.info(cluster.printAllLogs());
-    killAndRestartFollower.join();
-    killAndRestartLeader.join();
```
szetszwo: > Wait for restart futures before continuing to log assertions in RaftBasicTests.

You are right that we should join before printing the log.

How about we simply move cluster.printAllLogs() up?  The try-finally make the code harder to read.
```java
     Thread.sleep(cluster.getTimeoutMax().toIntExact(TimeUnit.MILLISECONDS) + 100);
-    log.info(cluster.printAllLogs());
     killAndRestartFollower.join();
     killAndRestartLeader.join();
+    log.info(cluster.printAllLogs());
```

ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
```diff
@@ -127,46 +128,41 @@ static void runTestBasicAppendEntries(
...
+        }
         if (async) {
-          client.async().send(message).thenAcceptAsync(reply -> {
-            if (!reply.isSuccess()) {
-              f.completeExceptionally(
-                  new AssertionError("Failed with reply " + reply));
-            } else if (asyncReplyCount.incrementAndGet() == messages.length) {
-              f.complete(null);
-            }
+          CompletableFuture.allOf(asyncReplies.toArray(new CompletableFuture<?>[0])).join();
```
szetszwo: Since join() is called below.  This allOf is not needed.  Let's remove it.

BTW, changing
```java
      final AtomicInteger asyncReplyCount = new AtomicInteger();
      final CompletableFuture<Void> f = new CompletableFuture<>();
```
to
```java
      final List<CompletableFuture<RaftClientReply>> asyncReplies = new ArrayList<>();
```
does make the code easier to understand (although the original code is also correct.)

szetszwo (commented):
@CRZbulabula , thanks for fixing the tests!  Please see the comments inlined.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
CRZbulabula: Done. Removed the `LeaderStateImpl` change from this PR, so the current PR diff is test-only now. I will handle that leadership check separately.

ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
CRZbulabula: Done. Restored the kill-leader timing so the leader restart is scheduled before client messages are sent. The final kill-leader assertion now checks that the expected messages appear in order, while allowing an extra state-machine entry produced by retry/failover.

ratis-test/src/test/java/org/apache/ratis/shell/cli/sh/ElectionCommandIntegrationTest.java
CRZbulabula: Done. Removed the redundant `toString()` calls.

ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
CRZbulabula: Done. Removed the try/finally restructuring and now join the restart futures before `cluster.printAllLogs()`.

ratis-server/src/test/java/org/apache/ratis/RaftBasicTests.java
CRZbulabula: Done. Removed the redundant `CompletableFuture.allOf(...).join()` and kept the reply list cleanup.

CRZbulabula:
Thanks for the review. I updated the PR to address the inline comments:

* Removed the `LeaderStateImpl` change from this PR; this PR is test-only now.
* Restored the kill-leader timing so leader restart is scheduled before client messages are sent.
* Kept the async reply list cleanup and removed the redundant `allOf` join.
* Joined restart futures before `cluster.printAllLogs()` without the try/finally restructuring.
* Removed the redundant `toString()` calls in the election command test.
* For kill-leader append tests, the final assertion now verifies the expected messages appear in order while tolerating an extra retry/failover state-machine entry.

Local verification:

`mvn -pl ratis-test -am -Dtest=TestLinearizableReadRepliedIndexWithGrpc,TestRaftAsyncWithGrpc#testBasicAppendEntriesAsyncKillLeader,TestElectionCommandIntegrationWithGrpc,TestRaftLogTruncateWithGrpc,TestRaftWithGrpc test`

`mvn -pl ratis-test -am -Dtest=TestRaftReconfigurationWithSimulatedRpc#testKillLeaderDuringReconf test`

CRZbulabula:
Pushed `fb3ee4ee` to address the latest CI failures.

Summary:

- `TestManualRestoreSnapshot` now verifies that the restarted follower has restored and applied at least the saved snapshot index, instead of racing against the leader current lastApplied value.
- `testAddServerForWaitReady` now waits until the new server impls report running after the ADD configuration change, and always cleans up the START_COMPLETE injection in a finally block.

Local verification:

- `mvn -pl ratis-examples -am -Dtest=TestManualRestoreSnapshot test`
- `mvn -pl ratis-test -am -Dtest=TestLeaderElectionWithSimulatedRpc#testAddServerForWaitReady,TestLeaderElectionWithGrpc#testAddServerForWaitReady,TestLeaderElectionWithNetty#testAddServerForWaitReady test`

GitHub Actions for `fb3ee4ee` are green; coverage is skipped as expected.

ratis-server/src/test/java/org/apache/ratis/server/impl/LeaderElectionTests.java
```diff
@@ -148,29 +148,32 @@ public void testAddServerForWaitReady() throws IOException, InterruptedException
...
+          for (RaftServer server : cluster.getServers()) {
+            RaftServerProxy proxy = (RaftServerProxy) server;
+            proxy.getImpls().forEach(s -> assertTrue(s.isRunning()));
+          }
+        }, 10, HUNDRED_MILLIS, "all server impls running", LOG);
       }
+    } finally {
+      cluster.shutdown();
+      CodeInjectionForTesting.remove(RaftServerImpl.START_COMPLETE);
     }
```
szetszwo, line 176: Adding finally is good but the double try-block is not needed.  Since this is only a test, I am fine with the current change.
```diff
+++ b/ratis-server/src/test/java/org/apache/ratis/server/impl/LeaderElectionTests.java
@@ -162,15 +162,16 @@ public abstract class LeaderElectionTests<CLUSTER extends MiniRaftCluster>
               .setServersInNewConf(peerChanges.getAddedPeers())
               .setMode(SetConfigurationRequest.Mode.ADD).build());
       assertTrue(reply.isSuccess());
-      for (RaftServer server : cluster.getServers()) {
-        RaftServerProxy proxy = (RaftServerProxy) server;
-        proxy.getImpls().forEach(s -> {
-          assertTrue(s.isRunning());
-        });
-      }
+      JavaUtils.attempt(() -> {
+        for (RaftServer server : cluster.getServers()) {
+          RaftServerProxy proxy = (RaftServerProxy) server;
+          proxy.getImpls().forEach(s -> assertTrue(s.isRunning()));
+        }
+      }, 10, HUNDRED_MILLIS, "all server impls running", LOG);
+    } finally {
+      cluster.shutdown();
+      CodeInjectionForTesting.remove(RaftServerImpl.START_COMPLETE);
     }
-    cluster.shutdown();;
-    CodeInjectionForTesting.remove(RaftServerImpl.START_COMPLETE);
   }
```

szetszwo (approved):
@CRZbulabula , thanks for the update!

+1 the change looks good.

PR #1476, opened 2026-06-03
Query counter bug fix - query counter in example server

haroon3rd:

This patch fixes an issue in the membership example where the `query` command displayed an empty counter value even after successful counter updates.

The root cause was an encoding mismatch between the server and client implementations. The `CounterStateMachine` returns the counter value as a 4-byte binary integer (`ByteString`), while the membership example client attempted to decode the response using UTF-8 string conversion. As a result, the query output appeared empty.

This patch updates the `queryCounter()` implementation in `RaftCluster` to correctly deserialize the binary response using `ByteBuffer.getInt()`, allowing the current counter value to be displayed correctly.

Apache JIRA: RATIS-XXXX

(Replace with the actual JIRA issue link after creating the issue.)

## How was this patch tested?

The fix was verified manually using the Ratis membership example.

Test procedure:

1. Build and run the membership example.
2. Start a cluster with multiple peers.
3. Execute the `incr` command multiple times.
4. Execute the `query` command.

Before the fix:

```text
>>> incr
>>> incr
>>> query
Current counter value:
```

After the fix:

```text
>>> incr
>>> incr
>>> query
Current counter value: 2
```

The returned counter value now correctly reflects the state maintained by `CounterStateMachine`.

### DIFF
--- a/ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/RaftCluster.java
+++ b/ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/RaftCluster.java
@@ -143,15 +143,24 @@ public class RaftCluster {
   }
 
   public void queryCounter() throws IOException {
-    RaftClient client = createClient();
-    try {
-      RaftClientReply reply = client.io().sendReadOnly(CounterCommand.GET.getMessage());
-      String count = reply.getMessage().getContent().toStringUtf8();
-      System.out.println("Current counter value: " + count);
-    } finally {
-      client.close();
-    }
+  RaftClient client = createClient();
+  try {
+    RaftClientReply reply = client.io().sendReadOnly(
+        CounterCommand.GET.getMessage());
+
+    ByteBuffer buffer =
+        reply.getMessage()
+             .getContent()
+             .asReadOnlyByteBuffer();
+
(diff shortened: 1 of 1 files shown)


szetszwo:
@haroon3rd , thanks a lot for working on this!  Could you file a [JIRA](https://issues.apache.org/jira/browse/RATIS)?  In Ratis, we use JIRA system to track issues.

(Please request a JIRA account if you do not already have one.  We can approve it.)

ratis-examples/src/main/java/org/apache/ratis/examples/membership/server/RaftCluster.java
```diff
@@ -143,15 +143,24 @@ public void counterIncrement() throws IOException {
...
   public void queryCounter() throws IOException {
-    RaftClient client = createClient();
-    try {
-      RaftClientReply reply = client.io().sendReadOnly(CounterCommand.GET.getMessage());
-      String count = reply.getMessage().getContent().toStringUtf8();
-      System.out.println("Current counter value: " + count);
-    } finally {
-      client.close();
-    }
+  RaftClient client = createClient();
```
amaliujia, line 146: Nit: wrong indents

PR #1481, opened 2026-06-13
RATIS-2546. Add input stream to DataStreamApi for read operations in Client

peterxcli:

* Added a new `DataStreamInput` interface and its implementation `DataStreamInputImpl` for asynchronous, zero-copy read-only data streaming, including the `readAsync()` method for consuming replies and proper resource release on close.
* Extended the `DataStreamApi` interface and its implementation to provide `streamReadOnly()` methods for creating read-only streams.
* Introduced the new `DataStreamReplyByteBuf` class to support replies backed by Netty `ByteBuf`.
* Added a new overload of `streamAsync()` in the `DataStreamClientRpc` interface that takes a reply consumer, supporting multiple replies per request.

https://issues.apache.org/jira/browse/RATIS-2546

## How was this patch tested?


peterxcli:
cc @szetszwo

peterxcli:
The no-MD5 sequential reads comparison:
Key size | Buffer | ReadBlock stream | Ratis data stream | Ratis / ReadBlock bandwidth
-- | -- | -- | -- | --
256 MiB | 32 MiB | 333.23 MB/s | 448.04 MB/s | 1.34x bandwidth
256 MiB | 8 MiB | 651.57 MB/s | 753.94 MB/s | 1.16x bandwidth
256 MiB | 1 MiB | 346.72 MB/s | 726.91 MB/s | 2.10x bandwidth
256 MiB | 4 KiB | 303.35 MB/s | 777.43 MB/s | 2.56x bandwidth
500 MiB | 32 MiB | 731.13 MB/s | 445.67 MB/s | 0.61x bandwidth
500 MiB | 8 MiB | 587.45 MB/s | 1012.59 MB/s | 1.72x bandwidth
500 MiB | 1 MiB | 520.76 MB/s | 598.70 MB/s | 1.15x bandwidth
500 MiB | 4 KiB | 373.36 MB/s | 846.50 MB/s | 2.27x bandwidth
1 GiB | 32 MiB | 869.43 MB/s | 693.04 MB/s | 0.80x bandwidth
1 GiB | 8 MiB | 508.69 MB/s | 474.92 MB/s | 0.93x bandwidth
1 GiB | 1 MiB | 432.13 MB/s | 784.25 MB/s | 1.81x bandwidth
1 GiB | 4 KiB | 378.85 MB/s | 1087.56 MB/s | 2.87x bandwidth

The no-MD5 random reads comparison:

| Key size | Random read size | Direct stream | Ratis data stream | Bandwidth ratio | IOPS ratio | Direct elapsed | Ratis elapsed |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 256 MiB | 1 MiB | 28.92 MB/s | 103.03 MB/s | 3.56x bandwidth | 3.56x IOPS | 1.106 s | 0.311 s |
| 256 MiB | 4 KiB | 0.12 MB/s | 0.38 MB/s | 3.19x bandwidth | 3.19x IOPS | 1.059 s | 0.332 s |
| 500 MiB | 1 MiB | 44.18 MB/s | 255.86 MB/s | 5.79x bandwidth | 5.79x IOPS | 0.724 s | 0.125 s |
| 500 MiB | 4 KiB | 0.14 MB/s | 0.59 MB/s | 4.06x bandwidth | 4.06x IOPS | 0.865 s | 0.213 s |
| 1 GiB | 1 MiB | 33.08 MB/s | 193.64 MB/s | 5.85x bandwidth | 5.85x IOPS | 0.967 s | 0.165 s |
| 1 GiB | 4 KiB | 0.10 MB/s | 0.87 MB/s | 8.34x bandwidth | 8.34x IOPS | 1.197 s | 0.144 s |

I'm looking into why larger buffer would cause throughput to degrade.

ratis-common/src/main/java/org/apache/ratis/datastream/impl/DataStreamReplyByteBuf.java
```diff
@@ -0,0 +1,174 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Implements {@link DataStreamReply} with {@link ByteBuf}.
+ */
+public final class DataStreamReplyByteBuf extends DataStreamPacketImpl implements DataStreamReply {
```
szetszwo: Similar to DataStreamPacketByteBuffer, Let's add a new DataStreamPacketByteBuf class.

ratis-common/src/main/java/org/apache/ratis/datastream/impl/DataStreamReplyByteBuf.java
```diff
@@ -0,0 +1,174 @@
...
+  public static final class Builder {
```
szetszwo: Let's add a new DataStreamReplyBuilder class to share the code with DataStreamReplyByteBuffer.Builder.

ratis-common/src/main/java/org/apache/ratis/datastream/impl/DataStreamReplyByteBuf.java
```diff
@@ -0,0 +1,174 @@
...
+      return new DataStreamReplyByteBuf(
+          clientId, type, streamId, streamOffset, buffer, success, bytesWritten, commitInfos);
+    }
+  }
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  private final AtomicReference<ByteBuf> buffer;
```
szetszwo: Let's rename it to buf for ByteBuf.  It is easier to distinguish it with ByteBuffer.

szetszwo (commented):
@peterxcli , thanks a lot for working on this!

The PR is quite big.  Let's separate the DataStreamReply change first.  Please see the comments inlined and also  ~https://issues.apache.org/jira/secure/attachment/13082807/1481_review_DataStreamReply.patch~ (updated the link in the next comment.)

ratis-common/src/main/java/org/apache/ratis/protocol/DataStreamReply.java
```diff
@@ -22,12 +22,23 @@
...
+  /**
+   * Release resources owned by this reply.
+   */
+  default void release() {
+  }
+
+  @Override
+  default void close() {
+    release();
+  }
```
szetszwo: On a second thought, let's don't add release() and close() methods to DataStreamReply since 
- DataStreamReplyByteBuffer does not need them, and
- DataStreamRequest also does not have these methods.

See https://issues.apache.org/jira/secure/attachment/13082808/1481_review2_DataStreamReply.patch

peterxcli: @szetszwo thanks for the review and further refactor suggestion!
I've opened a separate PR for introducing new `DataStreamReply` for `ByteBuf` as your provided patch: https://github.com/apache/ratis/pull/1485
please take a look. Thanks!

szetszwo: Since DataStreamReply is a public API, we cannot add release() and close() methods; otherwise it is an incompatible change.

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
szetszwo: Since we have changed the decoder, we could start using DataStreamReplyByteBuf here.
```java
      public void channelRead(ChannelHandlerContext ctx, Object msg) {
        if (!(msg instanceof DataStreamReplyByteBuf)) {
          LOG.error("{}: unexpected message {}", name, msg.getClass());
          return;
        }
        try (DataStreamReplyByteBuf reply = (DataStreamReplyByteBuf) msg) {
          process(reply);
        }
      }
```

ratis-netty/src/main/java/org/apache/ratis/netty/server/ReadStreamManagement.java
```diff
@@ -48,13 +55,15 @@ public class ReadStreamManagement {
   static class ReadStream implements WritableByteChannel {
     private final ClientId clientId;
     private final long streamId;
+    private final RaftClientRequest request;
```
szetszwo: Store the terminalReply instead of the request:
```java
    private final DataStreamReplyByteBuffer terminalReply;
    private long streamOffset;

    ReadStream(RaftClientRequest request, long streamId, ChannelHandlerContext ctx) {
      this.clientId = request.getClientId();
      this.streamId = streamId;
      this.ctx = ctx;

      final RaftClientReply reply = RaftClientReply.newBuilder()
          .setRequest(request)
          .setSuccess()
          .build();
      this.terminalReply = DataStreamReplyByteBuffer.newBuilder()
          .setClientId(clientId)
          .setType(Type.STREAM_HEADER)
          .setStreamId(streamId)
          .setStreamOffset(0)
          .setBuffer(toRaftClientReplyProto(reply).toByteString().asReadOnlyByteBuffer())
          .setSuccess(true)
          .setBytesWritten(0)
          .build();
    }
```

szetszwo (commented):
@peterxcli , thanks for updating this!

Let's have one more simple change to add an executor and terminalReply (RATIS-2564).  Please see the comments inlined and also https://issues.apache.org/jira/secure/attachment/13082846/1481_review3_refactoring.patch

ratis-client/src/main/java/org/apache/ratis/client/DataStreamClientRpc.java
```diff
@@ -36,4 +37,11 @@ default CompletableFuture<DataStreamReply> streamAsync(DataStreamRequest request
     throw new UnsupportedOperationException(getClass() + " does not support "
         + JavaUtils.getCurrentStackTraceElement().getMethodName());
   }
+
+  /** Async call to send a request and receive multiple replies for the request. */
+  default CompletableFuture<DataStreamReply> streamAsync(
+      DataStreamRequest request, Consumer<DataStreamReply> replyConsumer) {
```
szetszwo: @peterxcli ,  Let's borrow the idea from [gRPC StreamObserver](https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/StreamObserver.html) and add a similar interface.  It is more flexible for future changes.
```java
//ratis-common
package org.apache.ratis.datastream;

/** An interface similar to gRPC {@link org.apache.ratis.thirdparty.io.grpc.stub.StreamObserver}. */
public interface DataStreamObserver<V> {
  void onNext(V value);

  // see if onError(Throwable) and onCompleted() are useful.  Or we may add them later.
}
```

Then our interface will be similar to gRPC service https://github.com/apache/ratis/blob/master/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcServerProtocolClient.java#L174-L181

```java
//DataStreamClientRpc
  /** Async call to send a request and receive multiple replies for the request. */
  default CompletableFuture<DataStreamReply> streamAsync(DataStreamRequest request,
      DataStreamObserver<DataStreamReplyByteBuf> replyHandler) {
    throw new UnsupportedOperationException(getClass() + " does not support "
        + JavaUtils.getCurrentStackTraceElement().getMethodName());
  }
```

peterxcli:
@szetszwo thanks for the grpc observer suggestion, just added a initial version of it.

I also add onError and onComplete interface method for the observer, because we can decide which hook to call base on the terminal reply.

Please take a look, Thanks!

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
```diff
@@ -236,6 +244,114 @@ private CompletableFuture<DataStreamReply> sendForward(DataStreamReply writeRepl
...
+      final DataStreamRequestHeader h = new DataStreamRequestHeader(header.getClientId(), Type.STREAM_HEADER,
+          header.getCallId(), 0, buffer.remaining(), StandardWriteOption.FLUSH, StandardWriteOption.CLOSE);
+      this.replyFuture = dataStreamClientRpc.streamAsync(new DataStreamRequestByteBuffer(h, buffer),
+          reply -> receive(reply.copy()));
+      replyFuture.whenComplete((reply, exception) -> {
+        if (exception != null) {
+          failReads(exception);
+        } else {
+          markEndOfStream();
+        }
```
amaliujia: Do we need to call `close()` or mark `closed = true;` here?

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
```diff
@@ -236,6 +244,114 @@ private CompletableFuture<DataStreamReply> sendForward(DataStreamReply writeRepl
...
+        return JavaUtils.completeExceptionally(newEndOfStreamException());
+      }
+      final CompletableFuture<DataStreamReply> f = new CompletableFuture<>();
+      pendingReads.add(f);
+      return f;
+    }
+
+    @Override
+    public synchronized void close() {
+      closed = true;
```
amaliujia: Add a `if (closed) return` to maintain idempotent close?

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
```diff
@@ -236,6 +244,114 @@ private CompletableFuture<DataStreamReply> sendForward(DataStreamReply writeRepl
...
+      for (DataStreamReply reply; (reply = replies.poll()) != null;) {
+        DataStreamReplyByteBuf.release(reply);
+      }
+      failReads(new AlreadyClosedException(clientId + ": stream already closed, request=" + header));
```
amaliujia: I do not understand why we need to do a `failReads` upon a ordinary close. Do we not expect `close` be call in the happy path and we only call close in bad cases?

ratis-common/src/main/java/org/apache/ratis/datastream/impl/DataStreamReplyByteBuf.java
```diff
@@ -118,4 +118,10 @@ static ByteBuffer copy(ByteBuf buf) {
     buf.readBytes(bytes);
     return ByteBuffer.wrap(bytes);
   }
+
+  public static void release(DataStreamReply reply) {
+    if (reply instanceof DataStreamReplyByteBuf) {
+      ((DataStreamReplyByteBuf) reply).release();
+    }
```
amaliujia: What about else? Is that possible other instances are passed here? If no maybe add an assertion, otherwise handle other instances?

peterxcli: There are two classes implement the `DataStreamReply`, which is `DataStreamReplyByteBuf` and `DataStreamReplyByteBuffer`.
And only `DataStreamReplyByteBuf` need to release the Netty byteBuffer reference count

https://github.com/apache/ratis/blob/3cb6794bcb9fcd35dfd8bd67a885a70cc9de7a56/ratis-common/src/main/java/org/apache/ratis/datastream/impl/DataStreamPacketByteBuf.java#L58-L63

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
peterxcli: no, the caller might have not counsumed the fetched content yet, if we set close, the `readAsync` would fail,

peterxcli: I think I can refactor this to get better readability.

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
peterxcli: `close()` could happen while `readAsync()` calls are pending.

ratis-common/src/main/java/org/apache/ratis/datastream/DataStreamObserver.java
```diff
@@ -0,0 +1,30 @@
...
+/** An interface similar to gRPC {@link org.apache.ratis.thirdparty.io.grpc.stub.StreamObserver}. */
+@FunctionalInterface
+public interface DataStreamObserver<V> {
+  void onNext(V value);
+
+  default void onError(Throwable t) {
+  }
+
+  default void onCompleted() {
+  }
```
szetszwo: If we add them, we should enforce implementing them.  Let's remove default.
```java
  void onCompleted();

  void onError(Throwable throwable);
```

ratis-client/src/main/java/org/apache/ratis/client/api/DataStreamInput.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface DataStreamInput extends Closeable {
+  /**
+   * Read the next chunk in the stream asynchronously.
+   * The caller owns the returned {@link DataStreamReply} and should call
+   * {@link DataStreamReply#release()} after consuming it.
+   *
+   * @return a future of the reply.
+   */
+  CompletableFuture<DataStreamReply> readAsync();
```
szetszwo: We should use reference-counted:
```java
  /**
   * Read the next chunk in the stream asynchronously.
   * The caller owns the returned {@link DataStreamReply} which is a {@link ReferenceCountedObject}.
   * It must call {@link ReferenceCountedObject#release()} after consuming it.
   *
   * @return a future of the reference-counted reply.
   */
  CompletableFuture<ReferenceCountedObject<DataStreamReply>> readAsync();
```

ratis-client/src/main/java/org/apache/ratis/client/DataStreamClientRpc.java
```diff
@@ -36,4 +38,11 @@ default CompletableFuture<DataStreamReply> streamAsync(DataStreamRequest request
     throw new UnsupportedOperationException(getClass() + " does not support "
         + JavaUtils.getCurrentStackTraceElement().getMethodName());
   }
+
+  /** Async call to send a request and receive multiple replies for the request. */
+  default CompletableFuture<DataStreamReply> streamAsync(
+      DataStreamRequest request, DataStreamObserver<DataStreamReplyByteBuf> replyHandler) {
+    throw new UnsupportedOperationException(getClass() + " does not support "
+        + JavaUtils.getCurrentStackTraceElement().getMethodName());
+  }
```
szetszwo: We should use reference-counted:
```java
  /**
   * Async call to send a request to receive a stream of intermediate replies and a final reply.
   *
   * @param request the request
   * @param replyHandler to handle intermediate replies
   * @return a future the final reply
   */
  default CompletableFuture<DataStreamReply> streamAsync(DataStreamRequest request,
      DataStreamObserver<ReferenceCountedObject<DataStreamReply>> replyHandler) {
    throw new UnsupportedOperationException(getClass() + " does not support "
        + JavaUtils.getCurrentStackTraceElement().getMethodName());
  }
```

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
szetszwo: This will release the reply even before user application reading it.  We should use reference-counted:
```java
        final DataStreamReplyByteBuf reply = (DataStreamReplyByteBuf) msg;
        final ReferenceCountedObject<DataStreamReply> ref = ReferenceCountedObject.<DataStreamReply>newBuilder()
            .setValue((DataStreamReplyByteBuf) msg)
            .setReleaseMethod(r -> {
              if (r != null) {
                Preconditions.assertSame(reply, r, "reply");
                reply.release();
              }
            }).build();
        try (UncheckedAutoCloseableSupplier<DataStreamReply> ignored = ref.retainAndReleaseOnClose()) {
          process(ref);
        }
```

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
```diff
@@ -236,6 +245,132 @@ private CompletableFuture<DataStreamReply> sendForward(DataStreamReply writeRepl
...
+    private DataStreamInputImpl(RaftClientRequest request) {
+      this.header = request;
+      final ByteBuffer buffer = ClientProtoUtils.toRaftClientRequestProtoByteBuffer(header);
+      final DataStreamRequestHeader h = new DataStreamRequestHeader(header.getClientId(), Type.STREAM_HEADER,
+          header.getCallId(), 0, buffer.remaining(), StandardWriteOption.FLUSH, StandardWriteOption.CLOSE);
+      this.replyFuture = dataStreamClientRpc.streamAsync(new DataStreamRequestByteBuffer(h, buffer),
+          new DataStreamObserver<DataStreamReplyByteBuf>() {
+            @Override
+            public void onNext(DataStreamReplyByteBuf reply) {
+              receive(reply.copy());
```
szetszwo: We should use reference-counted in the queues and avoid copying the reply.

szetszwo (commented):
@peterxcli , thanks for the update!

In DataStreamInput, readAsync() should return a ReferenceCountedObject so the user can retain/release the buffer.

Please see the comments inlined and also https://issues.apache.org/jira/secure/attachment/13082885/1481_review4.patch

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
amaliujia: It reads much better, thanks!

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamInputImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    reply.retain();
+    for (CompletableFuture<ReferenceCountedObject<DataStreamReply>> pending;
+         (pending = pendingReads.poll()) != null; ) {
+      if (pending.complete(reply)) {
+        return;
+      }
+    }
+    replies.add(reply);
```
szetszwo, line 80: Since all entries in pendingReads must not be completed, remove the loop.
```java
    reply.retain();
    final CompletableFuture<ReferenceCountedObject<DataStreamReply>> pending = pendingReads.poll();
    if (pending != null) {
      final boolean completed = pending.complete(reply);
      Preconditions.assertTrue(completed);
      return;
    }
    replies.add(reply);
```

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamInputImpl.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  @Override
+  public synchronized void onError(Throwable throwable) {
+    // An error case, release the replies
```
szetszwo, line 86: Add requireNonNull:
```
    Objects.requireNonNull(throwable, "throwable == null");
```

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamInputImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    DataStreamObserver<ReferenceCountedObject<DataStreamReply>> {
+  private final RaftClientRequest header;
+  private final ClientId clientId;
+  private final Queue<ReferenceCountedObject<DataStreamReply>> replies = new LinkedList<>();
+  private final Queue<CompletableFuture<ReferenceCountedObject<DataStreamReply>>> pendingReads = new LinkedList<>();
+
+  /*
+   * null                  : the stream is open.
+   * AlreadyClosedException: the stream is closed.
+   * Other exception       : the stream is failed.
```
szetszwo, line 53: Add EOFException:
```java
   * null                  : the stream is open.
   * EOFException          : the stream is ended, i.e. no more data.
   * AlreadyClosedException: the stream is closed by the caller.
   * Other exception       : the stream is failed.
```

ratis-netty/src/main/java/org/apache/ratis/netty/server/ReadStreamManagement.java
```diff
@@ -64,20 +64,19 @@ static class ReadStream implements WritableByteChannel {
...
+        .build();
       this.terminalReply = DataStreamReplyByteBuffer.newBuilder()
-          .setClientId(clientId)
-          .setType(Type.STREAM_HEADER)
-          .setStreamId(streamId)
-          .setStreamOffset(0)
-          .setBuffer(toRaftClientReplyProto(reply).toByteString().asReadOnlyByteBuffer())
-          .setSuccess(true)
-          .setBytesWritten(0)
-          .build();
```
szetszwo, line 80: All the changes in ReadStreamManagement are whitespace change.  Let's revert them since this change is already quite big.

ratis-common/src/main/java/org/apache/ratis/protocol/DataStreamReply.java
```diff
@@ -30,4 +30,4 @@ public interface DataStreamReply extends DataStreamPacket {
 
   /** @return the commit information when the reply is created. */
   Collection<CommitInfoProto> getCommitInfos();
-}
\ No newline at end of file
+}
```
szetszwo: Revert white space change.

ratis-common/src/main/java/org/apache/ratis/datastream/impl/DataStreamReplyByteBuf.java
```diff
@@ -118,4 +118,10 @@ static ByteBuffer copy(ByteBuf buf) {
...
+  }
```
szetszwo: Just found that this is used only in a test.  Let remove it.

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
```diff
@@ -358,15 +364,40 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
...
-          process(reply);
+        final DataStreamReplyByteBuf reply = (DataStreamReplyByteBuf) msg;
+        final ReferenceCountedObject<DataStreamReply> ref = ReferenceCountedObject.<DataStreamReply>newBuilder()
+            .setValue((DataStreamReplyByteBuf) msg)
+            .setReleaseMethod(r -> {
+              if (r != null) {
+                Preconditions.assertSame(reply, r, "reply");
+                reply.release();
+              }
+            }).build();
```
szetszwo: Let's change Preconditions.assertSame to return DataStreamReplyByteBuf:
```java
        final ReferenceCountedObject<DataStreamReply> ref = ReferenceCountedObject.<DataStreamReply>newBuilder()
            .setValue((DataStreamReplyByteBuf) msg)
            .setReleaseMethod(r -> Preconditions.assertSame(reply, r, "reply").release())
            .build();
```
```diff
+++ b/ratis-common/src/main/java/org/apache/ratis/util/Preconditions.java
@@ -88,9 +88,10 @@ public interface Preconditions {
         () -> name + ": expected == " + expected + " but computed == " + computed);
   }
 
-  static void assertSame(Object expected, Object computed, String name) {
+  static <T> T assertSame(T expected, Object computed, String name) {
     assertTrue(expected == computed,
         () -> name + ": expected == " + expected + " but computed == " + computed);
+    return expected;
   } 
```

ratis-test/src/test/java/org/apache/ratis/client/impl/TestDataStreamClientImpl.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    RaftClientRequest getRequest() {
+      return request.get();
+    }
+
+    void receive(DataStreamReplyByteBuf reply) {
+      final ReferenceCountedObject<DataStreamReply> ref = ReferenceCountedObject.<DataStreamReply>newBuilder()
+          .setValue(reply)
+          .setReleaseMethod(DataStreamReplyByteBuf::release)
```
szetszwo: Use assertSame:
```java
          .setReleaseMethod(r -> Preconditions.assertSame(reply, r, "reply").release())
```

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
```diff
@@ -513,6 +544,54 @@ public CompletableFuture<DataStreamReply> streamAsync(DataStreamRequest request)
     return f;
   }
 
+  @Override
+  public CompletableFuture<DataStreamReply> streamAsync(DataStreamRequest request,
+      DataStreamObserver<ReferenceCountedObject<DataStreamReply>> replyHandler) {
+    final CompletableFuture<DataStreamReply> f = new CompletableFuture<>();
+    final ClientInvocationId clientInvocationId = ClientInvocationId.valueOf(request);
+    final ClientReadStream replyEntry = new ClientReadStream(request, f, replyHandler);
```
szetszwo: Let's rename it to `readStream`.

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
```diff
@@ -513,6 +544,54 @@ public CompletableFuture<DataStreamReply> streamAsync(DataStreamRequest request)
...
+    }
+
+    final ChannelFuture channelFuture;
+    final Channel channel;
+    LOG.debug("{}: write read-only stream begin {}", this, request);
+    synchronized (replyEntry) {
+      channel = connection.getChannelUninterruptibly();
+      if (channel == null) {
+        readStreams.remove(clientInvocationId, replyEntry);
+        f.completeExceptionally(new AlreadyClosedException(this + ": Failed to send " + request));
```
szetszwo: Let's change the message to "Failed to getChannel":
```java
        f.completeExceptionally(new AlreadyClosedException(this + ": Failed to getChannel for " + request));
```

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
```diff
@@ -513,6 +544,54 @@ public CompletableFuture<DataStreamReply> streamAsync(DataStreamRequest request)
...
 
+  @Override
+  public CompletableFuture<DataStreamReply> streamAsync(DataStreamRequest request,
+      DataStreamObserver<ReferenceCountedObject<DataStreamReply>> replyHandler) {
+    final CompletableFuture<DataStreamReply> f = new CompletableFuture<>();
+    final ClientInvocationId clientInvocationId = ClientInvocationId.valueOf(request);
+    final ClientReadStream replyEntry = new ClientReadStream(request, f, replyHandler);
+    if (readStreams.putIfAbsent(clientInvocationId, replyEntry) != null) {
+      f.completeExceptionally(new AlreadyClosedException(this + ": A read-only stream already exists for "
+          + clientInvocationId));
```
szetszwo: Let's include the request:
```java
      f.completeExceptionally(new AlreadyClosedException(this + ": Read stream already exists for "
          + clientInvocationId + ", request=" + request));
```

szetszwo (commented):
@peterxcli , thanks for the update! 

- Found that all the checkstyle exclude and findbugs exclude are not needed.  Please revert them.
- See the comments inlined and also https://issues.apache.org/jira/secure/attachment/13082908/1481_review5.patch

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
szetszwo: Oops, this actually is not working.  Need to keep check not null first.
```java
   .setReleaseMethod(r -> {
              if (r != null) {
                Preconditions.assertSame(reply, r, "reply");
                reply.release();
              }
            }).build();
```

peterxcli:
@szetszwo found a new class `AsyncQueue` in the review patch, curious is that accidentally included or if you want to use it to refactor sth?

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamInputImpl.java
peterxcli: But if the `onNext` API caller cancel the one of the future, the other future might never get the reply

Related testcase added in this patch:
https://github.com/apache/ratis/blob/05e55403a5ef53d29be4afb9c197e83f56d3ac03/ratis-test/src/test/java/org/apache/ratis/client/impl/TestDataStreamClientImpl.java#L112-L141

And since you assertion for that case, should we remove the above testcase and add document and tell user DONT CANCEL the return future?

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientStreamRpc.java
peterxcli: Thanks, I plan to add as static RC builder function for `DataStreamReply`:

```java
public static ReferenceCountedObject<DataStreamReply> asReferenceCounted(DataStreamReplyByteBuf reply) {
  Objects.requireNonNull(reply, "reply == null");
  return ReferenceCountedObject.<DataStreamReply>newBuilder()
      .setValue(reply)
      .setReleaseMethod(r -> {
        if (r != null) {
          Preconditions.assertSame(reply, r, "reply").release();
        }
      })
      .build();
}
```
so prod and test can shared the same RC-DataStreamReply constructor code

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamInputImpl.java
peterxcli: I removed the `testReceiveSkipsCancelledPendingRead` first, if you think we should change the api behaviour, feel free to ask me to add it back. Thanks!

https://github.com/apache/ratis/pull/1481/changes/3c8c4780ca4631a169204fce51920eaddb387da9

peterxcli:
@szetszwo Thanks for the review again! If you have any idea on further splitting for this PR, please feel free to ask me to do that!

szetszwo (approved):
+1 the change looks good.

szetszwo:
@peterxcli , thanks a lot for working hard on this!

@amaliujia , thanks for reviewing this!

peterxcli:
Thanks @szetszwo for continuous review and refactor suggestion!
Thanks @amaliujia for the review!

PR #1493, opened 2026-06-25
RATIS-2572. Empty dir annotations/ in ratis-proto.jar

adoroszlai:

`maven-compiler-plugin` creates the directory for sources generated by annotation processors (defined by [`<generatedSourcesDirectory>`](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#generatedSourcesDirectory), with default value `${project.build.directory}/generated-sources/annotations`) even if there are no processors to run.  This will be fixed by https://github.com/apache/maven-compiler-plugin/commit/d958a62 in 4.0.0.

`ratis-proto` also has sources generated for proto definitions.  It is not clear why the empty `generated-sources/annotations` ends up in `ratis-proto.jar` in some environments, but not in others.

As a workaround, this PR sets the `<generatedSourcesDirectory>` to `target/generated-sources`, which is created anyway for `ratis-proto`.

Also add `<proc>none</proc>` to ensure annotation processors are indeed not used.

https://issues.apache.org/jira/browse/RATIS-2572

## How was this patch tested?

Verified no `annotations/` in `generated-sources`:

```bash
$ ./mvnw -am -pl :ratis-proto -DskipTests clean verify
...

$ ls ratis-proto/target/generated-sources
org
```

`ratis-proto.jar` does not contain `annotations/` for me even without this patch.  I hope @OneSizeFitsQuorum, the release manager of 3.2.2, can verify it locally.

CI:
https://github.com/adoroszlai/ratis/actions/runs/28187818390

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -526,6 +526,11 @@
             <compilerArgs>
               <arg>-Xlint:all,-options,-path</arg>
             </compilerArgs>
+            <!-- hopefully can be removed after upgrade to maven-compiler-plugin 4.0 or later,
+                 if empty annotations/ dir is fixed by: https://github.com/apache/maven-compiler-plugin/commit/d958a62 -->
+            <generatedSourcesDirectory>${project.build.directory}/generated-sources</generatedSourcesDirectory>
+            <!-- no annotation processing -->
+            <proc>none</proc>
           </configuration>
         </plugin>
 


szetszwo (approved):
+1 the change looks good.

OneSizeFitsQuorum:
Will do test soon

OneSizeFitsQuorum (approved):
Verified locally following the release build procedure, with an explanation of the likely root cause.

**Environment**
- OS: macOS (Darwin 25.3.0, arm64)
- JDK: OpenJDK 1.8.0_472 (Zulu 8.90)
- Maven: 3.9.16
- maven-compiler-plugin: 3.15.0
- Build: `./mvnw -am -pl :ratis-proto -DskipTests clean install -Prelease -Papache-release -Dgpg.skip=true -Dmaven.compiler.release=8`

**Likely root cause**
`maven-compiler-plugin` enables annotation processing by default and directs any AP-generated sources to `target/generated-sources/annotations`, creating that directory even when no annotation processor is configured or run (a known plugin quirk, fixed upstream in maven-compiler-plugin 4.0 by commit d958a62). In `ratis-proto`, the protobuf plugin already emits generated sources directly into `target/generated-sources` (so `org/...` lives there), making it a real source root. During the release build, `maven-source-plugin` (activated by the `release` profile) archives the `generated-sources` source root into the `-sources.jar`, and the empty `annotations/` sibling directory is archived along with it — ending up as an empty `annotations/` entry in the published sources jar. The main jar is unaffected because it is built from `target/classes`, where the empty directory is never materialized.

**Baseline (current master, without the patch):** reproduced —
```
$ unzip -l ratis-proto-3.3.0-SNAPSHOT-sources.jar | grep annotations
        0  ...   annotations/
```
`generated-sources/` contains `annotations/` + `org/`.

**With this patch (pr-1493):** same build → `generated-sources/` contains only `org/`; the sources jar has no `annotations/` entry. `<proc>none</proc>` disables annotation processing so the `annotations/` output directory is never created, and `<generatedSourcesDirectory>` pins the location to the proto-generated path.

**No regression:** `./mvnw -pl :ratis-server,:ratis-client -am -DskipTests clean verify` — all downstream modules compile successfully with the global `<proc>none</proc>`.

+1.

adoroszlai:
Thanks a lot @OneSizeFitsQuorum for testing it.

> The main jar is unaffected because it is built from `target/classes`, where the empty directory is never materialized.

It is affected, but only in the 3.2.2 release artifact, we cannot reproduce it.

```bash
$ curl -LOSs https://repo1.maven.org/maven2/org/apache/ratis/ratis-proto/3.2.2/ratis-proto-3.2.2.jar
$ unzip -t ratis-proto-3.2.2.jar | grep ': annotations'                                       
    testing: annotations/             OK
```

adoroszlai:
Thanks @OneSizeFitsQuorum, @szetszwo for the review.

PR #1494, opened 2026-06-26
RATIS-2573. Use Java 25 in CI

adoroszlai:

Use Java 25 for building and testing Ratis in CI, still targeting Java 8.

- Add options to allow using `sun.misc.Unsafe` and loading native libraries.  (These will also need to be allowed in Ratis scripts, separately.)
- Upgrade `byte-buddy` for Java 25 compatibility (used by Mockito, which we cannot upgrade due to dropping Java 8 support).
- Drop Java 11 from compile check matrix to keep resource usage the same.

https://issues.apache.org/jira/browse/RATIS-2573

## How was this patch tested?

CI:
https://github.com/adoroszlai/ratis/actions/runs/28223570864

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -194,6 +194,7 @@
     <slf4j.version>2.0.18</slf4j.version>
     <junit-bom.version>5.14.4</junit-bom.version>
     <mockito.version>4.11.0</mockito.version>
+    <byte-buddy.version>1.18.10</byte-buddy.version>
     <jacoco.version>0.8.15</jacoco.version>
     <flaky-test-groups>flaky | org.apache.ratis.test.tag.FlakyTest</flaky-test-groups>
 
@@ -407,6 +408,16 @@
         <groupId>org.mockito</groupId>
         <artifactId>mockito-core</artifactId>
         <version>${mockito.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>net.bytebuddy</groupId>
+        <artifactId>byte-buddy</artifactId>
+        <version>${byte-buddy.version}</version>
+      </dependency>
+      <dependency>
+      <groupId>net.bytebuddy</groupId>
+        <artifactId>byte-buddy-agent</artifactId>
+        <version>${byte-buddy.version}</version>
       </dependency>
     </dependencies>
   </dependencyManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -196,6 +196,12 @@
     <mockito.version>4.11.0</mockito.version>
     <jacoco.version>0.8.15</jacoco.version>
     <flaky-test-groups>flaky | org.apache.ratis.test.tag.FlakyTest</flaky-test-groups>
+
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -402,6 +409,16 @@
...
+      <dependency>
+        <groupId>net.bytebuddy</groupId>
+        <artifactId>byte-buddy</artifactId>
+        <version>${byte-buddy.version}</version>
+      </dependency>
+      <dependency>
+      <groupId>net.bytebuddy</groupId>
+        <artifactId>byte-buddy-agent</artifactId>
+        <version>${byte-buddy.version}</version>
+      </dependency>
```
szetszwo, line 421: Just a question: Which part is using byte-buddy?

adoroszlai: Mockito uses it to manipulate classes for mocking.

https://central.sonatype.com/artifact/org.mockito/mockito-core/4.11.0/dependencies

szetszwo: Then, why need to add the byte-buddy dependencies?   For setting the versions?  It will bring them automatically.

- Without byte-buddy dependencies
```
[INFO] \- org.mockito:mockito-core:jar:4.11.0:test
[INFO]    +- net.bytebuddy:byte-buddy:jar:1.12.19:test
[INFO]    +- net.bytebuddy:byte-buddy-agent:jar:1.12.19:test
[INFO]    \- org.objenesis:objenesis:jar:3.3:test
```

- With byte-buddy dependencies
```
[INFO] \- org.mockito:mockito-core:jar:4.11.0:test
[INFO]    +- net.bytebuddy:byte-buddy:jar:1.18.10:test
[INFO]    +- net.bytebuddy:byte-buddy-agent:jar:1.18.10:test
[INFO]    \- org.objenesis:objenesis:jar:3.3:test
```

adoroszlai: Without upgrading this transitive dependency, we get errors like:

```
Mockito cannot mock this class: class org.apache.ratis.server.impl.RaftServerImpl
Mockito cannot mock this class: class org.apache.ratis.grpc.server.GrpcLogAppender$RequestMap
Mockito cannot mock this class: interface org.apache.ratis.statemachine.StateMachine
Mockito cannot mock this class: class java.io.File
```

https://github.com/adoroszlai/ratis/actions/runs/28196396158

We cannot upgrade Mockito itself, because Mockito 5 dropped support for Java 8.

https://github.com/mockito/mockito/releases/tag/v5.0.0

szetszwo: I see the problem now -- Byte Buddy 1.12 does not work with Java 25.
- Caused by: java.lang.IllegalArgumentException: Java 25 (69) is not supported by the current version of Byte Buddy which officially supports Java 20 (64) - update Byte Buddy or set net.bytebuddy.experimental as a VM property

szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #1499, opened 2026-06-30
RATIS-2581. Route file operations through NIO.2

HTHou:

Route a few remaining Ratis file operations through Java NIO.2 so they can be intercepted by the default `FileSystemProvider`.

This patch removes direct use of `File.delete()` from `AtomicFileOutputStream#abort`, opens segmented log write channels through `FileChannel.open(Path, ...)` instead of `RandomAccessFile#getChannel`, and routes zero-size open log segment deletion through `FileUtils`.

With these changes, downstream projects that install a custom default `FileSystemProvider` can cover these cleanup and truncate paths without adding a Ratis-specific file destruction API.

https://issues.apache.org/jira/browse/RATIS-2581

## How was this patch tested?

- Ran `git diff --check` and `git diff --cached --check`.
- Ran `./mvnw -pl ratis-common -Dtest=TestFileUtils test`.
- Ran `./mvnw -pl ratis-server -am -DskipTests compile`.


szetszwo (approved):
+1 the change looks good.

PR #1504, opened 2026-07-02
RATIS-2586. zizmor check should reflect failure in fork

adoroszlai:

Same as https://github.com/apache/ratis-thirdparty/pull/138, for `apache/ratis`.

zizmor workflow added in RATIS-2493 provides feedback via "Code scanning results".  The workflow itself passes even if it finds issues to be fixed.  This does not work well in forks where code scanning is disabled.

To provide better feedback, the workflow should reflect scan results in forks.

Also:
- skip zizmor check for changes that do not touch workflows
- cancel run for obsolete commit (if new commit is pushed to the branch, but only in forks/PR)

https://issues.apache.org/jira/browse/RATIS-2586

## How was this patch tested?

Tested the same in `apache/ozone`:

Without the patch passes despite intentional violation:
https://github.com/adoroszlai/ozone/actions/runs/28570405264/job/84706670204#step:3:1058

Same violation with the patch fails:
https://github.com/adoroszlai/ozone/actions/runs/28570585965/job/84707233189#step:3:208

The patch without intentional violation passes (also in this repo):
https://github.com/adoroszlai/ozone/actions/runs/28570809169/job/84707932173#step:3:107
https://github.com/adoroszlai/ratis/actions/runs/28617573739


szetszwo (approved):
+1 the change looks good.

adoroszlai:

PR #1510, opened 2026-07-10
RATIS-2596. Unsupported TLS cipher suites may crash gRPC servers

fapifta:
In case TLS ciphers are configurable, it is possible to configure ciphers that are not supported by Netty.
This patch changes the IdentityCipherSuiteFilter to the SupportedCipherSuiteFilter in GrpcUtil.

https://issues.apache.org/jira/browse/RATIS-2596

## How was this patch tested?
What is changing is tested functionality, the issue has appeared already elsewhere, I did not perform any particular tests as this exact changed worked well in the previous occurance (HDDS-15176).


szetszwo (approved):
+1 the change looks good

PR #1515, opened 2026-07-12
RATIS-2601. GroupInfoReply conf is lost when converting to GroupInfoReplyProto

ryu1-sakai:

`ClientProtoUtils#toGroupInfoReplyProto()` doesn't set the field `conf` (`RaftConfigurationProto`) when it converts `GroupInfoReply` into `GroupInfoReplyProto`. Consequently, the converted proto lacks `conf` and `GroupInfoReply.getConf()` always returns an empty `Optional` on the client side.

This patch sets `conf` when it is present.

https://issues.apache.org/jira/browse/RATIS-2601

## How was this patch tested?

Added a round-trip test in `TestClientProtoUtils` (`GroupInfoReply` -> `RaftConfigurationProto` -> `GroupInfoReply`). It covers both null and non-null `conf` by `@ParemeterizedTest`.


szetszwo (approved):
+1 the change looks good.

PR #1516, opened 2026-07-13
RATIS-2603. Add DataStreamApi.Resolver for group-independent read-only requests

peterxcli:

RATIS-2603
Users such as Ozone may archive part of their data and mark it immutable. In Ozone, closed containers use a synthetic read pipeline whose ID does not correspond to a live Ratis group, causing DataStream reads to fail with GroupMismatchException. 

Introducing a new DataStreamReadResolver, which should provide an optional path to bypass group lookup and let the application define how such reads are handled. If the resolver declines the request, the existing Ratis read path remains unchanged.	

### How could Users Use this?

for example, in ozone, we register our resolver to DataStreamReadResolver in `XceiverServerRatis#newXceiverServerRatis` when `OzoneContainer` starts.
```java
public static XceiverServerRatis newXceiverServerRatis(...){
  // ...
  ClosedContainerReadResolver resolver = new ClosedContainerReadResolver(dispatcher, 
    containerController, datanodeDetails);
  DataStreamReadResolver.set(parameters, resolver);
  // ...
}
```

and the resolver implmentaion should be like:
```java
/** Resolves group-independent reads against immutable local containers. */
final class ClosedContainerReadResolver implements DataStreamReadResolver {
  private final ContainerDispatcher dispatcher;
  private final ContainerController containerController;
  private final String datanodeUuid;

  ClosedContainerReadResolver(ContainerDispatcher dispatcher,
      ContainerController containerController, DatanodeDetails datanode) {
    this.dispatcher = dispatcher;
    this.containerController = containerController;
    this.datanodeUuid = datanode.getUuidString();
  }

  @Override
  public StateMachine.DataApi resolve(RaftClientRequest request)
      throws IOException {
    final ContainerCommandRequestProto requestProto = ContainerCommandRequestMessage.toProto(
        request.getMessage().getContent(), request.getRaftGroupId());
    if (requestProto.getCmdType() != Type.ReadBlock
        || requestProto.hasDatanodeUuid()
            && !datanodeUuid.equals(requestProto.getDatanodeUuid())) {
      return null;
    }

    final Container<?> container =
        containerController.getContainer(requestProto.getContainerID());
    if (container == null || container.getContainerState() != CLOSED) {
      return null;
    }

    return new StateMachine.DataApi() {
      @Override
      public void query(Message ignored, WritableByteChannel stream) {
        try {
          ContainerStateMachine.streamReadBlock(
              dispatcher, requestProto, stream);
        } catch (IOException e) {
          throw new CompletionException(e);
        }
      }
    };
  }
}
```

https://issues.apache.org/jira/browse/RATIS-2603

## How was this patch tested?


peterxcli:
https://github.com/apache/ozone/pull/10743 is the patch that use the resolver.

cc @szetszwo @amaliujia, please could you take a look at them whenever you have time, thanks!

ratis-server-api/src/main/java/org/apache/ratis/server/DataStreamReadResolver.java
```diff
@@ -0,0 +1,47 @@
...
+public interface DataStreamReadResolver {
+  String PARAMETER_KEY = DataStreamReadResolver.class.getName();
+
+  /** Set the resolver in the given server parameters. */
+  static void set(Parameters parameters, DataStreamReadResolver resolver) {
+    parameters.put(PARAMETER_KEY, resolver, DataStreamReadResolver.class);
+  }
+
+  /** Get the resolver from the given server parameters, if configured. */
+  static DataStreamReadResolver get(Parameters parameters) {
```
amaliujia: Nit: you can add `@Nullable` notation to `Parameters`

ratis-netty/src/main/java/org/apache/ratis/netty/server/ReadStreamManagement.java
```diff
@@ -182,6 +199,22 @@ private boolean processImpl(DataStreamRequestByteBuf requestBuf, ChannelHandlerC
       return false;
     }
 
+    if (readResolver != null) {
+      final StateMachine.DataApi dataApi;
+      try {
+        dataApi = readResolver.resolve(request);
+      } catch (Throwable t) {
+        replyDataStreamException(server, t, request, requestBuf, ctx);
```
amaliujia, line 206: Why we do not choose to fall back to division when see exceptions from the resolver>

peterxcli: for example, the container is closed, it doesn't have the division anymore, if the resolve process thrown exception, then the fallback  to division would just get the the `GroupMismatchException`

ratis-server-api/src/main/java/org/apache/ratis/server/DataStreamReadResolver.java
```diff
@@ -0,0 +1,47 @@
...
+    return parameters != null
+        ? parameters.get(PARAMETER_KEY, DataStreamReadResolver.class)
+        : null;
+  }
```
szetszwo: Let'e move the conf key and methods to RaftServerConfigKeys.DataStream:
```java
@@ -17,7 +17,9 @@
  */
 package org.apache.ratis.server;
 
+import org.apache.ratis.conf.Parameters;
 import org.apache.ratis.conf.RaftProperties;
+import org.apache.ratis.server.api.DataStreamServerApi;
 import org.apache.ratis.util.TimeDuration;
@@ -847,6 +849,15 @@ public interface RaftServerConfigKeys {
     static void setClientPoolSize(RaftProperties properties, int num) {
       setInt(properties::setInt, CLIENT_POOL_SIZE_KEY, num);
     }
+
+    String SERVER_API_RESOLVER_PARAMETER = PREFIX + "-server-api-resolver";
+    Class<DataStreamServerApi.Resolver> SERVER_API_RESOLVER_CLASS = DataStreamServerApi.Resolver.class;
+    static DataStreamServerApi.Resolver serverApiResolver(Parameters parameters) {
+      return parameters != null ? parameters.get(SERVER_API_RESOLVER_PARAMETER, SERVER_API_RESOLVER_CLASS): null;
+    }
+    static void setServerApiResolver(Parameters parameters, DataStreamServerApi.Resolver resolver) {
+      parameters.put(SERVER_API_RESOLVER_PARAMETER, resolver, SERVER_API_RESOLVER_CLASS);
+    }
   }
```

szetszwo (commented):
@peterxcli , thanks a lot for working on this.  It is a really great idea!

Let's do a refactoring so that the new API is not in StateMachine; filed RATIS-2604.  We could also make it similar to the FileChannel.transferTo(..) method.  I have tried it a little bit; see https://issues.apache.org/jira/secure/attachment/13083178/1516_review.patch

```java
/** Server side data stream API. */
public interface DataStreamApi { // TODO: add this RATIS-2604
  /**
   * Similar to {@link java.nio.channels.FileChannel#transferTo(long, long, WritableByteChannel)}
   * except that the parameter of this method is request message
   * but not position, count in FileChannel.
   *
   * @param request the request message
   * @param stream the output stream to send the results
   * @return the number of bytes transferred
   */
  long transferTo(Message request, WritableByteChannel stream) throws IOException;

  /** For resolving {@link DataStreamApi}. */
  @FunctionalInterface
  interface Resolver { // TODO: add this in RATIS-2603
    /** @return the data API handling this request, or null if the API is not found. */
    DataStreamApi resolve(RaftClientRequest request);
  }
}
```

szetszwo (approved):
+1 the change looks good.

amaliujia (approved):

peterxcli:
Thanks @szetszwo for the review, refactoring suggestion and renaming title! @amaliujia for the review!

PR #1517, opened 2026-07-14
RATIS-2602. Pass CipherSuiteFilter to the TlsConf.

HTHou:

Make `CipherSuiteFilter` configurable through `TlsConf`. The default remains
`SupportedCipherSuiteFilter` to preserve the current safe behavior, while callers can select
`IdentityCipherSuiteFilter` for provider-specific cipher suites that are accepted by the TLS
provider but absent from Netty's supported set.

Pass the configured filter to both the gRPC and Netty DataStream SSL context builders, and preserve
the fluent return type in `GrpcTlsConfig.Builder`.

https://issues.apache.org/jira/browse/RATIS-2602

## How was this patch tested?

- Ran `TestGrpcTlsConfig` and `TestTlsConfWithNetty` on Java 8: 11 tests passed.
- Ran `TestGrpcTlsConfig` and `TestTlsConfWithNetty` on Java 17: 11 tests passed.
- Ran Checkstyle for the affected reactor: 0 violations.
- Ran `git diff --check origin/master...HEAD`.

### DIFF
--- a/ratis-common/src/main/java/org/apache/ratis/security/TlsConf.java
+++ b/ratis-common/src/main/java/org/apache/ratis/security/TlsConf.java
@@ -17,7 +17,9 @@
  */
 package org.apache.ratis.security;
 
+import org.apache.ratis.thirdparty.io.netty.handler.ssl.CipherSuiteFilter;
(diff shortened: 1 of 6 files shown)


szetszwo (approved):
+1 the change looks good.

PR #1519, opened 2026-07-14
RATIS-2605. Avoid advancing matchIndex from heartbeat AppendEntries success

FMX:

  This patch fixes leader-side AppendEntries success handling for heartbeat requests.

  A follower returns its local `nextIndex` for successful heartbeat AppendEntries replies. That value describes the follower's local log tail, but it does not prove that all entries
  up to `nextIndex - 1` match the leader's log. If the follower has a divergent uncommitted tail from an old leader, trusting the heartbeat reply `nextIndex` may incorrectly advance
  the leader-side `matchIndex`.

  Once `matchIndex` is polluted, later `INCONSISTENCY` handling can use `matchIndex + 1` as a lower bound and prevent `nextIndex` from backing off to the real common prefix.

  This patch preserves the AppendEntries request metadata together with the reply. For successful heartbeat replies, the leader advances only to `request.previousLog.index + 1`. For
  successful append replies, the existing behavior is preserved and `reply.nextIndex` is still used.

  Unit tests are added for:

  1. Heartbeat `SUCCESS` with `previousLog` advances only to `previousLog.index + 1`.
  2. Heartbeat `SUCCESS` without `previousLog` does not advance `matchIndex`.
  3. Append `SUCCESS` still advances using `reply.nextIndex`.

  https://issues.apache.org/jira/browse/RATIS-2605

  ## How was this patch tested?

  ```bash
  ./mvnw -pl ratis-server -am -Dtest=TestLogAppenderDefault test

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/leader/LogAppenderDefault.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/leader/LogAppenderDefault.java
@@ -55,8 +55,24 @@ class LogAppenderDefault extends LogAppenderBase {
     return CallId.getComparator();
   }
 
+  private static final class AppendEntriesResult {
+    private final AppendEntriesReplyProto reply;
+    private final long requestFirstIndex;
+    private final long requestPreviousIndex;
+    private final boolean requestHeartbeat;
+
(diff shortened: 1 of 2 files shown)


szetszwo (approved):
@FMX , thanks a lot for reporting and fixing the bug!

+1 the change looks good.

PR #1523, opened 2026-07-20
RATIS-2610. -groupid should check for non-empty ID

spacemonkd:

RATIS-2610. -groupid should check for non-empty ID

In [this](https://github.com/apache/ratis/blob/f4c3781468e756220836e88d953fd5598177afdf/ratis-shell/src/main/java/org/apache/ratis/shell/cli/CliUtils.java#L99) line a valid non-empty UUID returns `null`.
The check should be `!groupId.isEmpty()` .

https://issues.apache.org/jira/browse/RATIS-2610

## How was this patch tested?
Patch was tested via existing unit tests and newly added tests


spacemonkd:
@szetszwo could you take a look at this small change?

szetszwo (approved):
+1 the change looks good.

PR #1524, opened 2026-07-20
RATIS-2611. Clear pending queue on all install-snapshot replies

spacemonkd:

RATIS-2611. Clear pending queue on all install-snapshot replies

When calling the onNext method, in the various result cases [here](https://github.com/apache/ratis/blob/f4c3781468e756220836e88d953fd5598177afdf/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java#L676) some cases do not call `removePending()`.

NOT_LEADER
CONF_MISMATCH
UNRECOGNIZED
SNAPSHOT_EXPIRED
these cases do not clear the pending queue and this may possibly fail on the Preconditions.assertSame() check.

`pending.poll()` removes the oldest enqueued request index. The handler assumes that for each reply we will have one addPending() and then one removePending() call in order.

So if we take an example of chunk install with `SNAPSHOT_EXPIRED`

1. Let us assume we send chunk [0,1,2] -> three `addPending()` calls -> pending queue [0, 1, 2]
2. Reply with chunk 0 as success -> one `removePending()` call -> pending.poll() returns 0, condition passed
3. Reply with chunk 1 as SNAPSHOT_EXPIRED -> pending queue still has [1, 2]

This edge case is handled in the PR.

https://issues.apache.org/jira/browse/RATIS-2611

## How was this patch tested?
Unit tests

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java
@@ -695,11 +695,13 @@ public class GrpcLogAppender extends LogAppenderBase {
           break;
         case NOT_LEADER:
           onFollowerTerm(reply.getTerm());
+          pending.clear();
           break;
         case CONF_MISMATCH:
           LOG.error("{}: CONF_MISMATCH ({}): Leader {} has it set to {} but follower {} has it set to {}",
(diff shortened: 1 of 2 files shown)


spacemonkd:
@szetszwo could you take a look at this PR?

P.S. The tests have been generated via Cursor and have been validated locally.

PR #1531, opened 2026-07-25
RATIS-2621. Regenerate test certs with alt_names.

szetszwo:
See RATIS-2621

Updating to thirdparty 1.1.0 works after this: https://github.com/szetszwo/ratis/actions/runs/30171943304


adoroszlai (approved):

szetszwo:
@adoroszlai , thanks for reviewing and merging this!

PR #1532, opened 2026-07-26
RATIS-2616. Bump ratis-thirdparty-misc to 1.1.0.

szetszwo:
See RATIS-2616.

Also update dependabot to ignore ratis-thirdparty-misc since it is not able to update also the gRPC and protobuf versions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -177,11 +177,11 @@
     <maven.compiler.release>${javaVersion}</maven.compiler.release>
 
     <!-- Contains all shaded thirdparty dependencies; make sure to also update in ratis-bom/pom.xml -->
-    <ratis.thirdparty.version>1.0.11</ratis.thirdparty.version>
+    <ratis.thirdparty.version>1.1.0</ratis.thirdparty.version>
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -181,7 +181,7 @@
 
     <!-- Need these for the protobuf compiler. *MUST* match what is in ratis-thirdparty -->
     <shaded.protobuf.version>3.25.9</shaded.protobuf.version>
-    <shaded.grpc.version>1.82.1</shaded.grpc.version>
+    <shaded.grpc.version>1.82.2</shaded.grpc.version>
```
spacemonkd, line 184: This is still `1.89.1` in `ratis-thirdparty-misc`.
https://github.com/apache/ratis-thirdparty/blob/release-1.1.0/pom.xml#L69

spacemonkd:
Thanks for the change @szetszwo!
We can add the dependabot exception, but the minor version bump is not needed as it is still `1.82.1`

adoroszlai:
`ratis.thirdparty.version` bumped in #1529, dependabot ignore added in #1533, `shaded.grpc.version` should not be changed.

PR #1533, opened 2026-07-26
RATIS-2622. Set dependabot to ignore ratis-thirdparty.

szetszwo:
See RATIS-2622

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -181,7 +181,7 @@
 
     <!-- Need these for the protobuf compiler. *MUST* match what is in ratis-thirdparty -->
     <shaded.protobuf.version>3.25.9</shaded.protobuf.version>
-    <shaded.grpc.version>1.82.2</shaded.grpc.version>
+    <shaded.grpc.version>1.82.1</shaded.grpc.version>
 
     <!-- OpenTelemetry versions -->
     <opentelemetry.version>1.64.0</opentelemetry.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -181,7 +181,7 @@
 
     <!-- Need these for the protobuf compiler. *MUST* match what is in ratis-thirdparty -->
     <shaded.protobuf.version>3.25.9</shaded.protobuf.version>
-    <shaded.grpc.version>1.82.1</shaded.grpc.version>
+    <shaded.grpc.version>1.82.2</shaded.grpc.version>
```
adoroszlai: `ratis-thirdparty-misc-1.1.0` is still on 1.82.1, see https://github.com/apache/ratis-thirdparty/blob/1.1.0/pom.xml#L66-L69 (as mentioned in https://github.com/apache/ratis/pull/1529#discussion_r3651874823)

adoroszlai (commented):
Thanks @szetszwo for the patch.  I'm not sure this is needed, since we can push additional commits to dependabot's PRs like https://github.com/apache/ratis/pull/1529/commits/ec72e3231e0016b7842e6bb5349b37958a1ed1c7.

szetszwo:
> ... since we can push additional commits to dependabot's PRs ...

It is easy to forget and new committers may not aware of it.  Let's do it.

adoroszlai (approved):

PR #1534, opened 2026-07-29
RATIS-2625. Streaming: support sending commands in Data Stream

amaliujia:
**This PR is AI-assisted**

Adds a first-class mid-stream command plane to Ratis data streams. Callers can send opaque command payloads during an active stream via DataStreamOutput.commandAsync(). The server delivers them to StateMachine.DataStream.onCommand() instead of writing bytes to the data channel. Commands are ordered with data writes, replicated to remote peers like STREAM_DATA, and do not advance the stream byte offset.

Example usage:
```
// Client
out.getHeaderFuture().join();
out.writeAsync(dataBuffer).join();
out.commandAsync(myCommandBytes).join();  // opaque payload
out.writeAsync(moreData, StandardWriteOption.CLOSE).join();

// State machine
@Override
public CompletableFuture<?> onCommand(ByteBuffer command, long streamOffset) {
  // parse command, e.g. force(), metadata update, etc.
}
```

https://issues.apache.org/jira/browse/RATIS-2625

## How was this patch tested?
Unit tests

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
@@ -204,10 +204,6 @@ public class DataStreamClientImpl implements DataStreamClient {
       return commandAsyncImpl(src, src.remaining());
     }
 
-    public CompletableFuture<DataStreamReply> commandAsync(ByteBuf src) {
-      return commandAsyncImpl(src, src.readableBytes());
-    }
-
     boolean isClosed() {
       return closeFuture != null;
     }
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
@@ -181,7 +181,8 @@ public class DataStreamClientImpl implements DataStreamClient {
         return JavaUtils.completeExceptionally(new AlreadyClosedException(
             clientId + ": stream already closed, request=" + header));
       }
-      return combineHeader(send(Type.STREAM_CONTROL, data, length, Collections.emptyList()));
+      return combineHeader(send(Type.STREAM_CONTROL, data, length,
+          Collections.singleton(StandardWriteOption.FLUSH)));
     }
 
     public CompletableFuture<DataStreamReply> writeAsync(ByteBuf src, Iterable<WriteOption> options) {
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
@@ -173,18 +173,6 @@ public class DataStreamClientImpl implements DataStreamClient {
       return f;
     }
 
-    private CompletableFuture<DataStreamReply> commandAsyncImpl(Object data, long length) {
-      if (isClosed()) {
-        if (data instanceof ByteBuf) {
-          ((ByteBuf) data).release();
-        }
-        return JavaUtils.completeExceptionally(new AlreadyClosedException(
-            clientId + ": stream already closed, request=" + header));
-      }
-      return combineHeader(send(Type.STREAM_COMMAND, data, length,
-          Collections.singleton(StandardWriteOption.FLUSH)));
-    }
-
     public CompletableFuture<DataStreamReply> writeAsync(ByteBuf src, Iterable<WriteOption> options) {
       return writeAsyncImpl(src, src.readableBytes(), options);
     }
@@ -201,7 +189,12 @@ public class DataStreamClientImpl implements DataStreamClient {
 
     @Override
     public CompletableFuture<DataStreamReply> commandAsync(ByteBuffer src) {
-      return commandAsyncImpl(src, src.remaining());
+      if (isClosed()) {
+        return JavaUtils.completeExceptionally(new AlreadyClosedException(
+            clientId + ": stream already closed, request=" + header));
+      }
+      return combineHeader(send(Type.STREAM_COMMAND, src, src.remaining(),
+          Collections.singleton(StandardWriteOption.FLUSH)));
     }
 
     boolean isClosed() {
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
@@ -173,7 +173,7 @@ public class DataStreamClientImpl implements DataStreamClient {
       return f;
     }
 
-    private CompletableFuture<DataStreamReply> controlAsyncImpl(Object data, long length) {
+    private CompletableFuture<DataStreamReply> commandAsyncImpl(Object data, long length) {
       if (isClosed()) {
         if (data instanceof ByteBuf) {
           ((ByteBuf) data).release();
@@ -181,7 +181,7 @@ public class DataStreamClientImpl implements DataStreamClient {
         return JavaUtils.completeExceptionally(new AlreadyClosedException(
             clientId + ": stream already closed, request=" + header));
       }
-      return combineHeader(send(Type.STREAM_CONTROL, data, length,
+      return combineHeader(send(Type.STREAM_COMMAND, data, length,
           Collections.singleton(StandardWriteOption.FLUSH)));
     }
 
@@ -200,12 +200,12 @@ public class DataStreamClientImpl implements DataStreamClient {
     }
 
     @Override
-    public CompletableFuture<DataStreamReply> controlAsync(ByteBuffer src) {
-      return controlAsyncImpl(src, src.remaining());
+    public CompletableFuture<DataStreamReply> commandAsync(ByteBuffer src) {
+      return commandAsyncImpl(src, src.remaining());
     }
 
-    public CompletableFuture<DataStreamReply> controlAsync(ByteBuf src) {
-      return controlAsyncImpl(src, src.readableBytes());
+    public CompletableFuture<DataStreamReply> commandAsync(ByteBuf src) {
+      return commandAsyncImpl(src, src.readableBytes());
     }
 
     boolean isClosed() {
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
@@ -173,6 +173,18 @@ public class DataStreamClientImpl implements DataStreamClient {
       return f;
     }
 
+    private CompletableFuture<DataStreamReply> controlAsyncImpl(Object data, long length,
+        Iterable<WriteOption> options) {
+      if (isClosed()) {
+        if (data instanceof ByteBuf) {
+          ((ByteBuf) data).release();
+        }
+        return JavaUtils.completeExceptionally(new AlreadyClosedException(
+            clientId + ": stream already closed, request=" + header));
+      }
+      return combineHeader(send(Type.STREAM_CONTROL, data, length, options));
+    }
+
     public CompletableFuture<DataStreamReply> writeAsync(ByteBuf src, Iterable<WriteOption> options) {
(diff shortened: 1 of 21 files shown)


amaliujia:
@szetszwo this is the follow up for https://github.com/apache/ozone/pull/10823#discussion_r3676243122

ratis-server-api/src/main/java/org/apache/ratis/statemachine/StateMachine.java
```diff
@@ -385,6 +385,19 @@ interface DataStream {
...
+    /**
+     * Handle a command received in the middle of a data stream.
+     * The {@code streamOffset} indicates the current byte offset in the stream
+     * (i.e. the total number of data bytes received so far).
+     *
+     * @param command the command payload
+     * @param streamOffset the current stream byte offset
+     * @return a future for the command task
+     */
+    default CompletableFuture<?> onCommand(ByteBuffer command, long streamOffset) {
```
szetszwo: Since this is a commnad, it should return CompletableFuture\<ByteBuffer>.

szetszwo (commented):
@amaliujia , thanks for working on this!  The change looks quite good.

The StateMachine may return a result, we should send it back to the client; see https://issues.apache.org/jira/secure/attachment/13083627/1534_review2.patch

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
```diff
@@ -187,6 +199,11 @@ public CompletableFuture<DataStreamReply> writeAsync(FilePositionCount src, Writ
       return writeAsyncImpl(src, src.getCount(), Arrays.asList(options));
     }
 
+    @Override
+    public CompletableFuture<DataStreamReply> commandAsync(ByteBuffer src) {
+      return commandAsyncImpl(src, src.remaining());
```
szetszwo: commandAsyncImpl is used just once.  Let's inline the code.
```java
    public CompletableFuture<DataStreamReply> commandAsync(ByteBuffer src) {
      if (isClosed()) {
        return JavaUtils.completeExceptionally(new AlreadyClosedException(
            clientId + ": stream already closed, request=" + header));
      }
      return combineHeader(send(Type.STREAM_COMMAND, src, src.remaining(),
          Collections.singleton(StandardWriteOption.FLUSH)));
    }
```

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -86,23 +87,65 @@
...
+    ByteBuffer getCommandReply() {
+      return commandReply;
+    }
+  }
+
   static class LocalStream {
     private final CompletableFuture<DataStream> streamFuture;
-    private final AtomicReference<CompletableFuture<Long>> writeFuture;
-    private final RequestMetrics metrics;
+    private final AtomicReference<CompletableFuture<LocalResult>> writeFuture;
```
szetszwo: Let's rename it to `resultFuture`.

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -86,23 +87,65 @@
...
-      this.writeFuture = new AtomicReference<>(streamFuture.thenApply(s -> 0L));
-      this.metrics = metrics;
+      this.writeFuture = new AtomicReference<>(streamFuture.thenApply(s -> LocalResult.ZERO));
+      this.writeMetrics = writeMetrics;
+      this.commandMetrics = commandMetrics;
     }
 
-    CompletableFuture<Long> write(ByteBuf buf, Iterable<WriteOption> options,
+    CompletableFuture<LocalResult> write(ByteBuf buf, Iterable<WriteOption> options,
                                   Executor executor) {
```
szetszwo: Make it a single line.

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -469,24 +544,31 @@ private void readImpl(DataStreamRequestByteBuf request, ChannelHandlerContext ct
...
     } else if (request.getType() == Type.STREAM_DATA) {
-      localWrite = info.getLocal().write(request.slice(), request.getWriteOptionList(), writeExecutor);
+      localResult = info.getLocal().write(request.slice(), request.getWriteOptionList(), writeExecutor);
       remoteWrites = info.applyToRemotes(out -> out.write(request, requestExecutor));
+    } else if (request.getType() == Type.STREAM_COMMAND) {
+      // command is supposed to have small data size, just copy it
+      final ByteBuffer command = copyBuffer(request.slice());
+      localResult = info.getLocal().command(command, request.getStreamOffset(), writeExecutor);
+      remoteWrites = info.applyToRemotes(out -> out.command(
+          copyBuffer(request.slice()), requestExecutor));
```
szetszwo: duplicate() and reuse command, i.e.
```java
      final ByteBuffer command = copyBuffer(request.slice());
      localResult = info.getLocal().command(command.duplicate(), request.getStreamOffset(), writeExecutor);
      remoteWrites = info.applyToRemotes(out -> out.command(command, requestExecutor));
```

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -303,6 +359,22 @@ static CompletableFuture<Long> writeToAsync(ByteBuf buf,
...
+        .thenCompose(ignored -> stream.onCommand(command, streamOffset));
+  }
+
+  static ByteBuffer copyBuffer(ByteBuf buf) {
+    final ByteBuffer copy = ByteBuffer.allocate(buf.readableBytes());
+    for (ByteBuffer buffer : buf.nioBuffers()) {
+      copy.put(buffer);
+    }
+    copy.flip();
+    return copy;
```
szetszwo: Make it readonly
```java
    return copy.asReadOnlyBuffer();
```

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -303,6 +359,22 @@ static CompletableFuture<Long> writeToAsync(ByteBuf buf,
     return CompletableFuture.supplyAsync(() -> writeTo(buf, options, stream), e);
   }
 
+  static CompletableFuture<ByteBuffer> commandToAsync(ByteBuffer command, long streamOffset,
+      DataStream stream, Executor defaultExecutor) {
+    final Executor e = Optional.ofNullable(stream.getExecutor()).orElse(defaultExecutor);
+    return CompletableFuture.runAsync(() -> {}, e)
+        .thenCompose(ignored -> stream.onCommand(command, streamOffset));
```
szetszwo: Using completedFuture(null) is more efficient since it doesn't have to run an empty task.
```java
    return CompletableFuture.completedFuture(null)
        .thenComposeAsync(ignored -> stream.onCommand(command, streamOffset), e);
```

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
```diff
@@ -519,30 +601,48 @@ static void assertReplyCorrespondingToRequest(
...
-      final DataStreamRequestByteBuf request) {
+  private boolean checkSuccessRemoteWrite(List<CompletableFuture<DataStreamReply>> replyFutures,
+      LocalResult localResult, final DataStreamRequestByteBuf request) {
     for (CompletableFuture<DataStreamReply> replyFuture : replyFutures) {
       final DataStreamReply reply;
       try {
         reply = replyFuture.get(requestTimeout.getDuration(), requestTimeout.getUnit());
       } catch (Exception e) {
-        throw new CompletionException("Failed to get reply for bytesWritten=" + bytesWritten + ", " + request, e);
+        throw new CompletionException("Failed to get reply for " + localResult + ", " + request, e);
```
szetszwo, line 614: Add toString()
```java
    @Override
    public String toString() {
      return commandReply != null ? "commandReply:" + StringUtils.bytes2HexString(commandReply)
          : "byteWritten:" + byteWritten;
    }
```

szetszwo (commented):
@amaliujia , thanks for the update!  Please see the comments inlined.

ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
amaliujia: Thanks. I really learned something from this comment.

ratis-client/src/main/java/org/apache/ratis/client/impl/DataStreamClientImpl.java
```diff
@@ -187,6 +187,16 @@ public CompletableFuture<DataStreamReply> writeAsync(FilePositionCount src, Writ
       return writeAsyncImpl(src, src.getCount(), Arrays.asList(options));
     }
 
+    @Override
+    public CompletableFuture<DataStreamReply> commandAsync(ByteBuffer src) {
```
ss77892, line 191: What would happen on the client side if it issues two commands without data in between? As far as I can see, NettyClientReply, when it maps requests, considers only the stream offset and type, so the second command with the same offset wouldn't have a ReplyEntry.

amaliujia: Very good call. It seems to me that the first command's `ReplyEntry` will be used for the second command  and the second command's `RequestEntry` won't be in the map.  Anyway it looks like a mess without a handling.

So I instead fail the commands if there is already one at the same stream offset. After all, we do not expect the caller issue multiple commands at the same stream offset for now.

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientReplies.java
```diff
@@ -60,6 +60,19 @@ class ReplyMap {
...
+              + "; wait for the previous command reply before sending another");
+          f.completeExceptionally(exception);
+          return null;
+        }
+        final ReplyEntry entry = new ReplyEntry(isClose, f);
+        map.put(requestEntry, entry);
+        return entry;
+      }
       // ConcurrentHashMap.computeIfAbsent javadoc: the function is applied at most once per key.
       return map.computeIfAbsent(requestEntry, r -> new ReplyEntry(isClose, f));
```
szetszwo:
```java
    ReplyEntry submitRequest(RequestEntry requestEntry, boolean isClose, CompletableFuture<DataStreamReply> f) {
      LOG.debug("put {} to the map for {}", requestEntry, clientInvocationId);
      final MemoizedSupplier<ReplyEntry> supplier = MemoizedSupplier.valueOf(() -> new ReplyEntry(isClose, f));
      final ReplyEntry reply = map.computeIfAbsent(requestEntry, r -> supplier.get());
      if (requestEntry.type == Type.STREAM_COMMAND && !supplier.isInitialized()) {
        final IllegalStateException exception = new IllegalStateException(
            "STREAM_COMMAND already exist: " + requestEntry + " for " + clientInvocationId);
        f.completeExceptionally(exception);
        return null;
      }
      return reply;
    }
``` 
- We should only make one call to the map as above; otherwise, it is not atomic.  In this case, we may use MemoizedSupplier.
- Let's also make the exception message shorter; see below
Image

szetszwo (commented):
@ss77892 , Good catch one the problem!

@amaliujia , thanks for the update!  Please see the comment inlined.

szetszwo:
> RATIS-2625.  Streaming: support control plane in Data Stream

[Control plane](https://en.wikipedia.org/wiki/Control_plane) seems not the right term.  How about revising the title to below?
- Streaming: support sending commands in Data Stream

ratis-netty/src/main/java/org/apache/ratis/netty/client/NettyClientReplies.java
amaliujia: Done! 

Thanks for the hint. The map is concurrent map so two accesses (one containsKey, one put) are not atomic and will cause concurrency issue.

szetszwo (approved):
+1 the change looks good.

PR #1538, opened 2026-08-01
RATIS-2629. Handle Netty based request asynchronously and improve exception handling

spacemonkd:

RATIS-2629. Handle Netty based request asynchronously and improve exception handling

Currently NettyRpcService handles every inbound request synchronously on the Netty IO event-loop thread and its inbound handler does not override exceptionCaught.

This can cause:
- a slow or long-running request occupies the worker thread delaying all other requests on that event loop
- only IOException is converted into an error reply. Any other Throwable escapes the handler with no exceptionCaught, Netty's default handling only logs it and never writes a reply, so the client blocks until its request timeout.

This patch offloads the request handling out of the core IO path, and closes the channel to prevent blocking.
This mirrors how gRPC handles the requests.

https://issues.apache.org/jira/browse/RATIS-2629

## How was this patch tested?
Patch was tested via unit tests.

### DIFF
--- a/ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyRpcService.java
+++ b/ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyRpcService.java
@@ -44,6 +44,7 @@ import org.apache.ratis.util.ConcurrentUtils;
 import org.apache.ratis.util.JavaUtils;
 import org.apache.ratis.util.MemoizedSupplier;
 import org.apache.ratis.util.ProtoUtils;
+import org.apache.ratis.util.TimeDuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -148,11 +149,6 @@ public final class NettyRpcService extends RaftServerRpcWithProxy<NettyRpcProxy,
       }
     };
 
-    this.requestExecutor = ConcurrentUtils.newThreadPoolWithMax(
-        NettyConfigKeys.Server.asyncRequestThreadPoolCached(server.getProperties()),
-        NettyConfigKeys.Server.asyncRequestThreadPoolSize(server.getProperties()),
-        server.getId() + "-request-");
-
     final boolean useEpoll = NettyConfigKeys.Server.useEpoll(server.getProperties());
     this.bossGroup = NettyUtils.newEventLoopGroup(CLASS_NAME + "-bossGroup", 0, useEpoll);
     this.workerGroup = NettyUtils.newEventLoopGroup(CLASS_NAME + "-workerGroup",0, useEpoll);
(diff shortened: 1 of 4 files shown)


spacemonkd:
@szetszwo  could you take a look?

ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyRpcService.java
```diff
@@ -116,6 +136,11 @@ protected void initChannel(SocketChannel ch) {
       }
     };
 
+    this.requestExecutor = ConcurrentUtils.newThreadPoolWithMax(
```
ss77892: So, it's still a single worker pool (corePoolSize is 0). But instead of using TCP backpressure, we are pushing everything to the unlimited queue and creating pressure on the memory. Moreover, previously, Netty’s worker EventLoops handled connection shards independently, so a blocked request delayed only that shard’s RPCs. The new default single request worker queues all inbound RPCs, including heartbeats, so one slow request can delay heartbeats and trigger leader election.

spacemonkd: Yes this was one issue which I missed and faced before (hence the test failure in flaky test suite).
`corePoolSize=0` + an unbounded queue causes `requestExecutor` to be single-threaded, and since handle() blocks until commit, this causes the timeouts.
I have addressed this by switching to a fixed pool for now as the related change would increase LoC.

Filed https://issues.apache.org/jira/browse/RATIS-2637 for the improvement as a follow up.

ss77892: Backpressure control should be a part of these changes. You have removed the one that was going through TCP flow control and don't provide any replacement.  Another thing: we have ThreadPoolExecutor with corePoolSize=0 and unbounded queue. Let me quote "Java Concurrency in Practice": 
**[3] Developers are sometimes tempted to set the core size to zero so that the worker threads will eventually be torn down and therefore won't prevent the JVM from exiting, but this can cause some strange‐seeming behavior in thread pools that don't use a SynchronousQueue for their work queue (as newCachedThreadPool does). If the pool is already at the core size, ThreadPoolExecutor creates a new thread only if the work queue is full. So tasks submitted to a thread pool with a work queue that has any capacity and a core size of zero will not execute until the queue fills up, which is usually not what is desired.** 
So, it's still a single thread.

ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyRpcService.java
```diff
@@ -155,6 +192,7 @@ public void startImpl() throws IOException {
 
   @Override
   public void closeImpl() throws IOException {
+    ConcurrentUtils.shutdownAndWait(requestExecutor);
```
ss77892: shutdownAndWait is running with a 1-day timeout, and if the executor is stuck, we won't close the channel and event-loop group for a day, would we?

spacemonkd: Ah, silly mistake, I missed this. Thanks for pointing it out. Addressed in the latest comment

ratis-netty/src/main/java/org/apache/ratis/netty/server/NettyRpcService.java
spacemonkd: > Another thing: we have ThreadPoolExecutor with corePoolSize=0 and unbounded queue.
the corePoolSize=0 + unbounded-queue cached pool is effectively single-threaded, however  this PR now defaults to a fixed pool (corePoolSize=32), so inbound RPCs including heartbeats are no longer serialized behind one slow request.

So I have effectively switched back to the earlier behaviour for the time being.
Let's avoid making this patch bigger and address separately as right now this is effective the previous behaviour.

PR #1540, opened 2026-08-02
RATIS-2632. Apply backpressure on install-snapshot chunk loop

spacemonkd:

RATIS-2632. Apply backpressure on install-snapshot chunk loop

When the leader streams a snapshot to a lagging follower, the chunk send loop in [GrpcLogAppender.installSnapshot()](https://github.com/apache/ratis/blob/a1bbf47f98a83d3bbddb4357c8a9b96bcfd7c727/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java#L775-L783) calls `snapshotRequestObserver.onNext(request)` for every chunk without ever checking `CallStreamObserver.isReady()`.

The append-entries path already [checks isReady()](https://github.com/apache/ratis/blob/a1bbf47f98a83d3bbddb4357c8a9b96bcfd7c727/ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcLogAppender.java#L366), but the snapshot path does not, so it is possible that chunks are enqueued faster than the network can drain them.

While on the default path `INSTALL_SNAPSHOT_REQUEST_ELEMENT_LIMIT_DEFAULT` is set to 8 so `StreamObserverWithTimeout` applies a request-response window, but since we allow setting the limit to 0, the semaphore can be `null` and cause:

- Unbounded outbound buffering of snapshot chunks in the gRPC/Netty write queue when the follower is slow.
- Leader OOM / direct-memory exhaustion during large-snapshot catch-up, which can crash the leader and impact the whole group.

This is handled by having an explicit isReady() check, which causes minimal overhead on the default path (`element-limit=8`) but handles the case when it might be set to 0.

https://issues.apache.org/jira/browse/RATIS-2632

## How was this patch tested?
Patch was tested via unit tests.

### DIFF
--- a/ratis-grpc/src/main/java/org/apache/ratis/grpc/util/StreamObserverWithTimeout.java
+++ b/ratis-grpc/src/main/java/org/apache/ratis/grpc/util/StreamObserverWithTimeout.java
@@ -19,6 +19,7 @@ package org.apache.ratis.grpc.util;
 
 import org.apache.ratis.protocol.exceptions.TimeoutIOException;
 import org.apache.ratis.thirdparty.io.grpc.ClientInterceptor;
+import org.apache.ratis.thirdparty.io.grpc.stub.CallStreamObserver;
 import org.apache.ratis.thirdparty.io.grpc.stub.StreamObserver;
 import org.apache.ratis.util.JavaUtils;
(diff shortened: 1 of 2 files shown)


spacemonkd:
@szetszwo @ss77892 could you take a look at this PR?

PR #1543, opened 2026-08-03
RATIS-2640. Fix AdminApi.setConfiguration(RaftPeer[], RaftPeer[]) dropping the servers array

hooji:

Fix the `AdminApi.setConfiguration(RaftPeer[] serversInNewConf, RaftPeer[] listenersInNewConf)` default method. It built its `Arguments` with

```java
.setListenersInNewConf(serversInNewConf)
.setListenersInNewConf(listenersInNewConf)
```

so the servers array was passed to the **listeners** setter and then overwritten by the real listeners. The builder's server list was never set, and `Arguments`'s constructor then hit `Preconditions.assertUnique(null)` — every call to this overload failed with a `NullPointerException`, i.e. the overload has never worked as its javadoc documents ("The same as setConfiguration(Arrays.asList(serversInNewConf), Arrays.asList(listenersInNewConf))"). No in-repo caller uses it, which is presumably why it went unnoticed.

The fix routes the servers through `setServersInNewConf(...)`, mirroring the List-based overload directly above it, and adds a regression test (`org.apache.ratis.client.TestAdminApi` in ratis-test) that captures the built `Arguments` through a stub `AdminApi` and asserts both two-argument overloads.

Found while evaluating Ratis for adoption; present at `ratis-3.2.2` and on current master (7eedc1dee).

https://issues.apache.org/jira/browse/RATIS-2640

## How was this patch tested?

New unit test `org.apache.ratis.client.TestAdminApi`:

- **Before the fix**: `testSetConfigurationWithArrays` fails with
  `java.lang.NullPointerException ... at org.apache.ratis.util.Preconditions.assertUnique(Preconditions.java:142) ... at org.apache.ratis.protocol.SetConfigurationRequest$Arguments.<init>(SetConfigurationRequest.java:59) ... at org.apache.ratis.client.api.AdminApi.setConfiguration(AdminApi.java:65)`
- **After the fix**: `Tests run: 2, Failures: 0, Errors: 0, Skipped: 0`

Also ran `checkstyle:check` and `apache-rat:check` on the changed modules (clean). Built and tested with JDK 21.

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/api/AdminApi.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/api/AdminApi.java
@@ -60,7 +60,7 @@ public interface AdminApi {
       throws IOException {
     return setConfiguration(SetConfigurationRequest.Arguments
         .newBuilder()
-        .setListenersInNewConf(serversInNewConf)
+        .setServersInNewConf(serversInNewConf)
         .setListenersInNewConf(listenersInNewConf)
         .build());
   }
--- a/ratis-test/src/test/java/org/apache/ratis/client/TestAdminApi.java
+++ b/ratis-test/src/test/java/org/apache/ratis/client/TestAdminApi.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


szetszwo (commented):
@hooji , Good catch on the bug!

+1 the change looks good.

szetszwo:
@hooji , could you file a [JIRA](https://issues.apache.org/jira/projects/RATIS/)?  We, Apache Ratis, use JIRA to track issues.

hooji:
Sorry, I'm not on JIRA.  We just noticed the bug while doing an AI survey of the code prior to making a decision between Ratis and SOFAJRaft.  Can you file the JIRA for us?

Here's a copy of the final evaluation if you're interested.  Written by Fable 5:

[RAFT_LIBRARY_EVALUATION.md](https://github.com/user-attachments/files/30686409/RAFT_LIBRARY_EVALUATION.md)

adoroszlai:
> Sorry, I'm not on JIRA.

@hooji Your request for a Jira user is approved, now you can proceed creating the issue.  Thanks.

hooji:
JIRA 2640 filed

szetszwo (approved):
@hooji , thanks for working on this and also filing the JIRA!

+1

PR #1546, opened 2026-08-05
RATIS-2642. Command API in data stream can be optional

amaliujia:

Command API in data stream has a default implementation so the the implementor is not forced to write it after the upgrade. This also means the new API is optional.

https://issues.apache.org/jira/browse/RATIS-2642

## How was this patch tested?

compliation pass.


amaliujia:
@szetszwo trying this API in Ozone and saw un-relevant tests failed to compile because they extend `DataStreamOutput`, and there is no need to ask those places to implement command API because that is irrelevant.

So thinking if have the new API as optional could be a good idea.

szetszwo (approved):
+1 the change looks good.

PR #1560, opened 2026-08-24
RATIS-2213. Reclaim unlinked data stream on close

rich7420:

A data stream's `DataStreamMap` entry is added on `STREAM_HEADER`
(`DataStreamManagement.computeDataStreamIfAbsent`) and removed by exactly two sites: the
Raft log worker when the `DATASTREAM` log entry is linked (`SegmentedRaftLogWorker`), or
`StreamInfo.cleanUp` on the error / channel-inactive path.

On a **successful CLOSE**, `DataStreamManagement` drains `streams` and the `channels` entry
but hands the `DataStreamMap` entry off to the log path — it does not remove it. If that
`DATASTREAM` log entry is never written (commit not logged: abort, leader change, follower
truncation), the entry is orphaned: the closed stream is no longer tracked by `streams` or
`channels`, so channel-inactive cleanup cannot reclaim it — and since the client may keep the
channel open (Ozone reuses it), channel-inactive may never fire. The orphaned
`CompletableFuture<DataStream>` pins the Netty direct buffers, so direct memory grows under
sustained streaming write. This matches the MAT analysis on RATIS-2213 / HDDS-11939.

This PR adds a close-time backstop (`scheduleUnlinkedStreamCleanup`): after the data-stream
request timeout, remove the `DataStreamMap` entry **iff it is still unlinked** and clean the
`DataStream`. A non-null remove result means the log worker has not linked it, so the
`DataStream` is not yet owned by the state machine and is safe to clean; if it was already
linked, remove returns null and the backstop is a no-op.

Open to feedback on the approach — in particular: reusing `raft.client.data-stream.request.timeout`
as the "abandoned" threshold vs. a dedicated grace period, and whether a link-completion callback
from the log worker would be preferable to a per-close scheduled task.

https://issues.apache.org/jira/browse/RATIS-2213

## How was this patch tested?
 https://github.com/rich7420/ratis/actions/runs/32698135039

### DIFF
--- a/ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
+++ b/ratis-netty/src/main/java/org/apache/ratis/netty/server/DataStreamManagement.java
@@ -522,6 +522,26 @@ public class DataStreamManagement {
     return removed;
   }
 
+  /**
+   * RATIS-2213: a closed stream's DataStreamMap entry is normally removed by the Raft log worker when the
+   * DATASTREAM log entry is linked. If that never happens (the commit is not logged -- e.g. abort, leader
+   * change or follower truncation), the entry and the DataStream's Netty buffers would leak: the closed
+   * stream is no longer tracked by {@code streams} or {@code channels}, and the client may keep the channel
+   * open (Ozone reuses it), so channel-inactive cleanup cannot reclaim it. Schedule a backstop that reclaims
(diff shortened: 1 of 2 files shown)


rich7420:
cc @SteNicholas

chungen0126:
@szetszwo PTAL

ferhui:
@xichen01 @ivandika3

szetszwo (approved):
+1 the change looks good.

PR #1561, opened 2026-08-24
RATIS-2638. Add a TLS handshake failure listener

HTHou:

Add an optional generic `ServerCredentials` configuration to `ratis-grpc`. Applications can register credentials through `GrpcConfigKeys.Server.setCredentials(...)`, and Ratis passes them directly to `NettyServerBuilder.forAddress(...)`. This supports custom credential implementations without specializing the configuration API for one use case.

This pull request also adds `TlsHandshakeFailureServerCredentials.create(GrpcTlsConfig, Consumer<TlsHandshakeFailureServerCredentials.Event>)` for applications that need to observe inbound gRPC TLS handshake failures. Ratis reports the original cause and the physical local and remote socket addresses. The callback runs on a transport event-loop thread and must not block. Listener failures are isolated from the TLS failure and channel-closing behavior. Ratis does not perform audit logging or require callers to persist the exception; applications may use only the endpoint information when that is sufficient for their audit policy.

The TLS handshake failure credentials wrap the shaded gRPC protocol negotiator inside Ratis. They report TLS and ALPN negotiation failures at most once per physical connection and ignore channel closure without an `SSLException`. Configured server credentials apply to SERVER, CLIENT, and ADMIN server endpoints, including shared ports. When no credentials are configured, the existing server creation and `SslContext` paths remain unchanged.

https://issues.apache.org/jira/browse/RATIS-2638

## How was this patch tested?

Added `TestRaftServerWithGrpc#testTlsHandshakeFailureListener`, which verifies that a valid TLS request does not trigger the listener and that a plaintext connection to the TLS port reports one inbound failure with the expected local and remote addresses.

Also ran:

- `mvn -pl ratis-test -am -Dtest=TestRaftServerWithGrpc#testTlsHandshakeFailureListener -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -B -fae -DskipTests checkstyle:check`
- `mvn -pl ratis-grpc -DskipTests clean compile spotbugs:check`


ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcConfigKeys.java
```diff
@@ -293,6 +293,22 @@ static void setServicesCustomizer(Parameters parameters, GrpcServices.Customizer
       parameters.put(SERVICES_CUSTOMIZER_PARAMETER, customizer, SERVICES_CUSTOMIZER_CLASS);
     }
 
+    String TLS_HANDSHAKE_FAILURE_LISTENER_PARAMETER = PREFIX + ".tls.handshake.failure.listener";
+    Class<TlsHandshakeFailureListener> TLS_HANDSHAKE_FAILURE_LISTENER_CLASS =
+        TlsHandshakeFailureListener.class;
```
szetszwo:
```java
NettyServerBuilder forAddress(SocketAddress address, ServerCredentials creds) 
```
Since the forAddress parameter is ServerCredentials, let's use ServerCredentials as the class.  Then, it will also support other ServerCredentials.
```java
    String CREDENTIALS_PARAMETER = PREFIX + ".credentials";
    Class<ServerCredentials> CREDENTIALS_CLASS = ServerCredentials.class;
    static ServerCredentials credentials(Parameters parameters) {
      return parameters == null ? null
          : parameters.get(CREDENTIALS_PARAMETER, CREDENTIALS_CLASS);
    }
    static void setCredentials(Parameters parameters, ServerCredentials listener) {
      parameters.put(CREDENTIALS_PARAMETER, listener, CREDENTIALS_CLASS);
    }
```

szetszwo (commented):
@HTHou , thanks for working on this!  

I only have a comment on the conf so far.  We should try to make it more general to support any ServerCredentials.

This part of the grpc code is new to me.  Will need some time to review the other code.

ratis-grpc/src/main/java/org/apache/ratis/grpc/GrpcConfigKeys.java
HTHou: Thanks for the suggestion. Updated in 058baf387: the parameter now stores a generic `ServerCredentials` via `GrpcConfigKeys.Server.setCredentials(...)`, and `GrpcServicesImpl` passes it directly to `NettyServerBuilder.forAddress(...)`. The TLS listener is exposed through `TlsHandshakeFailureServerCredentials.create(...)`, and the test registers that implementation through the generic credentials configuration, so other `ServerCredentials` implementations are supported as well.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcServicesImpl.java
```diff
@@ -218,9 +234,10 @@ private NettyServerBuilder newNettyServerBuilder(String hostname, int port, SslC
         nettyServerBuilder.workerEventLoopGroup(serverWorkers);
       }
 
-      if (sslContext != null) {
+      if (serverCredentials != null) {
+        LOG.info("Setting server credentials for {}", address);
+      } else if (sslContext != null) {
         LOG.info("Setting TLS for {}", address);
-        nettyServerBuilder.sslContext(sslContext);
       }
```
szetszwo: Move the LOG to above if-statement.
```java
      if (serverCredentials != null) {
        LOG.info("Setting server credentials for {}", address);
        nettyServerBuilder = NettyServerBuilder.forAddress(address, serverCredentials);
      } else {
        nettyServerBuilder = NettyServerBuilder.forAddress(address);
        if (sslContext != null) {
          LOG.info("Setting sslContext for {}", address);
          nettyServerBuilder.sslContext(sslContext);
        }
      }
```

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureEvent.java
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.ratis.grpc;
+
+import java.net.SocketAddress;
+import java.util.Objects;
+
+/** Information about a failed TLS handshake. */
+public final class TlsHandshakeFailureEvent {
```
szetszwo: Let's move it as an inner class TlsHandshakeFailureServerCredentials.Event.

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureListener.java
```diff
@@ -0,0 +1,28 @@
...
+
+/** Receives failures detected while establishing TLS connections. */
+@FunctionalInterface
+public interface TlsHandshakeFailureListener {
+  /**
+   * Invoked on a transport event-loop thread. Implementations must not block and should hand off
+   * expensive work to another thread.
+   */
+  void onFailure(TlsHandshakeFailureEvent event);
+}
```
szetszwo: Let's use `Consumer<TlsHandshakeFailureEvent>` instead of adding a new interface.

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureServerCredentials.java
```diff
@@ -0,0 +1,216 @@
...
+    public AsciiString scheme() {
+      return HTTPS;
+    }
+
+    @Override
+    public ChannelHandler newHandler(GrpcHttp2ConnectionHandler grpcHandler) {
+      final ChannelHandler grpcNegotiationHandler =
+          InternalProtocolNegotiators.grpcNegotiationHandler(grpcHandler);
+      final ChannelHandler tlsHandler = new ServerTlsHandler(grpcNegotiationHandler, grpcHandler,
+          sslContext, listener, offloadExecutorPool);
```
szetszwo: It seems better to pass the executor, instead of the offloadExecutorPool.

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureServerCredentials.java
```diff
@@ -0,0 +1,216 @@
...
+        context.fireExceptionCaught(cause);
+        return;
+      }
+
+      final SslHandler sslHandler = context.pipeline().get(SslHandler.class);
+      if (!sslContext.applicationProtocolNegotiator().protocols()
+          .contains(sslHandler.applicationProtocol())) {
+        final RuntimeException cause = Status.UNAVAILABLE
+            .withDescription("Failed protocol negotiation: Unable to find compatible protocol")
+            .asRuntimeException();
```
szetszwo: How about using `asException()`?

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureServerCredentials.java
```diff
@@ -0,0 +1,216 @@
...
+        return;
+      }
+      propagateTlsComplete(context, sslHandler.engine().getSession());
+    }
+
+    private void notifyListener(ChannelHandlerContext context, Throwable cause) {
+      if (failureReported) {
+        return;
+      }
+      failureReported = true;
```
szetszwo: It is better to use AtomicBoolean and compareAndSet:
```java
      if (failureReported.compareAndSet(false, true)) {
        return;
      }
```

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureServerCredentials.java
```diff
@@ -0,0 +1,216 @@
...
+        }
+        context.fireExceptionCaught(cause);
+        return;
+      }
+
+      final SslHandler sslHandler = context.pipeline().get(SslHandler.class);
+      if (!sslContext.applicationProtocolNegotiator().protocols()
+          .contains(sslHandler.applicationProtocol())) {
+        final RuntimeException cause = Status.UNAVAILABLE
+            .withDescription("Failed protocol negotiation: Unable to find compatible protocol")
```
szetszwo: Include the protocol string in the exception:
```java
      final String protocol = sslHandler.applicationProtocol();
      if (!sslContext.applicationProtocolNegotiator().protocols().contains(protocol)) {
        final Exception cause = Status.UNAVAILABLE
            .withDescription("Failed protocol negotiation: Unable to find compatible protocol for " + protocol)
```

szetszwo (commented):
@HTHou , thanks for the update!   Please see the comments inlined and also https://issues.apache.org/jira/secure/attachment/13084161/1561_review.patch

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureServerCredentials.java
HTHou: Updated in 64ca1d77a. The ALPN negotiation failure now uses `Status.asException()`.

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureListener.java
HTHou: Updated in 64ca1d77a. `create(...)` now accepts `Consumer<TlsHandshakeFailureServerCredentials.Event>`, and the standalone listener interface has been removed. I kept the non-blocking transport event-loop requirement in the API documentation.

ratis-grpc/src/main/java/org/apache/ratis/grpc/server/GrpcServicesImpl.java
HTHou: Updated in 64ca1d77a. The log statements are now inside the corresponding `serverCredentials` and `sslContext` branches.

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureServerCredentials.java
HTHou: Updated in 64ca1d77a. `Negotiator` now passes its acquired executor to `ServerTlsHandler`, avoiding a second `getObject()` call while retaining the existing `returnObject()` lifecycle.

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureServerCredentials.java
HTHou: Updated in 64ca1d77a. `failureReported` is now an `AtomicBoolean`. I used `if (!failureReported.compareAndSet(false, true)) { return; }` so that the first failure reaches the listener and subsequent reports are skipped.

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureServerCredentials.java
HTHou: Updated in 64ca1d77a. The negotiated protocol is captured and included in the ALPN failure description.

ratis-grpc/src/main/java/org/apache/ratis/grpc/TlsHandshakeFailureEvent.java
HTHou: Updated in 64ca1d77a. The event is now the nested `TlsHandshakeFailureServerCredentials.Event`, and the standalone event class has been removed.

szetszwo (approved):
+1 the change looks good.

PR #1562, opened 2026-08-24
RATIS-2665: Reduce idle log worker close latency

ted-wq-x:

`SegmentedRaftLogWorker` polls its I/O task queue with a one-second timeout.
When the worker is idle, closing it may wait until the current poll times out.

This change offers an internal wake-up task after setting `running` to `false`,
allowing an idle worker to exit immediately. The wake-up task bypasses normal
task execution, exception handling, metrics, and future completion.

The worker is not interrupted, so active flush and log-sync operations retain
their existing shutdown behavior. No public API is changed.

https://issues.apache.org/jira/browse/RATIS-2665

## How was this patch tested?

Added `testCloseWakesUpIdleWorker` to verify that a worker blocked while polling
an empty task queue is awakened and closed promptly.

Validation results:

- The new regression test passed 20 consecutive runs.
- All 55 tests in `TestSegmentedRaftLog` passed.
- Maven Checkstyle completed with no violations.


szetszwo (approved):
+1 the change looks good.

ted-wq-x:
Thank you for your review. The CI error doesn't seem to be caused by my commit.

PR #1564, opened 2026-08-24
RATIS-2661. Persist bootstrap configuration for group recovery

OneSizeFitsQuorum:

Persist the initial Raft configuration in a separate bootstrap metadata file before a newly added group is allowed to start successfully.

During recovery, the persisted Raft configuration remains authoritative. When no Raft configuration entry has been persisted yet, the bootstrap configuration restores the initial membership so the division can start and elect a leader. The bootstrap file is removed after a formal configuration is persisted.

The recovery path also rejects conflicting bootstrap membership, while allowing the same peer IDs and priorities to be supplied with updated peer metadata such as addresses. Invalid empty or transitional bootstrap configurations fail startup instead of silently recovering with an unsafe configuration.

https://issues.apache.org/jira/browse/RATIS-2661

## How was this patch tested?

- TestGroupManagementWithGrpc: 10 tests passed.
- TestGroupManagementWithNetty: 10 tests passed.
- TestGroupManagementWithSimulatedRpc: 10 tests passed.
- TestRaftStorage: 16 tests passed.
- Maven Checkstyle completed with 0 violations.

The new regression test dynamically creates a three-peer group, blocks voting before the initial configuration entry is appended, restarts all peers through directory recovery, and verifies that the stored bootstrap membership allows the group to start and elect a leader.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
+++ b/ratis-server/src/main/java/org/apache/ratis/server/impl/RaftServerImpl.java
@@ -1938,7 +1938,11 @@ class RaftServerImpl implements RaftServer.Division,
     case CONFIGURATIONENTRY:
       // the reply should have already been set. only need to record
       // the new conf in the metadata file and notify the StateMachine.
-      state.writeRaftConfiguration(next);
+      try {
+        state.writeRaftConfiguration(next);
+      } catch (IOException e) {
+        throw new RaftLogIOException(e);
+      }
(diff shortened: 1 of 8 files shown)


OneSizeFitsQuorum:
@szetszwo @CRZbulabula PTAL

szetszwo (approved):
+1 the change looks good.

OneSizeFitsQuorum:
Thanks a lot for reviewing this @szetszwo

PR #1569, opened 2026-08-28
RATIS-2680. Maintain backward compatibility for cleanupOldSnapshots

amaliujia:

Follow-up to #1320 (RATIS-244), which changed `SimpleStateMachineStorage.cleanupOldSnapshots` to count only snapshots with MD5 files toward the retention limit. Meanwhile RATIS-244 maintained the backward compatibility for `SimpleStateMachineStorage.findLatestSnapshot`.

Overall speaking, there could be applications never considered creating MD5 files for taking snapshots, but because of old implementation of the Ratis, those applications didn't know that their snapshots should be considered incomplete or corrupted. Ratis should consider such applications and maintain backward compatibility. Otherwise once those applications upgrade, they will surprisingly find the behavior changes of the snapshot.

That change broke backward compatibility during upgrades where old snapshots (created before MD5 support, without .md5 files) coexist with new snapshots (with MD5 files) or there is no new snapshots (with MD5 files). In those cases, cleanup could either skip deletion entirely or retain the wrong set of snapshots, because old snapshots were excluded from the retention count even though they were valid before the upgrade.

This PR restores the original retention behavior as a fallback: when fewer than numSnapshotsRetained MD5 snapshots are found, retain the newest numSnapshotsRetained snapshots regardless of whether they have MD5 files. This covers:

1. All old snapshots without MD5 — retention policy applies as it did before RATIS-244
2. Mixed old and new snapshots after upgrade — old snapshots count toward retention when there are not enough new MD5 snapshots to fill the quota
3. When enough MD5 snapshots exist, the RATIS-244 behavior is unchanged: retention is based on MD5 snapshots, and snapshots without MD5 that fall beyond that cutoff are still cleaned up.

https://issues.apache.org/jira/browse/RATIS-2680

## How was this patch tested?

Unit Tests

This PR was assisted by Cursor.

### DIFF
--- a/ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
+++ b/ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
@@ -137,7 +137,7 @@ public class SimpleStateMachineStorage implements StateMachineStorage {
       }
     }
 
-    if (deleteIdx < 0 && allSnapshotFiles.size() > numSnapshotsRetained) {
+    if (deleteIdx == -1 && allSnapshotFiles.size() > numSnapshotsRetained) {
       // Before RATIS-244, the newest numSnapshotsRetained snapshots (with or without md5) are retained.
       // For backward compatibility:
       // - All snapshots without MD5 : retain the newest numSnapshotsRetained snapshots.
--- a/ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
+++ b/ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
@@ -135,6 +135,14 @@ public class SimpleStateMachineStorage implements StateMachineStorage {
       }
     }
 
+    // Backward compatibility: before MD5 files existed, all snapshots counted toward
+    // retention. When there are fewer MD5 snapshots than numSnapshotsRetained (e.g.
+    // all old snapshots without MD5, or old snapshots mixed with new ones after upgrade),
+    // fall back to retaining the newest numSnapshotsRetained snapshots regardless of MD5.
+    if (deleteIdx < 0 && allSnapshotFiles.size() > numSnapshotsRetained) {
+      deleteIdx = numSnapshotsRetained;
+    }
+
     if (deleteIdx > 0) {
       allSnapshotFiles.subList(deleteIdx, allSnapshotFiles.size())
           .stream()
--- a/ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
+++ b/ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
@@ -120,11 +120,13 @@ public class SimpleStateMachineStorage implements StateMachineStorage {
     final List<SingleFileSnapshotInfo> allSnapshotFiles = getSingleFileSnapshotInfos(stateMachineDir.toPath());
     allSnapshotFiles.sort(Comparator.comparing(SingleFileSnapshotInfo::getIndex).reversed());
     int numSnapshotsWithMd5 = 0;
+    int lastSnapshotsWithMd5 = -1;
     int deleteIdx = -1;
 
     for (int i = 0; i < allSnapshotFiles.size(); i++) {
       final SingleFileSnapshotInfo snapshot = allSnapshotFiles.get(i);
       if (snapshot.hasMd5()) {
+        lastSnapshotsWithMd5 = i;
         if (++numSnapshotsWithMd5 == numSnapshotsRetained) {
           // We have found the last snapshot with an MD5 file that needs to be retained
           deleteIdx = i + 1;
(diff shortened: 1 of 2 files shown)


amaliujia:
R: @spacemonkd @szetszwo 
cc @peterxcli @adoroszlai

ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
```diff
@@ -135,6 +135,14 @@ public void cleanupOldSnapshots(SnapshotRetentionPolicy snapshotRetentionPolicy)
       }
     }
 
+    // Backward compatibility: before MD5 files existed, all snapshots counted toward
+    // retention. When there are fewer MD5 snapshots than numSnapshotsRetained (e.g.
+    // all old snapshots without MD5, or old snapshots mixed with new ones after upgrade),
+    // fall back to retaining the newest numSnapshotsRetained snapshots regardless of MD5.
+    if (deleteIdx < 0 && allSnapshotFiles.size() > numSnapshotsRetained) {
```
peterxcli: I see `numSnapshotsRetained` is fetched from a policy, but looks like must be the number of all snapshot files that has md5 then the logic here could be valid. 
not sure other polcies also make this valid.

peterxcli: also there is a nit: we only need to do this if we havent found snapshot to delete?
```suggestion
    if (deleteIdx == -1 && allSnapshotFiles.size() > numSnapshotsRetained) {
```

peterxcli (commented):
@amaliujia thanks for the patch!

szetszwo:
@amaliujia , thanks for fixing this!  Please file a new JIRA.  It is easier to track.

@spacemonkd , please help reviewing this.

ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
```diff
@@ -135,6 +135,14 @@ public void cleanupOldSnapshots(SnapshotRetentionPolicy snapshotRetentionPolicy)
...
+      deleteIdx = numSnapshotsRetained;
+    }
+
```
szetszwo: When snapshots are mixed (with and without md5), let's retain all snapshots with md5 and the newer snapshots without md5.
```diff
+++ b/ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
@@ -120,11 +120,13 @@ public class SimpleStateMachineStorage implements StateMachineStorage {
     final List<SingleFileSnapshotInfo> allSnapshotFiles = getSingleFileSnapshotInfos(stateMachineDir.toPath());
     allSnapshotFiles.sort(Comparator.comparing(SingleFileSnapshotInfo::getIndex).reversed());
     int numSnapshotsWithMd5 = 0;
+    int lastSnapshotsWithMd5 = -1;
     int deleteIdx = -1;
 
     for (int i = 0; i < allSnapshotFiles.size(); i++) {
       final SingleFileSnapshotInfo snapshot = allSnapshotFiles.get(i);
       if (snapshot.hasMd5()) {
+        lastSnapshotsWithMd5 = i;
         if (++numSnapshotsWithMd5 == numSnapshotsRetained) {
           // We have found the last snapshot with an MD5 file that needs to be retained
           deleteIdx = i + 1;
@@ -135,6 +137,14 @@ public class SimpleStateMachineStorage implements StateMachineStorage {
       }
     }
 
+    if (deleteIdx < 0 && allSnapshotFiles.size() > numSnapshotsRetained) {
+      // Before RATIS-244, the newest numSnapshotsRetained snapshots (with or without md5) are retained.
+      // For backward compatibility:
+      // - All snapshots without MD5             : retain the newest numSnapshotsRetained snapshots.
+      // - Snapshots mixed (with and without md5): retain all snapshots with md5 and the newer snapshots without md5.
+      deleteIdx = Math.max(numSnapshotsRetained, lastSnapshotsWithMd5 + 1);
+    }
+
     if (deleteIdx > 0) {
       allSnapshotFiles.subList(deleteIdx, allSnapshotFiles.size())
           .stream()
```

szetszwo (commented):
@amaliujia , thanks for working on this!

Consider a  a corruption case:
- numSnapshotsRetained = 3
- Snapshots:
```
snapshot.1_23
snapshot.1_22
snapshot.1_21
snapshot.1_20
snapshot.1_20.md5
```
The newest 3 snapshots are corrupted.  We should not delete snapshot.1_20.

peterxcli:
> Consider a a corruption case:
> 
> * numSnapshotsRetained = 3
> * Snapshots:
> 
> ```
> snapshot.1_23
> snapshot.1_22
> snapshot.1_21
> snapshot.1_20
> snapshot.1_20.md5
> ```

out of curiosity, which scenario would make this happen? I would expect all of files that have md5 would has greater index than those doesnt because we snapshot order must be linear so files that produced with newer version must newer than older ones

szetszwo:
> out of curiosity, which scenario would make this happen?  ...

Suppose there is some data corruption.

ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
amaliujia: Ack. It makes sense to differentiate the corrupted snapshots after the upgrade with legit snapshots before the snapshots (both are lack of MD5 files).

Have applied this code suggestion. Will ask @spacemonkd to review as well.

ratis-server/src/main/java/org/apache/ratis/statemachine/impl/SimpleStateMachineStorage.java
amaliujia: sure. I applied by only checking `deleteIdx == -1` but not `< 0`

szetszwo (approved):
+1 the change looks good.

PR #1571, opened 2026-08-29
RATIS-2662. Make testKillLeaderDuringReconf deterministic.

slfan1989:

This pull request makes `RaftReconfigurationBaseTest#testKillLeaderDuringReconf` deterministic and removes its stale `@Flaky("RATIS-2251")` annotation.

#### Purpose of the test
`testKillLeaderDuringReconf` verifies that an in-progress Raft `reconfiguration` can complete after the original leader is killed.

#### Root Cause

The root cause was that the test was intended to start the previously created new peer, but instead called `restartServer()`. 

This method removed the original peer, recreated it using the current target configuration, and then started the new instance.  

Consequently, the recreated peer prematurely considered itself a voting member even though the target configuration had not yet been committed by the Raft cluster. 

If its election timeout expired before it received AppendEntries from the leader, it could start a `PRE_VOTE` round. 

Since the leader's committed configuration did not yet contain the new peer, the leader could reject the request with `shouldShutdown=true`, causing the new peer to enter the `CLOSED` state.

#### Unit Test Execution Workflow

> 1. The new peer was intentionally created with an empty configuration

The test first called:

```java
cluster.addNewPeers(1, false);
```

The second argument is `startNewPeer`, so the new peer, `s3`, was created but intentionally left stopped.

`MiniRaftCluster#addNewPeers` creates the new server with an empty Raft group:

```
final RaftGroup raftGroup =
    RaftGroup.valueOf(group.getGroupId(), Collections.emptyList());
```

Therefore, the expected initial state of `s3` was:

```
Peer:          s3
Configuration: []
Started:       false
```

The empty configuration was intentional. It indicated that `s3` was not yet part of the committed Raft configuration and needed to be bootstrapped by the leader.

> 2. The test prepared the target group before Raft committed it

The test then added `s3` and removed an existing follower. 

For example:

```
Committed Raft configuration: [s0, s1, s2]

The committed configuration was the membership configuration already accepted by Raft.

===========

MiniRaftCluster target group:  [s0, s2, s3]

The MiniRaftCluster group was only the target configuration prepared by the test. 
At this point, s3 was present in the target group but was not yet a committed **voting** member.
```

> 3. The original test did not start the previously created peer

The test was intended to start the existing `s3`, but it called:

```
cluster.restartServer(newPeerId, false);
```

The false argument is the format flag. It does not mean "do not start the peer".

More importantly, `restartServer` does not simply call `start()` on the existing server. Its implementation performs the following operations:

```
killServer(serverId);
removeServer(serverId);

final RaftServer proxy = putNewServer(serverId, raftGroup, format);
proxy.start();
```

The overload used by the test passes the current `cluster.group` to this method:

```
return restartServer(serverId, group, format);
```

As a result, restartServer:
1. closed the original `s3`;
2. removed the original server instance;
3. recreated another server with the same peer ID;
4. used the current target group `[s0, s2, s3]` as its startup group;
5. started the recreated server.
 
The original server with the empty configuration was therefore discarded.

> 4. The logs confirmed that s3 started with the target configuration

The relevant failure logs showed:

```
s3: addNew group-...:[s3, s0, s2]
s3: start as a follower, conf={... peers:[s3, s0, s2]}
```

Timestamps, thread names, and unrelated fields have been omitted from these excerpts.

These messages confirmed that the recreated `s3` did not start with the expected empty configuration. Instead, it started as a regular follower and considered itself a member of `[s3, s0, s2]`.

This established the first configuration mismatch:

```
Configuration believed by s3: [s3, s0, s2]
Configuration committed by the leader: [s0, s1, s2]
```

> 5. The incorrect configuration allowed s3 to start an election

Because the recreated `s3` considered itself a voting member, it started the normal follower election timer.

Two events could then race:

```
Event A: AppendEntries arrives from the leader
Event B: The election timeout of s3 expires
```

If AppendEntries arrived first, `s3` remained a follower and the test could pass.

If the election timeout expired first, `s3` changed to a candidate and started a PRE_VOTE round.

The logs confirmed the second path:

```
s3-FollowerState: change to CANDIDATE
```

This transition should not have occurred for the intended initializing peer before it was bootstrapped.

> 6. The leader rejected the PRE_VOTE because `s3` was not committed

When the leader received the PRE_VOTE request, its committed configuration still did not contain `s3`.

The logs showed:

```
s2-LEADER: reject PRE_VOTE from s3:
  s3 is not in current conf [s0, s1, s2]
```

This confirmed that the leader and `s3` had inconsistent views of the membership:

```
s3 believed:     s3 is already a voting member
Leader believed: s3 is not in the committed configuration
```

The leader's behavior was therefore consistent with its committed Raft configuration. The invalid state had been introduced by the test recreating `s3` with the target group too early.

> 7. The rejected PRE_VOTE caused s3 to shut down

Under the relevant timing conditions, the leader returned `shouldShutdown=true`.

The election code on `s3` interpreted this response as a `SHUTDOWN` result:

```
s3-LeaderElection: PRE_VOTE SHUTDOWN
s3: shutdown
```

Subsequent operations then encountered the closed server:

```
ServerNotReadyException: current state is CLOSED
```

The complete failure chain was therefore:

addNewPeers creates `s3` with an empty configuration

- restartServer removes the original `s3`
- restartServer recreates `s3` with the target configuration
- `s3` prematurely considers itself a voting member
- `s3` does not receive AppendEntries before its election timeout
- `s3` starts a PRE_VOTE round
- the leader rejects `s3` because it is not in the committed configuration
- the leader returns shouldShutdown=true
- `s3` receives PRE_VOTE SHUTDOWN
- `s3` enters the CLOSED state
- the reconfiguration cannot complete normally

#### Changes

- Start the existing new peer directly instead of recreating it with `restartServer`.
- Replace the fixed 1500 ms sleep with state-based waits for both the original and new leaders to bootstrap the new peer.
- Replace the manually managed thread with an `ExecutorService` and `Future`.
- Treat reconfiguration timeout and background-task exceptions as test failures.
- Require the new peer to remain `RUNNING` and always verify the final configuration.
- Cancel unfinished work and terminate the executor in `finally`.
- Replace the stale `@Flaky` annotation with an explicit test timeout.

RATIS-2662. Make testKillLeaderDuringReconf deterministic.

## How was this patch tested?

The complete GitHub Actions build passed successfully:

https://github.com/slfan1989/ratis/actions/runs/32639967700

The Simulated RPC variant was stress-tested 100 times using the `repeat-test` GitHub Actions workflow with `10` splits and `10` iterations per split.

All `10` matrix jobs, representing `100` test executions in total, completed successfully in the second workflow attempt:

https://github.com/slfan1989/ratis/actions/runs/33227408975/attempts/2

### DIFF
--- a/ratis-server/src/test/java/org/apache/ratis/server/impl/RaftReconfigurationBaseTest.java
+++ b/ratis-server/src/test/java/org/apache/ratis/server/impl/RaftReconfigurationBaseTest.java
@@ -41,7 +41,7 @@ import org.apache.ratis.server.raftlog.LogProtoUtils;
 import org.apache.ratis.server.raftlog.RaftLog;
 import org.apache.ratis.server.raftlog.RaftLogBase;
 import org.apache.ratis.server.storage.RaftStorageTestUtils;
-import org.apache.ratis.test.tag.Flaky;
+import org.apache.ratis.util.ConcurrentUtils;
 import org.apache.ratis.util.JavaUtils;
 import org.apache.ratis.util.LifeCycle;
 import org.apache.ratis.util.Slf4jUtils;
@@ -56,8 +56,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


slfan1989:
I’m working on a detailed explanation of the unit test workflow and will update the PR shortly.

ratis-server/src/test/java/org/apache/ratis/server/impl/RaftReconfigurationBaseTest.java
```diff
@@ -507,86 +513,119 @@ void runTestBootstrapReconf(int numNewPeer, boolean startNewPeer, CLUSTER cluste
...
-
-      final RaftPeerId killed = RaftTestUtil.waitAndKillLeader(cluster);
-      Assertions.assertEquals(leaderId, killed);
-      final RaftPeerId newLeaderId = RaftTestUtil.waitForLeader(cluster).getId();
-      LOG.info("newLeaderId: {}", newLeaderId);
-      TimeDuration.valueOf(1500, TimeUnit.MILLISECONDS).sleep();
 
-      LOG.info("start new peers: {}", c1.getAddedPeers());
-      for (RaftPeer np : c1.getAddedPeers()) {
-        cluster.restartServer(np.getId(), false);
```
slfan1989, line 569: Since the test adds exactly one peer, the loop is unnecessary. Starting the existing peer directly preserves its empty initial configuration and prevents it from prematurely participating in `PRE_VOTE` before being bootstrapped, eliminating the `PRE_VOTE` -> `SHUTDOWN` -> `CLOSED` race.

ratis-server/src/test/java/org/apache/ratis/server/impl/RaftReconfigurationBaseTest.java
```diff
@@ -507,86 +513,119 @@ void runTestBootstrapReconf(int numNewPeer, boolean startNewPeer, CLUSTER cluste
...
         }
-        final long commitIndex = leaderLog.getLastCommittedIndex();
-        Assertions.assertTrue(commitIndex <= 2, "commitIndex = " + commitIndex + " > 2");
-      }
-
-      final RaftPeerId killed = RaftTestUtil.waitAndKillLeader(cluster);
-      Assertions.assertEquals(leaderId, killed);
-      final RaftPeerId newLeaderId = RaftTestUtil.waitForLeader(cluster).getId();
-      LOG.info("newLeaderId: {}", newLeaderId);
-      TimeDuration.valueOf(1500, TimeUnit.MILLISECONDS).sleep();
```
slfan1989, line 565: A fixed `1500 ms` sleep does not guarantee that the new leader has taken over the reconfiguration and started bootstrapping the new peer. 
Replace it with a state-based wait that confirms the new leader recognizes `newPeerId` as a bootstrapping peer before the peer is started.

ratis-server/src/test/java/org/apache/ratis/server/impl/RaftReconfigurationBaseTest.java
```diff
@@ -507,86 +513,119 @@ void runTestBootstrapReconf(int numNewPeer, boolean startNewPeer, CLUSTER cluste
...
+              + setConfTask.isDone() + ", cancelled=" + setConfTask.isCancelled() + "], "
+              + cluster.printServers(), e);
+        } catch (ExecutionException e) {
+          throw new AssertionError("setConfiguration failed; newPeer=" + newPeerId
+              + ", newPeerState=" + newDivision.getInfo().getLifeCycleState() + ", "
+              + cluster.printServers(), e.getCause());
+        }
 
-      try {
-        setConf.get(10, TimeUnit.SECONDS);
```
slfan1989, line 573: A timeout means that the expected reconfiguration has not completed. Ignoring it allows the test to continue with an unknown state and may produce a false success, especially when the final checks are skipped for a closed peer. Treat the timeout as a test failure and include the
`task`, `peer`, and `cluster` states in the diagnostic message.

ratis-server/src/test/java/org/apache/ratis/server/impl/RaftReconfigurationBaseTest.java
```diff
@@ -507,86 +513,119 @@ void runTestBootstrapReconf(int numNewPeer, boolean startNewPeer, CLUSTER cluste
...
+    return ((RaftServerImpl) server).getRole().getLeaderState()
+        .filter(LeaderStateImpl::inStagingState)
+        .map(state -> state.isBootStrappingPeer(peerId))
+        .orElse(false);
+  }
+
   void runTestKillLeaderDuringReconf(CLUSTER cluster) throws Exception {
-    final AtomicBoolean clientRunning = new AtomicBoolean(true);
-    Thread clientThread = null;
+    final ExecutorService executor = ConcurrentUtils.newSingleThreadExecutor(
```
slfan1989, line 537: Run `setConfiguration` asynchronously in a named single-thread executor so that the test can kill the leader while reconfiguration is in progress. The returned `Future` allows the test to validate the reply, propagate failures, and clean up the background task reliably.

slfan1989: Replace the manually managed client thread, control flag, and `CompletableFuture` with an `ExecutorService` and `Future`.  

This simplifies the asynchronous workflow and provides explicit result handling, exception propagation, cancellation, timeout control, and deterministic cleanup.

slfan1989:
@szetszwo @adoroszlai Could you please take a look at this PR when you have time? 

I believe we have identified the root cause of the flakiness in `testKillLeaderDuringReconf`. 

I have documented the failure sequence, supporting logs, root-cause analysis, and the rationale for the changes in the PR description.

Any feedback would be greatly appreciated. Thanks!

szetszwo (approved):
+1 the change looks good.

slfan1989:
> +1 the change looks good.

@szetszwo Thanks for the review and the +1! I really appreciate it.

PR #1575, opened 2026-09-01
RATIS-2677. Replace redundant setFirstRequest() call with an assertion

lugency:

Replace the redundant `setFirstRequest()` call in `OrderedAsync#sendRequestWithRetry()` with an assertion. The first-request flag has already been set by `SlidingWindow$Client#sendOrDelayRequest()` before the request is constructed. The assertion verifies this invariant.

This change also includes the first-request information in the corresponding log representations.

[RATIS-2677](https://issues.apache.org/jira/browse/RATIS-2677):
 In `OrderedAsync`, `setFirstRequest()` is not needed.

## How was this patch tested?

Tested locally:

- `TestRaftAsyncWithGrpc#testBasicAppendEntriesAsync`
- `TestRaftAsyncWithGrpc#testBasicAppendEntriesAsyncKillLeader`

### DIFF
--- a/ratis-client/src/main/java/org/apache/ratis/client/impl/OrderedAsync.java
+++ b/ratis-client/src/main/java/org/apache/ratis/client/impl/OrderedAsync.java
@@ -114,7 +114,7 @@ public final class OrderedAsync {
 
     @Override
     public String toString() {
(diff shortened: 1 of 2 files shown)


szetszwo (approved):
+1 the change looks good.

PR #1578, opened 2026-09-03
RATIS-2682. Use the original Raft group for client retries in testKillLeaderDuringReconf

slfan1989:

This pull request fixes a remaining race in `testKillLeaderDuringReconf`.

The test previously created the Raft client after calling `addNewPeers()` and `removePeers()`. These methods update `cluster.group` to the target configuration, which may contain the unstarted peer and exclude an original
voting member that can become the new leader after the original leader is killed.

With `SimulatedRpc`, a request sent to the unstarted peer may wait for `3` seconds before timing out. 
Repeated retries against that peer can exhaust the test's waiting period before the `setConfiguration` request reaches the new leader. 

The test then fails with an error such as:

```text
New leader s2 is not bootstrapping peer s3
```

RATIS-2682. Use the original Raft group for client retries in testKillLeaderDuringReconf

## How was this patch tested?

The target test passed 100 runs using 10 parallel splits with 10 iterations per split:

https://github.com/slfan1989/ratis/actions/runs/33755713261


szetszwo (approved):
+1 the change looks good.

slfan1989:
@szetszwo Thanks for the review and for merging the PR!

PR #1583, opened 2026-09-09
RATIS-2675. Fix TestTlsConfWithNetty may fail with "address already in use"

szetszwo:

RATIS-2675

  `TestTlsConfWithNetty.randomPort()` picked a random integer in `[50000, 60000)` without checking whether the port was actually free, so any process (a concurrent test, a lingering CI job, or an ordinary listener) already bound in that range made the subsequent `ServerBootstrap...bind(port).syncUninterruptibly()` throw `BindException: Address already in use`.

  Route the helper through `NetUtils.getFreePort()`, which probes each candidate port with a `ServerSocket` before handing it out — the same helper `MiniRaftCluster` and other Ratis tests already use.

  Fixes intermittent `TestTlsConfWithNetty` failures with `BindException: Address already in use`.

  ## Does this PR introduce any user-facing change?

  No. Test-only change.

  ## How was this patch tested?

  `mvn -pl ratis-test -Dtest=TestTlsConfWithNetty test` — all 8 tests pass.

(Spend $4.8412 including setup)


amaliujia (approved):
LGTM

szetszwo:
@amaliujia , thanks for reviewing this!

PR #1587, opened 2026-09-14
RATIS-2685. Follower gauges are stale after LogAppender restart

symious:

When a LogAppender is restarted, a new `FollowerInfo` is created, but the old follower gauges (`follower_<id>_next_index`, `match_index`, `rpc_response_time`) are never removed. The registry does not replace an existing gauge, so these gauges keep reading the stopped `FollowerInfo` and stop changing.

This PR removes the follower gauges when senders are stopped and before re-registering them.

https://issues.apache.org/jira/browse/RATIS-2685

## How was this patch tested?

Added `TestLogAppenderMetrics#testFollowerGaugesAfterReAdd`, which fails without the fix.


ratis-server/src/main/java/org/apache/ratis/server/metrics/LogAppenderMetrics.java
```diff
@@ -47,8 +47,16 @@ private static RatisMetricRegistry createRegistry(String serverId) {
...
     getRegistry().gauge(String.format(FOLLOWER_NEXT_INDEX, id), () -> getNextIndex::getAsLong);
     getRegistry().gauge(String.format(FOLLOWER_MATCH_INDEX, id), () -> getMatchIndex::getAsLong);
     getRegistry().gauge(String.format(FOLLOWER_RPC_RESP_TIME, id), () -> () -> getLastRpcTime.get().elapsedTimeMs());
   }
+
+  public void removeFollowerGauges(RaftPeerId id) {
+    getRegistry().remove(String.format(FOLLOWER_NEXT_INDEX, id));
+    getRegistry().remove(String.format(FOLLOWER_MATCH_INDEX, id));
+    getRegistry().remove(String.format(FOLLOWER_RPC_RESP_TIME, id));
+  }
```
szetszwo: Let's add a private replaceGauge method:
```java
  public void addFollowerGauges(RaftPeerId id, LongSupplier getNextIndex, LongSupplier getMatchIndex,
      Supplier<Timestamp> getLastRpcTime) {
    replaceGauge(String.format(FOLLOWER_NEXT_INDEX, id), getNextIndex);
    replaceGauge(String.format(FOLLOWER_MATCH_INDEX, id), getMatchIndex);
    replaceGauge(String.format(FOLLOWER_RPC_RESP_TIME, id), () -> getLastRpcTime.get().elapsedTimeMs());
  }

  private void replaceGauge(String name, LongSupplier supplier) {
    getRegistry().remove(name);
    getRegistry().gauge(name, () -> supplier::getAsLong);
  }
```

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
```diff
@@ -698,6 +698,7 @@ private void stopAndRemoveSenders(Predicate<LogAppender> predicate) {
   private void stopAndRemoveSenders(Collection<LogAppender> toStop) {
     toStop.forEach(LogAppender::stopAsync);
     senders.removeAll(toStop);
+    toStop.forEach(s -> logAppenderMetrics.removeFollowerGauges(s.getFollowerId()));
```
szetszwo: After the change in LogAppenderMetrics, this is not needed.  Let's don't add this line?

szetszwo (commented):
@symious , thanks for working on this!  Please see the comments inlined.

ratis-server/src/main/java/org/apache/ratis/server/metrics/LogAppenderMetrics.java
symious: Updated, PTAL.

ratis-server/src/main/java/org/apache/ratis/server/impl/LeaderStateImpl.java
symious: Updated, PTAL.

szetszwo (approved):
+1 the change looks good.

symious:
@szetszwo Test passed locally, could you help to retry the failed test?

szetszwo (approved):
+1 the change looks good.

symious:
@szetszwo Thank you for the review and merge.

PR #1589, opened 2026-09-15
RATIS-2695. Expose NotLeaderException generation in Division API

ivandika3:

See https://issues.apache.org/jira/browse/RATIS-2695

https://issues.apache.org/jira/browse/RATIS-2695

## How was this patch tested?

UT.


ratis-server-api/src/main/java/org/apache/ratis/server/RaftServer.java
```diff
@@ -82,6 +83,27 @@ default RaftPeer getPeer() {
...
+     * ID is unknown or is the ID of this division.</p>
+     */
+    default NotLeaderException newNotLeaderException() {
+      if (!getInfo().getLifeCycleState().isRunning()) {
+        return new NotLeaderException(getMemberId(), null, null);
+      }
+      RaftPeerId leaderId = getInfo().getLeaderId();
+      if (leaderId == null || leaderId.equals(getId())) {
+        // No idea about who is the current leader. Or the peer is the current
+        // leader, but it is about to step down. set the suggested leader as null.
```
amaliujia, line 121: It seems to be an issue when a user calls `newNotLeaderException` on a true leader, and in this case it shouldn't be as an `exception`, but should return a status to indicate this current leader is a true leader, etc.

The returned `leaderId=null` cannot distinguish if current leader is true leader, or current leader is about to step down, for example.

amaliujia: If we choose to expose this logic to application, I think we need a `checkLeader` API so it should support:
1. a server is a leader.
2. a server is a follower but it believes it is a leader
3. a server is a follower and it knows it is not a leader.

ivandika3: Thanks @amaliujia for the review.

> It seems to be an issue when a user calls newNotLeaderException on a true leader, and in this case it shouldn't be as an exception, but should return a status to indicate this current leader is a true leader, etc.

Yes, I also thought about this, but since SCM calls `checkLeader` before every `triggerNotLeaderException` then we simply need to handle the `NotLeaderException` generation. The `newNotLeaderException` will still get generated even if the current peer is a true leader. I have added the Javadoc for this.

> If we choose to expose this logic to application, I think we need a checkLeader API so it should support:

Good idea, I added a support for `checkLeaderReady`. This would support check the leadership and leadership readiness. I don't include the `LeaderSteppingDownException` for now. Previously Ozone SCM converted all to `NotLeaderException` even if leader is not ready yet. The `checkLeaderReady` would allow SCM client to not failover by throwing `LeaderNotReadyException`.

Regarding "a server is a follower but it believes it is a leader" and "a server is a follower and it knows it is not a leader", we cannot differentiate this without either leader doing a majority heartbeat check which can kill performance. So since Ozone does not guarantee linearizability in all cases, I think just checking the leader status locally is enough.

Ideally, Ozone should submit all requests to Ratis, even for reads (that will be handled by `StateMachine#query`) so Ozone doesn't need to check leadership, etc. However, from https://github.com/apache/ratis/pull/1448, we discovered that the protobuf serde in the current Raft submit has very high overhead, among other reasons.

amaliujia: Thanks for adding the extra API to make the user-facing API complete, as per my understanding we expect the APIs to be used in a general case as:
1. is my current node a ready leader by calling `checkLeaderReady`?
2. If not a leader (either not ready or ready), which is the suggested leader as a best hint by calling `newNotLeaderException`.

I am wondering if we only need a single new API as `checkLeaderReady` as a general use case. A caller will always call `checkLeaderReady` to determine 1. or 2. above?

ivandika3: > I am wondering if we only need a single new API as checkLeaderReady as a general use case. A caller will always call checkLeaderReady

Most of the time the user should use exception returned in `checkLeaderReady` (or null if the current peer is a leader), but `newNotLeaderException` might still needed for some custom use cases like when the server is stopping (similar to the `NotLeaderException` generated in `LeaderStateImpl#stop`). Also if we remove `newNotLeaderException`, the `Division#checkLeaderReady` cannot be a default method and we need the subclasses to implement this which can introduce the duplicate logic we are trying to avoid.

If you think that `checkLeaderReady` is enough, I can remove `newNotLeaderException`.

amaliujia: Sure I was only wondering if we need `newNotLeaderException` or no as a public API. If you could find a use case that will only need to call `newNotLeaderException` that is fine.

amaliujia (approved):
LGTM.

This is to expose new APIs but not touching existing API so I have less concern on the patch.

szetszwo:
@ivandika3 , @amaliujia , thanks for working on this!  Sorry for checking this late.  I do have some comments on the new public APIs

- I suggest to make the new method more general by adding an enum.
- Then, we could also clean up the code in RaftServerImpl; see https://issues.apache.org/jira/secure/attachment/13084585/1589_review.patch

What do you think?
```java
//RaftServer.Division (revised)
    /** @return the current leadership status of this division. */
    LeadershipStatus getCurrentLeadershipStatus();

    /**
     * Convert the given leadership status (not necessarily the current status) to an exception.
     *
     * @return For {@link LeadershipStatus#LEADER_READY}, return null.
     *         Otherwise, return an exception corresponding to {@link LeadershipStatus#getExceptionClass()}.
     */
    RaftException newLeadershipException(LeadershipStatus leadershipStatus);
```
```java
//RaftServer.Division (revised)
  enum LeadershipStatus {
    /** The division is a leader and it is ready. */
    LEADER_READY(null),
    /** The division is a leader, but it is NOT ready. */
    LEADER_NOT_READY(LeaderNotReadyException.class),
    /** The division is a leader, but it is stepping down. */
    LEADER_STEPPING_DOWN(LeaderSteppingDownException.class),
    /** The division is NOT a leader; */
    NOT_LEADER(NotLeaderException.class);

    private final Class<? extends RaftException> exceptionClass;

    LeadershipStatus(Class<? extends RaftException> exceptionClass) {
      this.exceptionClass = exceptionClass;
    }

    /**
     *  @return For {@link #LEADER_READY}, return null.
     *          Otherwise, return the corresponding exception class.
     */
    public Class<? extends RaftException> getExceptionClass() {
      return exceptionClass;
    }
  }
```

ivandika3:
Thanks @szetszwo for the review. Raised RATIS-2705.

szetszwo:
@ivandika3 , thanks for filing [RATIS-2705](https://issues.apache.org/jira/browse/RATIS-2705)!  I just revised the code in my previous comment:
- Renamed the first method to `getCurrentLeadershipStatus` for emphasizing "current".
- Moved the enum to `RaftServer.Division` since the status is per-division but not per-server.

PR #1591, opened 2026-09-15
RATIS-2659. IOUtils.readObject should apply an allow-list filter

szetszwo:

- According to Google AI responding to the question above, we should use ObjectInputFilter in IOUtils.readObject(..). Unfortunately, ObjectInputFilter is added to Java 9 but not in Java 8.

- Therefore, we will create our own filter by extending ObjectInputStream and overriding the resolveClass(..) method.  (Suggested by Claude)

- For compatibility, we have to support the current format for awhile before changing it to a safer format such as JSON.

Google AI comment

RATIS-2659

## How was this patch tested?

Added a new test (mostly by Claude).


adoroszlai (commented):
Thanks @szetszwo for the patch.  Should `newObjectInputStream` be used in examples, too?

https://github.com/apache/ratis/blob/2268ee5b74615bc4c54876098aa09d3a6d1cb1f4/ratis-examples/src/main/java/org/apache/ratis/examples/arithmetic/ArithmeticStateMachine.java#L132-L133

https://github.com/apache/ratis/blob/2268ee5b74615bc4c54876098aa09d3a6d1cb1f4/ratis-examples/src/main/java/org/apache/ratis/examples/counter/server/CounterStateMachine.java#L219

szetszwo:
@adoroszlai , thanks for the reviewing this!

> ... Should newObjectInputStream be used in examples, too?

Sure, let's also fix them.

adoroszlai (approved):

szetszwo:
@adoroszlai , thanks for reviewing and merging this!

PR #1595, opened 2026-09-17
RATIS-2701. LogSegment.putEntryCache() compute cache size incorrectly over existing keys

amaliujia:

Fixes incorrect totalCacheSize accounting in LogSegment.putEntryCache() when a cache entry is overwritten under the same key.

When duplicate inserts occur (e.g. concurrent segment loads), the method was computing the size of the replaced entry from the new value instead of the previous one. That made the delta zero on overwrite, so totalCacheSize could drift when the new entry had a different size than the old one. This in turn fed wrong memory accounting into SegmentedRaftLogCache.

The fix is a one-line change: use getEntrySize(previous, Op.REMOVE_CACHE) instead of getEntrySize(value, Op.REMOVE_CACHE).

Added testPutEntryCacheDuplicateKeyUpdatesTotalCacheSize to verify that overwriting a cached entry (small → large → small) keeps totalCacheSize aligned with the current entry size.

https://issues.apache.org/jira/browse/RATIS-2701

## How was this patch tested?

Unit Test

This PR is assisted by Cursor.


amaliujia:
R: @szetszwo

szetszwo (approved):
@amaliujia, Good catch on the bug!

+1 the change looks good

ratis-server/src/main/java/org/apache/ratis/server/raftlog/segmented/LogSegment.java
```diff
@@ -492,7 +492,7 @@ void putEntryCache(TermIndex key, LogEntryProto value, Op op) {
     long previousSize = 0;
     if (previous != null) {
       // Different threads maybe load LogSegment file into cache at the same time, so duplicate maybe happen
-      previousSize = getEntrySize(value, Op.REMOVE_CACHE);
+      previousSize = getEntrySize(previous, Op.REMOVE_CACHE);
```
szetszwo, line 495: Since value and previous are at the same TermIndex, they have to be the same. (If not, it is a more serious bug.)

It is still good to fix this minor bug.

PR #1596, opened 2026-09-17
RATIS-2700. Add a disallow-list for ObjectInputStream

szetszwo:
See RATIS-2700


ratis-common/src/main/java/org/apache/ratis/util/IOUtils.java
```diff
@@ -61,36 +68,80 @@ public interface IOUtils {
       "org.apache.ratis."
   ));
 
-  static void assertPackageForObjectInputStream(String fullClassName) throws InvalidClassException {
+  private static final Set<String> DISALLOWED_CLASSES;
+  static {
+    // all the classes below are subjected to java deserialization vulnerabilities
+    final List<String> disallowed = Arrays.asList(
```
amaliujia, line 74: I am curious how do we acquire this list? Is there an official place to check?

E.g. what if there are false positives or true negatives or missed class names.

szetszwo: By best effort.  See the comments in [RATIS-2700](https://issues.apache.org/jira/browse/RATIS-2700).

amaliujia: Ack the list sourced from AI models.

amaliujia (approved):

szetszwo:
@amaliujia , added a test for list.  Could you take a look again?

ratis-common/src/test/java/org/apache/ratis/util/TestIOUtils.java
```diff
@@ -50,19 +68,91 @@ public void testReadObjectAllowsStackTraceElements() {
...
+  public void testAllowedList() throws Exception {
+    // BigDecimal is in java.math which is in neither the allow-list nor the disallow-list
+    final BigDecimal pi = new BigDecimal("3.14");
+    assertDisallowed(pi, false);
+
+    // Hide it inside a list
+    final List<Object> list = new ArrayList<>();
+    list.add("testing");
+    list.add(pi);
+    assertDisallowed(list, false);
```
amaliujia, line 80: just a readability suggestion:

rename `assertDisallowed(list, false)` to `assertAllowed(list)` and handle the true/false inside.

amaliujia (approved):
LGTM, just a minor comment

ratis-common/src/test/java/org/apache/ratis/util/TestIOUtils.java
szetszwo: > rename assertDisallowed(list, false) to assertAllowed(list) and handle the true/false inside.

assertDisallowed actually is correct:  The list has two elements, String and BigDecimal.  
- assertDisallowed(list, ..): The assertion is to make sure that BigDecimal is disallowed, although it is inside the list.
- assertDisallowed(.., false): the false here mean "ArrayList is NOT in the disallowed list"

amaliujia: ah sure thanks!

PR #1606, opened 2026-09-21
RATIS-2711. Fix double position counting and remove redundant forced assignment

lugency:

- Remove `LimitedInputStream#read(byte[])` to avoid counting bytes twice.
- Remove the redundant `forced = false` assignment in `BufferedWriteChannel#flushBuffer()`.
- Add a regression test covering direct reads and reads through `DataInputStream`.
- Suppress the resource-leak warning in `SegmentedRaftLogReader#newInstance()`

https://issues.apache.org/jira/browse/RATIS-2711

## How was this patch tested?

Tested locally:
- `TestSegmentedRaftLogReader#testLimitedInputStreamPosition`
- `TestBufferedWriteChannel#testWriteToBuffer`


szetszwo (approved):
+1 the change looks good.

